PR #8, opened 2017-02-02
See ya later libhttp-parser

quinnj:


quinnj:
@tkelman, do you know what's going on w/ appveyor here? Is it a setting I need to tweak? It seems to not actually be running any tests.

tkelman:
It isn't running any tests because all you're running in https://github.com/JuliaWeb/HTTP.jl/blob/bdaa2fb1106aa088ea1569cb09653e7ccbc409f8/appveyor.yml is ` - C:\projects\julia\bin\julia -e "versioninfo();` ... without even a close paren

quinnj:
ah, thanks for taking a look. I'll get it cleaned up.

Keno:
I'm very happy about this. When I originally wrote all this stuff I considered just re-writing it in julia, but time got in the way (I did for the URIParser though). This is great.

Keno:
Don't check in `http_compiles.csv`?

src/uri.jl
```diff
@@ -1,435 +1,144 @@
...
-    args = Pair{Symbol}[(nms[i] => getfield(uri, i)) for i = 1:nfields(HTTP.URI)]
-    for (k, v) in kwargs
-        for (i, arg) in enumerate(args)
-            if arg[1] == k
-                args[i] = Pair(k, v)
-            end
-        end
-    end
-    return URI([arg[2] for arg in args]...)
+function URI(;hostname::String="", path::String="",
```
Keno: Please don't duplicate this functionality from URIParser.jl

quinnj: I've moved all the URIParser code here and rewrote parts. Mainly storing it like libhttp-parser w/ a `NTuple{7, Offset}`.

Keno: Why not just depend on that package?

Keno: Ah, I see what you're saying. Is there any way we can support both use cases in URIParser.jl?

Keno: I'd really hate to have two divergent implementations of URI parsing.

Keno: E.g. move this parsing code to URIParser, and have a final (optional) to convert it to a URI object with extracted components.

quinnj: Couple of reasons:

  * In general, the webstack has been poorly maintained, I believe partly because things were too separated out and there was quite a bit of code duplication across the packages (or code in packages where it didn't belong); downstream packages were almost exclusively depending on Requests or HttpServer anyway
  * The URIParser code is pretty out-dated and the API isn't very consistent; while cleaning it up, I ended up going down the rabbit-hole of porting the libhttp-parser code over and so decided to restructure entirely anyway
  * The new representation is quite a bit more efficient now in terms of storage and performance

It might make sense to split some functionality back off in the future, but for now, I'd of the belief that the webstack will be better maintained if issues and dev can be centralized as much as possible.

quinnj: I'm planning on doing a more thorough review of all the webstack downstream dependencies to figure out what exactly people need; if there are enough cases where people only need URI parsing w/o request functionality, I'm happy to update URIParser.

tkelman: There are plenty of packages that depend on URIParser but not Requests or HttpServer. Switching to a larger monolithic replacement isn't going to be desirable for those packages, so you're going to cause a split here.

Keno: BinDeps.jl uses URI without requests.

tkelman: and Elly, FilePaths, HDFS, LanguageServer, PkgDev, RDF, Transit, and WinRPM

PR #15, opened 2017-02-12
do not output the default port in the header

bicycle1885:
Removing default port number in the Host field would be a conventional way for HTTP clients (at least, curl and Google Chrome do this by default). Some silly web servers do not support this and throw an error when a port number is included in the host field (e.g. https://rest.ensembl.org/documentation/info/archive_id_get).


quinnj:
Ah, thanks for the input here.

PR #21, opened 2017-02-16
Ensure that we never have the "running within tasks" issue again.

staticfloat:
Adds a test running within a background task


quinnj:
Merging, I'll fix 0.6 later today.

Issue #30, opened 2017-02-27
Thoughts of defining `take!(::Response)`?


malmaud:
I wonder if it would be a good idea to essentially implement the `IOBuffer` API from Julia .6 for `Response` objects.

quinnj:
like
```julia
take!(::Type{String}, Response)
take!(Response) = take!(Vector{UInt8}, Response)
```
?

I really like that; it seems much less ad-hoc than the current `string` and `bytes` methods I threw in there. It also helps with indicating that you're actually _consuming_ the response, since you actually are consuming the data in teh underlying `FIFOBuffer`.

quinnj:
Implemented here #31

PR #31, opened 2017-02-28
Addresses #30

quinnj:


tkelman:
This is spelled `String(take!(x))` everywhere else. Why spell it differently here, and only here? That's even your implementation.

PR #32, opened 2017-03-01
Multipart patches

staticfloat:
This PR has a little bit of cleanup of the multipart uploading handling; first, we ensure that the `Content-Type` header for multipart uploads is on its own line, there were cases where it could be smooshed up against the `Content-Disposition`.

Next, we ensure that even `IOBuffer`'s will have their places `mark()`'ed and then `reset()` when printing out verbosely.  Otherwise, the buffers would get consumed during printing and would never make it out onto the wire.

Finally, we cleanup the `hasname()` stuff and ensure that we always have a `filename=` part, because the web frameworks I'm using require it, and I figure it can't hurt.  This is the most questionable design choice here, as I'd like some way to control this independently of the key we're passing in with the dict, but this is at least a reasonable default.


staticfloat:
In particular, I think we might need some way to specify metadata about the entries in the multipart `Dict`.  For example, although `HTTP.sniff()` is really cool, I'm uploading `audio/flac` files which aren't autodetected.  I'm not sure what the best API is, maybe an `HTTP.Multipart` type that has fields for things like filename, mimetype, IO source to pull data from, etc?

quinnj:
Working on this; I'm on iteration about 6, so we'll see how many different ways I need to think about solving this........

PR #39, opened 2017-03-18
use String(take!(r)) instead of take!(String, r)

tkelman:
no other IO types do the latter


quinnj:
Ah, thanks for catching this. Not sure why I thought it was this way.

tkelman:
maybe thinking of unsafe_wrap?

nalimilan:
Funny, that was my first reaction when I read code doing `take!(String, ...)`. :-)

PR #42, opened 2017-04-18
Follow standard for `read(..., ::Type{UInt8})`

### LINKED ISSUE #41: FIFOBuffer does not follow `read(::Any, ::Type{UInt8})` standard
I'm trying to use `readbytes!` with `FIFOBuffer` but I'm encountering errors since `read(::FIFOBuffer, ::Type{UInt8})` returns a `Tuple{UInt8,Bool}` rather than the typical `UInt8`.

```julia
julia> using HTTP

julia> f = HTTP.FIFOBuffer(5);

julia> b = Array{UInt8}(3);

julia> write(f, [0x01, 0x02, 0x03, 0x04, 0x05])
5

julia> readbytes!(f, b)
ERROR: MethodError: Cannot `convert` an object of type Tuple{UInt8,Bool} to an object of type UInt8
This may have arisen from a call to the constructor UInt8(...),
since type constructors fall back to convert methods.
 in readbytes!(::HTTP.FIFOBuffer, ::Array{UInt8,1}, ::Int64) at ./io.jl:351
 in readbytes!(::HTTP.FIFOBuffer, ::Array{UInt8,1}) at ./io.jl:342

julia> io = IOBuffer([0x01, 0x02, 0x03, 0x04, 0x05])
IOBuffer(data=UInt8[...], readable=true, writable=false, seekable=true, append=false, size=5, maxsize=Inf, ptr=1, mark=-1)

julia> readbytes!(io, b)
3
```

Could you use a different function name for this behaviour if it is required?

omus:
Fixes #41

### DIFF
--- a/src/fifobuffer.jl
+++ b/src/fifobuffer.jl
@@ -158,7 +158,7 @@ function Base.read(f::FIFOBuffer, nb::Int)
     return bytes
 end
 
-function Base.read(f::FIFOBuffer, ::Type{UInt8})
+function Base.read(f::FIFOBuffer, ::Type{Tuple{UInt8,Bool}})
     f.nb == 0 && return 0x00, false
     # data to read
     @inbounds b = f.buffer[f.f]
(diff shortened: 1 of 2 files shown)


quinnj:
Looks great!

PR #43, opened 2017-04-18
Respect user defined threshold

omus:
Uses the user defined threshold as the step instead of always using 50000000. Also avoids calling `flush` when verbose is false.


quinnj:
Thanks!

PR #44, opened 2017-04-19
Explicitly test versions on Travis, mention 0.6 in README

ararslan:
I changed `release` in the Travis YAML to explicit entries for 0.5 and 0.6. The latter is currently equivalent to `nightly` but won't be once 0.6 branches into a release, so I figured it's easier to change it now and it'll just continue to work as expected once 0.6 is released. I also made the 0.6 support clear in the README.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ os:
   - linux
   - osx
 julia:
-  - release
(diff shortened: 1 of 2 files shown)


quinnj:
Great! Thanks!

PR #49, opened 2017-05-08
Fix early EOF from when we know Content-Length

### LINKED ISSUE #45: Early EOF with streaming
In certain situations using `stream=true` will always return EOF:
```julia
julia> using HTTP

julia> stream = HTTP.body(HTTP.get("https://julialang.s3.amazonaws.com/bin/linux/x64/0.5/julia-0.5.1-linux-x86_64.tar.gz.asc", stream=true))
HTTP.FIFOBuffer(801,1048576,801,1,802,UInt8[0x2d,0x2d,0x2d,0x2d,0x2d,0x42,0x45,0x47,0x49,0x4e  …  0x54,0x55,0x52,0x45,0x2d,0x2d,0x2d,0x2d,0x2d,0x0a],Condition(Any[]),Task (done) @0x0000000114cd61d0,true)

julia> eof(stream)
true

julia> nb_available(stream)
801

julia> read(stream, 1)
1-element Array{UInt8,1}:
 0x2d

julia> eof(stream)
true
```

omus:
Fixes #45

### DIFF
--- a/src/fifobuffer.jl
+++ b/src/fifobuffer.jl
@@ -54,19 +54,20 @@ type FIFOBuffer <: IO
     buffer::Vector{UInt8}
     cond::Condition
     task::Task
-    eof::Bool
+    content_pos::Int64    # total number of bytes read from buffer
+    content_length::Int64 # expected number of bytes to read from buffer. content_pos == content_length marks EOF
 end
 
 const DEFAULT_MAX = Int64(typemax(Int32))^Int64(2)
 
 FIFOBuffer(f::FIFOBuffer) = f
-FIFOBuffer(max) = FIFOBuffer(0, max, 0, 1, 1, UInt8[], Condition(), current_task(), false)
+FIFOBuffer(max) = FIFOBuffer(0, max, 0, 1, 1, UInt8[], Condition(), current_task(), 0, -1)
 FIFOBuffer() = FIFOBuffer(DEFAULT_MAX)
 
 FIFOBuffer(str::String) = FIFOBuffer(Vector{UInt8}(str))
 function FIFOBuffer(bytes::Vector{UInt8})
     len = length(bytes)
-    return FIFOBuffer(len, len, len, 1, 1, bytes, Condition(), current_task(), false)
+    return FIFOBuffer(len, len, len, 1, 1, bytes, Condition(), current_task(), 0, -1)
 end
 FIFOBuffer(io::IOStream) = FIFOBuffer(read(io))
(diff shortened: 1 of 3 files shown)


src/fifobuffer.jl
```diff
@@ -79,10 +80,10 @@ Base.read(f::FIFOBuffer) = readavailable(f)
 Base.flush(f::FIFOBuffer) = nothing
 
 function Base.eof(f::FIFOBuffer)
-    return f.eof && f.nb == 0
+    return f.content_length >= 0 && f.content_pos >= f.content_length || f.content_length < 0 && f.nb == 0
 end
 function Base.close(f::FIFOBuffer)
-    f.eof = true
+    f.content_length = f.content_pos
```
omus: Not strictly necessary.

omus:
Note that this only fixes the problem for responses that contain ["Content-Length" in the headers](https://tools.ietf.org/html/rfc7230#section-3.3.2). For responses that that don't we'll still observe the early EOF bug.

src/fifobuffer.jl
quinnj: I think this case is a little more problematic; what if the buffer gets closed before `content_pos` has reached `content_length`?

quinnj: I guess it somewhat depends on if `content_pos` is tracking the # of bytes *written* to the buffer, or # of bytes *read* from buffer. I think the latter is simpler and is probably why you went that route?

omus: > I think this case is a little more problematic; what if the buffer gets closed before `content_pos` has reached `content_length`?

If a user attempts to read the last available byte in the buffer then the read will be blocked. It is up to the buffer writer to either close the `FIFOBuffer` which could potentially stop a read before it finished reading all available data or update the `content_length` such that reading automatically stops.

> I guess it somewhat depends on if `content_pos` is tracking the # of bytes *written* to the buffer, or # of bytes *read* from buffer.

`content_pos` is definitely tracking the number of bytes *read* from the buffer. It may make sense to rename these `read_pos` instead of `content_pos` and `expected_length` instead of `content_length`.

quinnj: I actually think the correct behavior here is to do nothing:
1. If there was no `content_length`, it doesn't matter what we do
2. If there was a `content_length`, we don't want to change anything because `eof` relies on having `content_pos` >= `content_length`; we wouldn't want an early `eof` because the buffer got closed. `close` means no more writing, but doesn't preclude reading what's left in the buffer.

PR #50, opened 2017-05-10
Treat server option args as keyword args for ServerOptions

joshbode:


quinnj:
Thanks!

joshbode:
no worries :)

Issue #60, opened 2017-06-19
12 "Transfer-Encoding: chunked" Tests Fail Due to httpbin issue


jlewis91:
It looks like there is an [unresolved httpbin issue / feature regression](https://github.com/kennethreitz/httpbin/issues/340) with chunked POST requests which is causing 10 of the 12 current test failures. Can anyone confirm this?

quinnj:
Yes, this is an upstream error. I can run the same tests against a local julia server & it runs fine.

samoconnor:
https://github.com/kennethreitz/httpbin/issues/340#issuecomment-330176449

samoconnor:
Can we just remove these tests? They're not providing us with any value as they are.
The httpbin guys seem to be unsure which part of their hosting stack is causing the problem and don't seem to have control over all parts of the stack. We have other test coverage for chunked encoding.

quinnj:
I'm fine w/ that.

PR #61, opened 2017-06-25
Custom Associative type for Header

garborg:
Makes Header keys case-insensitive. Motivated by http libs in other
languages and https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2.

I haven't worked much with the current iteration of the type system or the new String type -- open to any hints.

I was surprised there wasn't something in place already. I thought of two approaches:

- custom subtype of `Associative` with feature parity w/ `Dict` (undertaken here)
- custom subtype of `AbstractString`

The custom string method would be more pervasively efficient (e.g. skip canonization when manually inserting keys that were previously in a Headers object), but I shied away from it for a couple reasons:

-  for automatic conversion on Dict insert, `CanonicalHeaderKey(s) == s` would have to hold
- ordering would have to disagree with ordering of the underlying `Char`s
- wasn't sure how subtle or involved getting feature parity with `String` would be
- `AbstractString` isn't as widely accepted as String, or as it should be

If you'd go the other way, I'd love to hear more.

Where I wasn't sure about implementation, I took inspiration from Go's 'net' suite of packages. Main thing I noticed there and didn't add is a global store of common header keys. I'm guessing the cost of `String(v::Vector{UInt8})` is noticeable since I've seen optimizations for common headers elsewhere. Didn't want to get that involved before this made it in.

I also left the values strings rather than vectors. I haven't used both enough to have useful opinions there. For the use case where someone wants so say, send two different cookies each including commas, manually reaching into the dict (via `header.d`) seems like a convenient-enough escape hatch.

N.b. I hit #60 in local testing.

### DIFF
--- a/src/headers.jl
+++ b/src/headers.jl
@@ -58,11 +58,11 @@ immutable AlreadyCanonical
 end
 
 """
-    Headers() <: Associative{String,String}()
+    Headers <: Associative{String,String}
 
 A type to represent headers. Keys containing invalid characters are returned as is. Valid keys are canonicalized, uppercasing the first character and characters following a '-' and lowercasing other characters. So, for valid keys, Headers is case-insensitive.
 """
-immutable Headers{K<:String,V<:String} <: Associative{K,V}
+immutable Headers{K<:String,V<:String} <: Associative{String,String}
     d::Dict{K,V}
 
     function Headers{K,V}() where V where K
--- a/src/headers.jl
+++ b/src/headers.jl
@@ -1,0 +1,246 @@
+# See https://httpwg.github.io/specs/rfc7230.html#rule.token.separators
+const validHeaderBytes = IntSet(Int(c) for c in
+   raw"!#$%&'*+-.0123456789ABCDEFGHIJKLMNOPQRSTUWVXYZ^_`abcdefghijklmnopqrstuvwxyz|~"
+)
+
+validHeaderChar(c::Char) = isascii(c) && UInt8(c) in validHeaderBytes
+
+function canonicalHeaderKey(s::String)
+    # fast path for valid header keys
+    upper = true
+    for c in s
+        if !validHeaderChar(c)
+            return s
+        end
+        if upper && 'a' <= c <= 'z'
+            return _chk(s)
+        end
+        if !upper && 'A' <= c <= 'Z'
+            return _chk(s)
+        end
+
+        upper = c == '-'
+    end
+
+    return s
+end
+
+const toUpper = 'A' - 'a'
+
+function _chk(s::String)
+    # confirm its fixable before changing anything
+    for c in s
+        if !validHeaderChar(c)
+            return s
+        end
+    end
+
+    a = copy(Array{UInt8}(s))
+    upper = true
+    for (i, b) in enumerate(a)
+        if upper && UInt8('a') <= b <= UInt8('z')
+            b += toUpper
+        elseif !upper && UInt8('A') <= b <= UInt8('Z')
+            b -= toUpper
+        end
+        @inbounds a[i] = b
+        upper = b == UInt8('-')
+    end
+
+    return String(a)
+end
+
+"""
+Internal type used to enable fast merge, copy, and filter of Headers
+"""
+immutable AlreadyCanonical
+    d::Dict{String,String}
+end
+
+"""
+    Headers() <: Associative{String,String}()
+
+A type to represent headers. Keys containing invalid characters are returned as is. Valid keys are canonicalized, uppercasing the first character and characters following a '-' and lowercasing other characters. So, for valid keys, Headers is case-insensitive.
+"""
+immutable Headers{K<:String,V<:String} <: Associative{K,V}
+    d::Dict{K,V}
+
+    function Headers{K,V}() where V where K
+        new(Dict{K,V}())
+    end
+    function Headers{K,V}(kv) where V where K
+        d = Dict{K,V}()
+        for (k, v) in kv
+            ck = canonicalHeaderKey(k)
+            if haskey(d, ck)
+                error("Key collision -- multiple keys canonize as '$ck'")
+            else
+                d[ck] = v
+            end
+        end
+        new(d)
+    end
+    function Headers{K,V}(c::AlreadyCanonical) where V where K
+        new(c.d)
+    end
+end
+Headers() = Headers{String,String}()
+Headers(x) = Headers{String,String}(x)
+Headers(x::Pair...) = Headers{String,String}(x)
+
+Base.convert(::Type{Headers}, x::Associative{String,String}) = Headers(x)
+
+Base.similar(h::Headers{K,V}) where {K,V} = Headers{K,V}()
+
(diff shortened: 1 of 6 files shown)


quinnj:
Hey @garborg, thanks for the PR! I'm on vacation for the next week, so sorry if I don't find time to do a proper review until next week. I'm definitely in favor of the idea of doing our own `Headers` type.

From a quick glance, I think it looks good. People can still pass in a Dict to the high-level user-API functions, right? I think I'd also prefer we just split it out into its own `headers.jl` file; `types.jl` is getting pretty hefty at this point. Would you split it out and include the multipart form code as well?

Great stuff!

garborg:
Thanks, @quinnj! Took a stab at addressing your feedback. Have a good vacation.

src/headers.jl
```diff
@@ -0,0 +1,246 @@
...
+immutable AlreadyCanonical
+    d::Dict{String,String}
+end
+
+"""
+    Headers() <: Associative{String,String}()
+
+A type to represent headers. Keys containing invalid characters are returned as is. Valid keys are canonicalized, uppercasing the first character and characters following a '-' and lowercasing other characters. So, for valid keys, Headers is case-insensitive.
+"""
+immutable Headers{K<:String,V<:String} <: Associative{K,V}
```
garborg: SubString would be a little weird because canonicalization would replace arbitrary keys with new strings only to SubString them. I don't have any other string types in mind, but Id be happy to loosen the signature if anyone brought one.

garborg:
Re: using `canonicalize!(headers::Dict)`, here's what I was thinking:
First, peeking at go's implementation, what they did felt good to me. Those keys are case insensitive in the spec, so giving the header consumer the right answer even when they type a non-canonical key seems nice if not too expensive. The simpler option of exposing `canonicalize!` would save work on `getindex!`, but it also creates extra hashing, so bit of a wash there. Middleware operating on a `Dict{String,String}` would have to operate under an assumption about whether or not `canonicalize!` was called upstream, which seems problematic for writing middleware, at least if `canonicalize!` was meant to be called optionally in end-user-land. 

That said, I'm not a fan of how much code I added to get full feature parity with Dict, which is a maintenance concern. I'd be more comfortable writing code against `Headers` than against `Dict{String,String}`, personally, but I can't speak for anyone else.

Do you have reservations about the approach other than the complexity?

src/headers.jl
```diff
@@ -0,0 +1,246 @@
...
+immutable AlreadyCanonical
+    d::Dict{String,String}
+end
+
+"""
+    Headers <: Associative{String,String}
+
+A type to represent headers. Keys containing invalid characters are returned as is. Valid keys are canonicalized, uppercasing the first character and characters following a '-' and lowercasing other characters. So, for valid keys, Headers is case-insensitive.
+"""
+immutable Headers{K<:String,V<:String} <: Associative{String,String}
```
quinnj: I'd say we just take out the type parameters for now, just make it:
```julia
Headers <: Associative{String, String}
    d::Dict{String, String}
...
```

src/headers.jl
```diff
@@ -0,0 +1,246 @@
...
+        upper = b == UInt8('-')
+    end
+
+    return String(a)
+end
+
+"""
+Internal type used to enable fast merge, copy, and filter of Headers
+"""
+immutable AlreadyCanonical
```
quinnj: What do you think about getting rid of `AlreadyCanonical` and just having a boolean flag to the `Headers` constructor? It seems a little less heavy than needing another type.

src/headers.jl
```diff
@@ -0,0 +1,246 @@
...
+            return _chk(s)
+        end
+
+        upper = c == '-'
+    end
+
+    return s
+end
+
+const toUpper = 'A' - 'a'
```
quinnj: this is only used in `_chk`, so we can avoid the global and just put it at the top of the function definition.

src/headers.jl
```diff
@@ -0,0 +1,246 @@
...
+
+function _chk(s::String)
+    # confirm its fixable before changing anything
+    for c in s
+        if !validHeaderChar(c)
+            return s
+        end
+    end
+
+    a = copy(Array{UInt8}(s))
```
quinnj: I think it's a bit more common to say `Vector{UInt8}(s)`, since it's always a Vector.

src/headers.jl
```diff
@@ -0,0 +1,246 @@
...
+
+        upper = c == '-'
+    end
+
+    return s
+end
+
+const toUpper = 'A' - 'a'
+
+function _chk(s::String)
```
quinnj: maybe rename this to something more obvious, like `propercase`, or `canonicalize`?

src/headers.jl
```diff
@@ -0,0 +1,246 @@
+# See https://httpwg.github.io/specs/rfc7230.html#rule.token.separators
+const validHeaderBytes = IntSet(Int(c) for c in
+   raw"!#$%&'*+-.0123456789ABCDEFGHIJKLMNOPQRSTUWVXYZ^_`abcdefghijklmnopqrstuvwxyz|~"
+)
+
+validHeaderChar(c::Char) = isascii(c) && UInt8(c) in validHeaderBytes
```
quinnj: Maybe change it to `Int(c)` because it ends up getting converted to an Int anyway to check `in` w/ an `IntSet`?

src/headers.jl
quinnj: That should also simplify all the constructors you have here.

src/headers.jl
quinnj: I think having the boolean flag might also be nice because it provides an easier "escape hatch", if there are cases where the user needs to provide a header as-is (not exactly sure on the use-case here, but some kind of custom setup where the server only accepts specific header keys). It's a little easier to do `Header(d, alreadycanonical=true)` than fiddling w/ an HTTP internal type.

quinnj:
So I ended up pulling pieces out of this into commit https://github.com/JuliaWeb/HTTP.jl/commit/36b6143cb65686abe2b2b78327e95ca2dfb8ae22 (sorry my git-fu isn't strong enough to preserve authorship @garborg!). Here's the lightweight approach I took:
- Just have a single `canonicalize!` function that takes and changes a string directly
- Only apply `canonicalize!` in the parser code
  - this has the interesting property in that only Requests/Responses that are received from *_external_* sources get canonicalized
  - for the most part, I think this is what you want; sure, we're not sending out canonicalized headers, but we also don't have to deal w/ them
  - it allows users to create whatever headers they want when sending (doesn't mess w/ your stuff)

I figure we can try this out for now, but the idea occurred to me the other day and I figured it was worth trying it out for the simplicity.

samoconnor:
> doesn't mess w/ your stuff

👍🏻

garborg:
That sounds great @quinnj. I wasn't super happy with the data structur either, and tried a custom string type as an alternative, but that also wasn't a slam dunk.

If you're into using unexported stuff  -- this seemed a lot faster:
```julia
    a = Base.StringVector(s.len)
    copy!(a, s)
    @inbounds for (i, b) in enumerate(a)
```

PR #62, opened 2017-07-12
Escaping improvements for AWS compatibility.

samoconnor:
See http://docs.aws.amazon.com/general/latest/gr/signature-version-2.html

Sometimes the order of the query string matters, so `escape(Vector{Pair})` is handy.

### DIFF
--- a/src/uri.jl
+++ b/src/uri.jl
@@ -145,13 +145,13 @@ function Base.isvalid(uri::URI)
 end
 
 lower(c::UInt8) = c | 0x20
-@inline shouldencode(c) = c < 0x1f || c > 0x7f || @anyeq(c,
-                            UInt8(';'), UInt8('/'), UInt8('?'), UInt8(':'),
-                            UInt8('@'), UInt8('='), UInt8('&'), UInt8(' '),
-                            UInt8('"'), UInt8('<'), UInt8('>'), UInt8('#'),
-                            UInt8('%'), UInt8('{'), UInt8('}'), UInt8('|'),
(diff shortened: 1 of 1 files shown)


src/uri.jl
```diff
@@ -145,13 +145,13 @@ function Base.isvalid(uri::URI)
...
-                            UInt8('%'), UInt8('{'), UInt8('}'), UInt8('|'),
-                            UInt8('\\'), UInt8('^'), UInt8('~'), UInt8('['),
-                            UInt8(']'), UInt8('`'))
+# RFC3986 Unreserved Characters (and '~' Unsafe per RFC1738).
+@inline shouldencode(c) = !((c >= UInt8('A') && c <= UInt8('Z'))
+                         || (c >= UInt8('a') && c <= UInt8('z'))
+                         || (c >= UInt8('0') && c <= UInt8('9'))
+                         || c == UInt8('-')
+                         || c == UInt8('.')
+                         || c == UInt8('_'))
```
quinnj: So AWS is more permissive in what they allow in their query parameter keys/values? The link you shared was a bit unwieldy; are we just encoding too much right now?

samoconnor: Looking at this again, maybe this would be simpler:

```julia
shouldencode(c::Char) = !((isascii(c) && isalnum(c)) || c in "-._")
```

or perhaps:

```julia
const urlsafe = [(isascii(c) && isalnum(c)) || c in "-._" for c in Vector{Char}(0x00:0xFF)]
shouldencode(c) = !urlsafe[c+1]
```

quinnj: Looks great, thanks.

PR #66, opened 2017-07-28
Avoid early EOF

### LINKED ISSUE #45: Early EOF with streaming

quinnj:
Fix #45. When we get a Content-Length header, set the resulting FIFOBuffer's expectedlength field to ensure we don't get an early eof in cases where we're streaming responses.

cc: @omus 

This is mostly the same PR as #49, but I wanted to try to implement from scratch to ensure I fully understood the issues. I made a few changes, like when we set the `expectedlength` field (in the parser itself, we track the content-length header, so we piggy back on that).

Overall I'm on board with this, and my original hesitation of "ruining the FIFOBuffer" type by tying it too close to HTTP is assuaged; this is a natural operation to expect w/ a type like FIFOBuffer where things can be read/written simultaneously. You want a way to "set" an expected length.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -866,6 +866,7 @@ function parse!(r::Union{Request, Response}, parser, bytes, len=length(bytes);
                             @err(HPE_BODY_OVERFLOW)
                         end
                         parser.content_length = t
+                        r.body.expectedlength = Int64(t)
                      end
 
                 #= Transfer-Encoding: chunked =#
--- a/src/fifobuffer.jl
+++ b/src/fifobuffer.jl
@@ -51,6 +51,8 @@ mutable struct FIFOBuffer <: IO
     nb::Int64  # number of bytes available to read in buffer
     f::Int64   # buffer index that should be read next, unless nb == 0, then buffer is empty
     l::Int64   # buffer index that should be written to next, unless nb == len, then buffer is full
+    expectedlength::Int64 # when a known # of bytes will pass through the buffer (i.e. known Content-Length header)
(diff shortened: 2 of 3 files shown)


src/parser.jl
```diff
@@ -866,6 +866,7 @@ function parse!(r::Union{Request, Response}, parser, bytes, len=length(bytes);
                             @err(HPE_BODY_OVERFLOW)
                         end
                         parser.content_length = t
+                        r.body.expectedlength = Int64(t)
```
omus: Probably should be named `expected_length` to match styling of parser above.

quinnj: ew, underscores. :) The `content_length` for the parser was based on the port from the libhttpparser C library.

iamed2: Underscores help non-English speakers parse out words as the separation may not be intuitive

PR #68, opened 2017-08-02
Add an abstract type Handler, and Routing capabilties

quinnj:
A Server uses a `Handler` to handle requests/responses. Includes implementations for `HandlerFunction` which just wraps a function of the form `f(request, response) => Response`, and `Router`, which can register mappings for method/scheme/host/paths to specific handlers.

This is a pretty decent baseline for routing capabilities. I took a decent amount of inspiration from goLang's basic routing functionalities, and sprinkled some good julia dispatch on the problem and I like how simple the implementation is.

I need to add some more docs here too before merging.

### DIFF
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -31,6 +31,8 @@ include("multipart.jl")
 include("types.jl")
 include("parser.jl")
 include("client.jl")
+include("handlers.jl")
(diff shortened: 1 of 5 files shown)


quinnj:
cc: @sarvjeetsinghghotra

PR #77, opened 2017-08-23
Overhaul of client-side code for more robust error-handling

quinnj:


src/utils.jl
```diff
@@ -13,6 +13,35 @@ function escapeHTML(i::String)
...
+
+macro retry(N, expr)
+    :(@retryif Any $N $(esc(expr)))
+end
+
+macro retryif(cond, expr)
+    :(@retryif $(esc(cond)) 2 $(esc(expr)))
+end
+
+macro retryif(cond, N, expr)
```
samoconnor: Hi @quinnj 
I'm a fan of macros to make it harder to get retries catch wrong!
Do you have any interest in making a PR against [Retry.jl](https://github.com/samoconnor/Retry.jl) to add these retry macros there?
It would be a good thing to have more consistent utilities for this stuff across packages.
(Some day some of this should be in `Base` to encourage everyone to do error handling better).
I'd be happy to move `Retry.jl` to JuliaWeb to make collaboration easier.

src/utils.jl
```diff
@@ -13,6 +13,35 @@ function escapeHTML(i::String)
...
+    quote
+        local __r__
+        for i = 1:$N
+            try
+                __r__ = $(esc(expr))
+                break
+            catch e
+                typeof(e) <: $(esc(cond)) || rethrow(e)
+                i == $N && rethrow(e)
+                sleep(0.1)
```
samoconnor: You might want to consider exponential back-off and random jitter in your delay
https://github.com/samoconnor/Retry.jl/blob/master/src/repeat_try.jl#L178

src/utils.jl
quinnj: I definitely took inspiration from Retry.jl for these macros. Yeah, I think they'd live well over there. I figure we can let the dust settle on all this new error-handling logic to make sure everything goes smoothly.

src/utils.jl
quinnj: Good idea!

Issue #78, opened 2017-08-23
MethodError: no method matching escape(::Array{Pair{String,String},1})


samoconnor:
Reported by @jingpengw https://github.com/JuliaCloud/AWSSDK.jl/issues/1

```julia
julia> using Requests ; Requests.format_query_str(["Foo" => "Bar"])
"Foo=Bar"

julia> using HTTP; HTTP.escape(Dict(["Foo" => "Bar"]))
"Foo=Bar"

using HTTP; HTTP.escape(["Foo" => "Bar"])
ERROR: MethodError: no method matching escape(::Array{Pair{String,String},1})
```

samoconnor:
Hi @quinnj,

The fix below allows `escape(x)` to work for any iterable collection of pairs/tuples.
 
```julia
julia> HTTP.escape([("foo", "bar"), (1, 2)])
"foo=bar&1=2"

julia> HTTP.escape(Dict(["foo" => "bar", 1 => 2]))
"1=2&foo=bar"

julia> HTTP.escape(["foo" => "bar", 1 => 2])
"foo=bar&1=2"
```

```patch
@@ -185,12 +185,11 @@ function escape(io, k, A::Vector{String})
 end

 escape(p::Pair) = escape([p])
-function escape(d::Union{Dict,Vector{Pair}})
+function escape(d)
     io = IOBuffer()
-    len = length(d)
     for (i, (k,v)) in enumerate(d)
+        i == 1 || write(io, "&")
         escape(io, k, v)
-        i == len || write(io, "&")
     end
     return String(take!(io))
 end
```

The alternative to allowing any type for `d` in `escape(d)` would be to allow something like `Union{Associative, Vector{Pair{A,B}}, Vector{Tuple{A,B}}, ... others???}`. It seems best to allow anything. What do you think?

quinnj:
Thanks for the fix!

PR #82, opened 2017-08-24
Simplify HTTP.escape()

samoconnor:
Less code more type flexibility.

Following up on https://github.com/JuliaWeb/HTTP.jl/pull/62#discussion_r127386028 and https://github.com/JuliaWeb/HTTP.jl/issues/78#issuecomment-324493216

### DIFF
--- a/src/uri.jl
+++ b/src/uri.jl
@@ -153,7 +153,10 @@ function Base.isvalid(uri::URI)
 end
 
 # RFC3986 Unreserved Characters (and '~' Unsafe per RFC1738).
-@inline isunsafe(c::Char) = !(c in "-._") && (!isascii(c) || !isalnum(c))
+@inline issafe(c::Char) = c == '-' ||
+                          c == '.' ||
+                          c == '_' ||
+                          (isascii(c) && isalnum(c))
 
 utf8_chars(str::AbstractString) = (Char(c) for c in Vector{UInt8}(str))
 
@@ -161,8 +164,8 @@ utf8_chars(str::AbstractString) = (Char(c) for c in Vector{UInt8}(str))
 function escape end
 
 escape(c::Char) = string('%', uppercase(hex(c,2)))
-escape(str::AbstractString, check::Function=isunsafe) = 
-    join(check(c) ? escape(c) : c for c in utf8_chars(str))
+escape(str::AbstractString, safe::Function=issafe) = 
+    join(safe(c) ? c : escape(c) for c in utf8_chars(str))
 
 escape(bytes::Vector{UInt8}) = bytes
 escape(v::Number) = escape(string(v))
--- a/src/uri.jl
+++ b/src/uri.jl
@@ -153,46 +153,24 @@ function Base.isvalid(uri::URI)
 end
 
 # RFC3986 Unreserved Characters (and '~' Unsafe per RFC1738).
-@inline shouldencode(c) = !((c >= UInt8('A') && c <= UInt8('Z'))
-                         || (c >= UInt8('a') && c <= UInt8('z'))
-                         || (c >= UInt8('0') && c <= UInt8('9'))
-                         || c == UInt8('-')
-                         || c == UInt8('.')
-                         || c == UInt8('_'))
-hexstring(x) = string('%', uppercase(hex(x,2)))
+@inline isunsafe(c::Char) = !(c in "-._") && (!isascii(c) || !isalnum(c))
+
+utf8_chars(str::AbstractString) = (Char(c) for c in Vector{UInt8}(str))
 
 "percent-encode a string, dict, or pair for a uri"
 function escape end
 
-escape(v::Number) = string(v)
-escape(v::Nullable{T}) where {T} = Base.isnull(v) ? "" : string(get(v))
-function escape(str::AbstractString, f=shouldencode)
-    out = IOBuffer()
-    for c in Vector{UInt8}(str)
-        write(out, f(c) ? hexstring(Int(c)) : c)
-    end
-    return String(take!(out))
-end
+escape(c::Char) = string('%', uppercase(hex(c,2)))
+escape(str::AbstractString, check::Function=isunsafe) = 
+    join(check(c) ? escape(c) : c for c in utf8_chars(str))
+
 escape(bytes::Vector{UInt8}) = bytes
+escape(v::Number) = escape(string(v))
+escape(v::Nullable) = Base.isnull(v) ? "" : escape(get(v))
 
-escape(io, k, v) = write(io, escape(k), "=", escape(v))
-function escape(io, k, A::Vector{String})
-    len = length(A)
-    for (i, v) in enumerate(A)
-        write(io, escape(k), "=", escape(v))
-        i == len || write(io, "&")
-    end
-end
-
-escape(p::Pair) = escape([p])
-function escape(d)
-    io = IOBuffer()
-    for (i, (k,v)) in enumerate(d)
-        i == 1 || write(io, "&")
-        escape(io, k, v)
-    end
-    return String(take!(io))
-end
+escape(key, value) = string(escape(key), "=", escape(value))
+escape(key, values::Vector) = escape(key => v for v in values)
+escape(query) = join((escape(k, v) for (k,v) in query), "&")
 
 "unescape a percent-encoded uri/url"
 function unescape(str)
--- a/test/uri.jl
+++ b/test/uri.jl
@@ -46,7 +46,7 @@ URLTest(nm::String, url::String, isconnect::Bool, shouldthrow::Bool) = URLTest(n
 
     @test isempty(HTTP.URIs.Offset())
     @test HTTP.lower(UInt8('A')) == UInt8('a')
-    @test HTTP.URIs.hexstring(1) == "%01"
+    @test HTTP.escape(Char(1)) == "%01"
 
     @test HTTP.escape(Dict("key1"=>"value1", "key2"=>["value2", "value3"])) == "key2=value2&key2=value3&key1=value1"
 


src/uri.jl
```diff
@@ -153,46 +153,24 @@ function Base.isvalid(uri::URI)
...
 
 # RFC3986 Unreserved Characters (and '~' Unsafe per RFC1738).
-@inline shouldencode(c) = !((c >= UInt8('A') && c <= UInt8('Z'))
-                         || (c >= UInt8('a') && c <= UInt8('z'))
-                         || (c >= UInt8('0') && c <= UInt8('9'))
-                         || c == UInt8('-')
-                         || c == UInt8('.')
-                         || c == UInt8('_'))
-hexstring(x) = string('%', uppercase(hex(x,2)))
+@inline isunsafe(c::Char) = !(c in "-._") && (!isascii(c) || !isalnum(c))
```
quinnj: No `in`s please; they're way too slow vs. explicit `x == y || ...` comparisons

samoconnor: Do you have a performance test for this stuff?
If not, perhaps if you could provide a use-case example that captures your performance concerns I could add a performance test and ensure that I'm not causing significant regressions.

src/uri.jl
```diff
@@ -153,46 +153,24 @@ function Base.isvalid(uri::URI)
...
-function escape(io, k, A::Vector{String})
-    len = length(A)
-    for (i, v) in enumerate(A)
-        write(io, escape(k), "=", escape(v))
-        i == len || write(io, "&")
-    end
-end
+escape(bytes::Vector{UInt8}) = bytes
+escape(v::Number) = escape(string(v))
+escape(v::Nullable) = Base.isnull(v) ? "" : escape(get(v))
```
quinnj: good catch on this one; we weren't escaping nullables before.

src/uri.jl
```diff
@@ -153,46 +153,24 @@ function Base.isvalid(uri::URI)
...
 
-escape(io, k, v) = write(io, escape(k), "=", escape(v))
-function escape(io, k, A::Vector{String})
-    len = length(A)
-    for (i, v) in enumerate(A)
-        write(io, escape(k), "=", escape(v))
-        i == len || write(io, "&")
-    end
-end
+escape(bytes::Vector{UInt8}) = bytes
```
quinnj: I'm not sure about this one; I think we should escape raw bytes.

samoconnor: (Note: the raw bytes were not escaped before this PR either.)
I assumed that it was intended as a way to pass in bytes that are not to be escaped. e.g. if you have some pre-escaped data. Other than that, I can't see a use-case for passing raw bytes as part of a URL or query string.

src/uri.jl
quinnj:
```julia
julia> function f(c, N)
           s = 0
           for i = 1:N
               if c in "-._"
                   s += 1
               end
           end
           return s
       end
f (generic function with 1 method)

julia> function f2(c, N)
           s = 0
           for i = 1:N
               if c == '-' || c == '.' || c == '_'
                   s += 1
               end
           end
           return s
       end
f2 (generic function with 1 method)

julia> using BenchmarkTools

julia> b1 = @benchmarkable f('_', 1000000)
Benchmark(evals=1, seconds=5.0, samples=10000)

julia> b2 = @benchmarkable f2('_', 1000000)
Benchmark(evals=1, seconds=5.0, samples=10000)

julia> tune!(b1); tune!(b2)
Benchmark(evals=1000, seconds=5.0, samples=10000)

julia> run(b1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     12.349 ms (0.00% GC)
  median time:      13.523 ms (0.00% GC)
  mean time:        13.676 ms (0.00% GC)
  maximum time:     17.001 ms (0.00% GC)
  --------------
  samples:          366
  evals/sample:     1

julia> run(b2)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.376 ns (0.00% GC)
  median time:      1.648 ns (0.00% GC)
  mean time:        1.888 ns (0.00% GC)
  maximum time:     35.564 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

quinnj: Over 1000x slower :)

src/uri.jl
quinnj: Hmmm, yeah, that seems sensible. Sounds fine to me then.

src/uri.jl
samoconnor: Only ~10% slower in context:

```julia
julia> @inline shouldencode(c) = !((c >= UInt8('A') && c <= UInt8('Z'))
                                 || (c >= UInt8('a') && c <= UInt8('z'))
                                 || (c >= UInt8('0') && c <= UInt8('9'))
                                 || c == UInt8('-')
                                 || c == UInt8('.')
                                 || c == UInt8('_'))

julia> @inline isunsafe(c::Char) = !(c in "-._") && (!isascii(c) || !isalnum(c))

julia> utf8_chars(str::AbstractString) = (Char(c) for c in Vector{UInt8}(str))

julia> escape(c::Char) = string('%', uppercase(hex(c,2)))

julia> newescape(str::AbstractString, check::Function=isunsafe) =
           join(check(c) ? escape(c) : c for c in utf8_chars(str))

julia> oldescape(str::AbstractString, check::Function=shouldencode) =
           join(check(Int(c)) ? escape(c) : c for c in utf8_chars(str))

julia> d = String(Vector{UInt8}(rand(0:255,1000000)))
1000000-element Array{UInt8,1}

julia> bnew = @benchmarkable newescape(d)

julia> bold = @benchmarkable oldescape(d)

julia> run(bold)
BenchmarkTools.Trial:
  memory estimate:  281.08 MiB
  allocs estimate:  5966566
  --------------
  minimum time:     225.803 ms (11.35% GC)
  median time:      234.066 ms (11.58% GC)
  mean time:        248.353 ms (17.21% GC)
  maximum time:     319.342 ms (36.01% GC)
  --------------
  samples:          21
  evals/sample:     1

julia> run(bnew)
BenchmarkTools.Trial:
  memory estimate:  281.08 MiB
  allocs estimate:  5966566
  --------------
  minimum time:     246.922 ms (10.48% GC)
  median time:      256.928 ms (11.49% GC)
  mean time:        268.947 ms (15.56% GC)
  maximum time:     342.289 ms (33.90% GC)
  --------------
  samples:          19
  evals/sample:     1

julia> d = Vector{UInt8}(rand(0:255,10))
10-element Array{UInt8,1}

julia> run(bold)
BenchmarkTools.Trial:
  memory estimate:  3.13 KiB
  allocs estimate:  69
  --------------
  minimum time:     2.474 μs (0.00% GC)
  median time:      2.602 μs (0.00% GC)
  mean time:        2.932 μs (8.06% GC)
  maximum time:     2.427 ms (97.39% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> run(bnew)
BenchmarkTools.Trial:
  memory estimate:  3.13 KiB
  allocs estimate:  69
  --------------
  minimum time:     2.701 μs (0.00% GC)
  median time:      2.825 μs (0.00% GC)
  mean time:        3.198 μs (7.53% GC)
  maximum time:     2.478 ms (97.23% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> d = Vector{UInt8}(rand(0:255,1000))
1000-element Array{UInt8,1}

julia> run(bold)
BenchmarkTools.Trial:
  memory estimate:  291.33 KiB
  allocs estimate:  6068
  --------------
  minimum time:     198.116 μs (0.00% GC)
  median time:      208.255 μs (0.00% GC)
  mean time:        247.330 μs (11.94% GC)
  maximum time:     4.300 ms (93.01% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> run(bnew)
BenchmarkTools.Trial:
  memory estimate:  291.33 KiB
  allocs estimate:  6068
  --------------
  minimum time:     220.771 μs (0.00% GC)
  median time:      230.028 μs (0.00% GC)
  mean time:        278.296 μs (11.06% GC)
  maximum time:     3.848 ms (89.57% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

samoconnor: A little tweak to make the `in` list `const` reduces the slowdown to ~5%:
```julia
const safe = Vector{UInt8}("-._")
@inline isunsafe(c) = !(UInt8(c) in safe) && (!isascii(c) || !isalnum(c))
```

samoconnor: Custom `isalnum()` to avoid ccalls into utf8 libraries reduces the slowdown to ~2%:
```julia
@inline isascii(c::UInt8) = c < UInt8(0x80)

@inline isalnum(c::UInt8) = ((c >= UInt8('A') && c <= UInt8('Z'))
                          || (c >= UInt8('a') && c <= UInt8('z'))
                          || (c >= UInt8('0') && c <= UInt8('9')))

const safe = Vector{UInt8}("-._")

@inline isunsafe(c::UInt8) = !(c in safe) && (!isascii(c) || !isalnum(c))
```

samoconnor: However, when I change the benchmark to escape a more real-world example of escaping a query string Dict() the slowdown is around 3-4% even without these tweaks.
So, in summary, the "No ins please; they're way too slow" does not seem to apply in this use case.
If the benchmark was to construct a URI, including escaping a query string, the slowdown would be on the order of 1-2%. If the benchmark was `GET`-ing a URI, even from localhost, the slowdown would be well and truely in the noise.

samoconnor:
@quinnj, see https://github.com/JuliaWeb/HTTP.jl/pull/82/commits/0bfeb3a51890531bc1fc5d97f4283925cc5fae02, `in` removal per your request.

quinnj:
Looks great, thanks!

PR #84, opened 2017-08-28
Add path and method based routing with variable names and values.

sarvjeetsinghghotra:
Provides routing based on:
- **Path**. Supports:
    1. Regular expressions
    2. Variables names, hence variable values can be passed through the path in the request
- **Method**: Allows multiple methods for a route - `["get", "post"]`
- It also allows setting `defaulthandler` when nothing matches
- The order in which the `route`s  are `register!` determines its priority when there is more than one match for a request.

It follows very similar interface/design to the `handler.jl` (ultimately to the mux lib of golang). Please refer routing.jl test file (included here) for better clarity of the implementation and design.

ToDo: Error handling (soon).

Coming up: Routing based on - **host**, **scheme**, **header**, **query** and **custom routing function** (Almost done, little debugging is going on).


sarvjeetsinghghotra:
PS: Can you please suggest what's failing in the test/server.jl due to the change ?

quinnj:
Hey @sarvjeetsinghghotra, here's a few thoughts:

- I can't say I'm a huge fan of throwing out all the existing routing code; it works well and is simple and small. It already supports method/path/scheme matching out of the box. I would much rather we *build upon* and *enhance* the existing code instead of throwing it all out for something that looks to be more complicated and probably slower
- On that note, it'd be nice to show the advantages of going with one approach over another; showing how code ends up being drastically simpler for the user, or significant performance gains, or much more powerful functionality are generally required to warrant such a huge code churn
- I'm not a huge fan of "The order in which the routes are register! determines its priority"; it sounds annoying for me to be debugging my server and realize that a route that was defined earlier was inadvertently stealing routes from something later. The advantage with the current code is that it's based on *_specificity_*, the most *_specific_* route defined will be used, with ambiguous cases throwing errors.

sarvjeetsinghghotra:
Hey @quinnj,

I agree with you, it's not a good idea to throw the complete code without comparing their merits and drawbacks, sorry about that. We should totally compare the two implementations.

Currently, I am going through handler.jl code. Will get back to you with a comparison (as per my understanding/knowledge).

PR #88, opened 2017-08-31
Warning for new users about precompile problems.

samoconnor:
Hi @quinnj,

It might be a good idea to say something on README.md about the start-up delay.
For some of my HTTP.jl use-cases it's a show-stopper, but for others, it makes no difference at all because there are other packages that already take a long time to start up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,10 @@ julia> Pkg.add("HTTP")
 
 The package is tested against Julia 0.6 & current master on Linux, OS X, and Windows.
 
+## Known Limitations
+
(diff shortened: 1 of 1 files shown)


quinnj:
Just pushed fixes for precompiles; I think we're in good shape now, definitely useable in normal applications. Things will improve further once we're able to precompile things for other modules (ref [here](https://github.com/JuliaLang/julia/issues/23548)).

samoconnor:
See performance comparison note: https://github.com/JuliaWeb/HTTP.jl/issues/58#issuecomment-339852609

PR #106, opened 2017-11-01
add escaping for symbols

KristofferC:
Fixes an error when using symbols as keys in Dicts passed to HTTP.

### DIFF
--- a/src/uri.jl
+++ b/src/uri.jl
@@ -173,6 +173,7 @@ escape(str::AbstractString, safe::Function=issafe) =
 
 escape(bytes::Vector{UInt8}) = bytes
 escape(v::Number) = escape(string(v))
+escape(v::Symbol) = escape(string(v))
 escape(v::Nullable) = Base.isnull(v) ? "" : escape(get(v))
 
 escape(key, value) = string(escape(key), "=", escape(value))


src/uri.jl
```diff
@@ -173,6 +173,7 @@ escape(str::AbstractString, safe::Function=issafe) =
 
 escape(bytes::Vector{UInt8}) = bytes
 escape(v::Number) = escape(string(v))
+escape(v::Symbol) = escape(string(v))
```
samoconnor: @quinnj, @KristofferC what do you think about just providing `escape(v) = escape(string(v))` ?

KristofferC: Yeah, I thought about the same thing just now. I think it makes sense.

samoconnor: One of my users was expecting `escape(::DateTime)` to work  https://github.com/JuliaCloud/AWSSDK.jl/issues/4 .
@quinnj what do you think of adding `escape(v) = escape(string(v))` ?

quinnj: Sure.

samoconnor: ... ok, so now that I've look at "just doing it", there is a hitch:
`escape(query)` is already defined to work Dicts, Vectors{Pair} etc, anything that acts like an iterator over pairs (`... for (k,v) in query`).
Best to leave it as is until Julia has some kind of "protocols" feature that allows dispatching on "dict-like or sequence of pair-like thing".

quinnj: Let's add a case for `Dates.TimeType` for now then.

Issue #63, opened 2017-07-14
Migrating from Requests.jl to HTTP.jl, experience and suggestions...


samoconnor:
Hi @quinnj,

I read (your announcement in Feb)[https://discourse.julialang.org/t/ann-new-http-jl-package/2048] with interest. My AWSCore.jl package relies on Requests.jl and I found the experience of getting it to work reliably was less than ideal, so the idea of a cleaner HTTP implementation appeals to me. 

I've just finished an experimental port to HTTP.jl. On the whole, it looks like a big improvement. The new HTTP API pretty much just does what I want with no fuss. Compare my [new much simpler http.jl](https://github.com/samoconnor/AWSCore.jl/blob/http_jl_branch/src/http.jl)   to my [old http.jl](https://github.com/samoconnor/AWSCore.jl/blob/master/src/http.jl).

Note that the [AWSCore.jl/src/http.jl layer](https://github.com/samoconnor/AWSCore.jl/blob/http_jl_branch/src/http.jl) does not disappear completely, it still adds the following features to HTTP.jl:

 - `HTTPException` type for non `20x` result status.
 - Retry with exponential back-off  for status `< 200` or `>= 500` and `UVError`.

Would you consider a PR to introduce a `HTTPException` type to HTTP.jl?

Would you consider a PR to add exponential back-off using [Retry.jl](https://github.com/samoconnor/Retry.jl) ?

My other suggestion is to consider separating some of HTTP.jl's functionality into layers (seperate files and distinct APIs). HTTP.jl currently implements both stateless low-level HTTP wire-protocol stuff and high-level stateful stuff (e.g. Cookies, Authentication, Retries, Redirect processing). It's great to have all this functionality, but it would be cleaner to factor it out a little.

I suggest something like this:

Low level request/response layer:
 - Just send a request and read the response.
 - No default headers, just encode and send headers and body passed by the caller. 

Middle retry layer:
 - Retry with exponential back-off  for status `< 200` or `>= 500` and low level IO Exceptions.
 - Throw an exception for non `20x` status.

High level stateful layer:
 - Useful default headers (User-Agent, Accept etc)
 - Authentication
 - Redirect handling (e.g. handle issues like authentication signatures changing on redirect).
 - Transfer-Encoding?
 - Cookies

A library like AWSCore.jl would use the middle layer directly (rather than using the high-level API with a bunch of options to disable default behaviour as it does now).

quinnj:
[...]

PR #107, opened 2017-11-01
Fix throw(nothing) from HTTP.request()

samoconnor:
`processresponse!()` was returning `nothing` instead of `StatusError`.

This was difficult to track down because the `nothing` exception value was used inside `Retry.jl` as a flag to mean the exception should be ignored/retried.
I've updated `Retry.jl` so that it does not catch `nothing` exceptions : https://github.com/samoconnor/Retry.jl/commit/a98715e6a3e9ba87283e4d6336f819bcb834b9aa

### DIFF
--- a/src/client.jl
+++ b/src/client.jl
@@ -301,11 +301,13 @@ function processresponse!(client, conn, response, host, method, maintask, stream
         elseif stream && headerscomplete
             @log "processing the rest of response asynchronously"
             response.body.task = @async processresponse!(client, conn, response, host, method, maintask, false, tm, canonicalizeheaders, false)
-            return true, nothing
+            return true, StatusError(status(response), response)
         end
     end
     # shouldn't ever reach here
     dead!(conn)
+    @assert false
+    # FIXME Can't return ::Bool without ::Exception
     return false
 end
 
@@ -327,6 +329,7 @@ function request(client::Client, req::Request, opts::RequestOptions, stream::Boo
     reset!(client.parser)
     success, err = processresponse!(client, conn, response, host, HTTP.method(req), current_task(), stream, opts.readtimeout::Float64, opts.canonicalizeheaders::Bool, verbose)
     if !success
+        @assert err != nothing
         retry >= opts.retries::Int && throw(err)
         return request(client, req, opts, stream, history, retry + 1, verbose)
     end
@@ -336,9 +339,11 @@ function request(client::Client, req::Request, opts::RequestOptions, stream::Boo
     if opts.allowredirects::Bool && req.method != HEAD && (300 <= status(response) < 400)
         return redirect(response, client, req, opts, stream, history, retry, verbose)
     end
-    if success && ((200 <= status(response) < 300) || !opts.statusraise::Bool)
+    @assert success
+    if (200 <= status(response) < 300) || !opts.statusraise::Bool
         return response
     else
+        @assert err != nothing
         retry >= opts.retries::Int && throw(err)
         return request(client, req, opts, stream, history, retry + 1, verbose)
     end


src/client.jl
```diff
@@ -301,11 +301,13 @@ function processresponse!(client, conn, response, host, method, maintask, stream
         elseif stream && headerscomplete
             @log "processing the rest of response asynchronously"
             response.body.task = @async processresponse!(client, conn, response, host, method, maintask, false, tm, canonicalizeheaders, false)
-            return true, nothing
+            return true, StatusError(status(response), response)
```
quinnj: I don't think I understand the issue. Why were you ever catching a `nothing` exception? In this code path, we're returning `true` for successful, so there's no error?

samoconnor:
`status, err = (true, nothing)` is returned [from here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L304), [to here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L328).
But then, [down here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L342), depending on the status code, we do`throw(err)`, which is in this case `nothing`.

Note that [up here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L300) we seem to correctly return `(true, StatusError(...))`.

src/client.jl
samoconnor: `status, err = (true, nothing)` is returned [from here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L304), [to here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L328).
But then, [down here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L342), depending on the status code, we do`throw(err)`, which is in this case `nothing`.

Note that [up here](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/client.jl#L300) we seem to correctly return `(true, StatusError(...))`.

(Please excuse double post)

quinnj:
Ah yes, I see now. kk, I actually already have a fix for this locally. Just running tests now and I'll push.

samoconnor:
Ok, if you're going to push a seperate change rather than merging this PR, please also consider:

- [ ] this https://github.com/JuliaWeb/HTTP.jl/pull/107/files#diff-29a732d52d0661ba454242753a687486R310
- [ ] and this https://github.com/JuliaWeb/HTTP.jl/pull/107/files#diff-29a732d52d0661ba454242753a687486R343
 - [ ] and the other asserts.

src/client.jl
samoconnor: It feels to me that the problem stems from code that implements a bunch of features in one place rather than separating them into distinct layers. (i.e. is seems wrong for `processresponse!()` to have knowledge of the possibility that the caller will need an exception object for a non-error result because the caller considers some status codes to be errors.

I'd like to see a seperate basic HTTP layer that has:
 - No default header values,
 - No interpretation of status codes,
 - No status code exceptions
 - No retries,
 - No redirects,
 - No cookies,
 - No authentication,

i.e there should be a layer that just sends a request and parses the response.

I think if that was done, then the higher order functions could be cleanly layered on top and the probability of pitfalls like this one would be greatly reduced. In fact each of the features listed above could be implemented as a seperate module, configurably, chained together, each with a `processresponse!()` function.

The user-level API could still enable a useful default set of high level features, nothing needs to change there.

See also [previous notes on layering](https://github.com/JuliaWeb/HTTP.jl/issues/63#issue-242916708)

quinnj: Great thoughts, but I'm just not sure it's really that practical. That's a great list for a "basic HTTP layer", but it's also a bit arbitrary. Why include error-handling at all? What about re-using open connections? Logging? Connect retries?

The point I'm getting at is the complete feature list that `client.jl` currently provides is even much longer than your list above, so where exactly do you draw the line in terms of "layers"? That seems like a lot of work for not a lot of gain to me. Right now, things are very configurable via the `RequestOptions` object; you can flip switches on and off as needed with negligible impact on performance and keeps all the code in one place, fairly easy to follow. I certainly looked into breaking things up into layers, but it just seemed like a bigger design challenge than is really worth once you get into the weeds.

samoconnor: I would remove everything from the lowest layer that can be removed while leaving the ability to send a single request (header + body) and receive/parse a single response into header + body. This layer would assume that nothing goes wrong with the network. Dealing with network exceptions and deciding on retries vs rethrowing vs  throwing custom error types would be dealt with by higher layers.

Yes, absolutely retries and connection reuse I would implement as seperate layers. I would keep the layers thin and independently unit-testable. I would not combine multiple features into a layer unless they were unavoidably interdependent.

Logging I see as a service plane orthogonal to the layer stack, any layer should be able to log.

I've worked on a number of high performance network stacks that were rigorously factorised into layers. I don't think a layered approach implies a performance penalty. In fact the ability to configurably omit some layers completely can be a performance gain.

At the end of the day, any design that functions correctly and is easy test, easy to understand and easy to maintain is a good design. However, this particular bug smells to me like a sign that this design might benefit from more separation into layers and better defined interfaces between layers.

PR #116, opened 2017-11-15
Fix #113 and add Inf timeout option

### LINKED ISSUE #113: timeout macro returns exception value instead of rethrowing
The [`timeout` macro](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/utils.jl#L52) executes `expr` in a new task, waits for a timeout or `istaskdone`, then returns `task.result`. In the case where `expr` throws and exception, the exception is returned as a value when it should be re-thrown.

The result is:
```
ERROR: MethodError: no method matching length(::Base.UVError)
 [1] processresponse!(::HTTP.Client, ::HTTP.Connection{MbedTLS.SSLContext}, ::HTTP.Response, ::String, ::HTTP.Method, ::Task, ::Bool, ::Float64, ::Bool, ::Bool)
at /Users/sam/.julia/v0.6/HTTP/src/client.jl:287
```

This patch uses `wait()` to return the task result, or rethrow the exception if needed.
```patch
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -58,7 +58,7 @@ macro timeout(t, expr, then, pollint=0.01)
             sleep($pollint)
         end
         istaskdone(tsk) || $(esc(then))
-        tsk.result
+        wait(tsk)
     end
 end
```

Also, does `macro timeout` have a 10ms sleep penalty for every call?
Maybe this would be better, so that if `expr` returns quickly there is no timeout:
```patch
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -54,6 +54,7 @@ macro timeout(t, expr, then, pollint=0.01)
         tm = Float64($(esc(t)))
         start = time()
         tsk = @async $(esc(expr))
+        yield(tsk)
         while !istaskdone(tsk) && (time() - start < tm)
             sleep($pollint)
         end
```

samoconnor:


src/utils.jl
```diff
@@ -51,14 +51,19 @@ checking if `expr` has finished executing (short for polling interval).
...
-        tsk = @async $(esc(expr))
-        while !istaskdone(tsk) && (time() - start < tm)
-            sleep($pollint)
+        if $(esc(t)) == Inf
+            $(esc(expr))
+        else
+            tm = Float64($(esc(t)))
+            start = time()
+            tsk = @async $(esc(expr))
+            yield()
```
quinnj: Why the `yield` here?

src/utils.jl
```diff
@@ -51,14 +51,19 @@ checking if `expr` has finished executing (short for polling interval).
...
+            while !istaskdone(tsk) && (time() - start < tm)
+                sleep($pollint)
+            end
+            istaskdone(tsk) || $(esc(then))
+            wait(tsk)
```
quinnj: Can you remind me why the change to `wait` here?

src/utils.jl
samoconnor: Per comment in #113, my concern is that `istaskdone` will return `false` and cause a 10ms sleep, even if `expr` would have executed in a very short time. The `yield()` gives the `expr` task a chance to run before checking `istaskdone`.
e.g.
```julia
function f()
       t = @async sum([1,2,3])
       !istaskdone(t) && sleep(0.01)
end
julia> f()
julia> @time f()
  0.013942 seconds (12 allocations: 1.031 KiB)
```
vs
```julia
function f()
       t = @async sum([1,2,3])
       yield()
       !istaskdone(t) && sleep(0.01)
end
julia> f()
julia> @time f()
0.000011 seconds (8 allocations: 880 bytes)
```

src/utils.jl
samoconnor: Because `wait` rethrows the task exception if needed, whereas the current code ends up at `no method matching length(::Base.UVError)`

From #113 ...

The [`timeout` macro](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/utils.jl#L52) executes `expr` in a new task, waits for a timeout or `istaskdone`, then returns `task.result`. In the case where `expr` throws and exception, the exception is returned as a value when it should be re-thrown.

The result is:
```
ERROR: MethodError: no method matching length(::Base.UVError)
 [1] processresponse!(::HTTP.Client, ::HTTP.Connection{MbedTLS.SSLContext}, ::HTTP.Response, ::String, ::HTTP.Method, ::Task, ::Bool, ::Float64, ::Bool, ::Bool)
at /Users/sam/.julia/v0.6/HTTP/src/client.jl:287
```

quinnj:
Thanks!

Issue #117, opened 2017-11-15
@async HTTP.get returns empty body.


samoconnor:
[ Update: See PR #119 ]
```julia
@sync for i in 1:3
    @async @show String(HTTP.get("http://octe.ch"))
end

String(HTTP.get("http://octe.ch")) = "<HEAD>\n<TITLE>octe.ch</TITLE> ... [snip] ... </BODY>\n"
String(HTTP.get("http://octe.ch")) = ""
String(HTTP.get("http://octe.ch")) = ""
```

I've also tried this with my branch where `HTTP.FIFOBuffer` is replaced by `Base.BufferStream`. The behaviour is the same. https://github.com/JuliaWeb/HTTP.jl/issues/110

_Thanks @jingpengw for reporting this https://github.com/samoconnor/AWSS3.jl/issues/19#issuecomment-344756111,
and contributing a test case https://github.com/samoconnor/AWSS3.jl/commit/5abbcaa150e4b5a60c5e8e48b2ee313d72c1ca06_

samoconnor:
It looks like there is non-task-safe global state in `HTTP.DEFAULT_CLIENT`.
Constructing a new `HTTP.Client` for each `get` works:
```julia
@sync for i in 1:3
           @async @show String(HTTP.get(HTTP.Client(), "http://octe.ch"))
       end
String(HTTP.get(HTTP.Client(), "http://octe.ch")) = "<HEAD>\n<TITLE>octe.ch</TITLE>... "
String(HTTP.get(HTTP.Client(), "http://octe.ch")) = "<HEAD>\n<TITLE>octe.ch</TITLE>... "
String(HTTP.get(HTTP.Client(), "http://octe.ch")) = "<HEAD>\n<TITLE>octe.ch</TITLE>... "
```

samoconnor:
The patch below puts a `ReentrantLock` around connection pool access and moves the `Parser` object from the `Client` struct to  the `Connection` struct.

i.e. Don't share a connection between tasks and don't share a parser between connections.

This works. Not sure if it is the best fix.

```patch
diff --git a/src/client.jl b/src/client.jl
index 7f774bb..e9d6b6c 100644
--- a/src/client.jl
+++ b/src/client.jl
@@ -11,10 +11,11 @@ mutable struct Connection{I <: IO}
     id::Int
     socket::I
     state::ConnectionState
+    parser::Parser
 end
 
-Connection(tcp::IO) = Connection(0, tcp, Busy)
[...]

PR #119, opened 2017-11-17
Don't share parser between tasks. Fix issue #117.

samoconnor:
See discussion in #117.
Don't share a connection between tasks and don't share a parser between connections.


quinnj:
Thanks, this all looks correct/good from what I can tell.

PR #124, opened 2017-11-22
Add test for parsing one charactar at a time, fix #125 and #126

### LINKED ISSUE #125: Main for loop never terminates for parse!(r, p, Vector{UInt8}("Some-Header:"))
The main `for` loop in `parse!()` never terminates when the last parser input is fragmented so that the last character in the input `bytes` is the `:` at the end of a header.

i.e. If the parser will lock up if the request data is fragmented so that the first call to `readavailable(socket)` returns `"HTTP/1.1 200 OK\r\nDate:"`.

PR https://github.com/JuliaWeb/HTTP.jl/pull/124 adds a test that parses 1 byte at a time to demonstrate the problem and adds a loop invariant `@assert` to the `for` loop to stop in looping forever.

```julia
julia> HTTP.parse(HTTP.Response, "HTTP/1.1 200 OK\r\nDate:")
ERROR: AssertionError: p > old_p
```

samoconnor:


samoconnor:
Updated PR with fix for #125

samoconnor:
updated with (failing) test for parsing requests 1 char at a time

samoconnor:
Updated with fix for #126

quinnj:
I'll need a few more days to review this; it touches several very low-level parts of the parser, so we want to be careful w/ performance and correctness.

samoconnor:
Thorough review sounds good.
Do you have a feel for whether #125 and #126 are inherited from the C code that you ported?, or if they are simply errors introduced while porting?

samoconnor:
#135

PR #132, opened 2017-12-04
Allow any Associative as a header

omus:
Allows alternative `Dict` types such as `PyCall.PyDict` to be used as a header.

### DIFF
--- a/src/client.jl
+++ b/src/client.jl
@@ -365,7 +365,7 @@ request(client::Client, req::Request;
 
 # build Request
 function request(client::Client, method, uri::URI;
-                 headers::Dict=Headers(),
+                 headers::Associative=Headers(),
                  body=FIFOBuffers.EMPTYBODY,
                  stream::Bool=false,
                  verbose::Bool=false,
@@ -408,7 +408,7 @@ an `IOBuffer` for in-memory data, or even an `HTTP.FIFOBuffer`. For complete con
 
 Additional keyword arguments supported, include:
 
-  * `headers::Dict`: headers given as Dict to be sent with the request
+  * `headers::Associative`: headers given as Dict to be sent with the request
   * `body`: a request body can be given as a `String`, `Vector{UInt8}`, `IO`, `HTTP.FIFOBuffer` or `Dict`; see example below for how to utilize `HTTP.FIFOBuffer` for "streaming" request bodies; a `Dict` argument will be converted to a multipart form upload
   * `stream::Bool=false`: enable response body streaming; depending on the response body size, the request will return before the full body has been received; as the response body is read, additional bytes will be recieved and put in the response body. Readers should read until `eof(response.body) == true`; see below for an example of response streaming
   * `chunksize::Int`: if a request body is larger than `chunksize`, the "chunked-transfer" http mechanism will be used and chunks will be sent no larger than `chunksize`; default = `nothing`
--- a/src/client.jl
+++ b/src/client.jl
@@ -324,7 +324,7 @@ function request(client::Client, req::Request, opts::RequestOptions, stream::Boo
     # maybe allow retrying for all kinds of errors?
     p = port(u)
     conn = @retryif ClosedError 4 connectandsend(client, sch, host, ifelse(p == "", "80", p), req, opts, verbose)
-    
+
     response = Response(stream ? 2^24 : FIFOBuffers.DEFAULT_MAX, req)
     reset!(conn.parser)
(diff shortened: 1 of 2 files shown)


omus:
I could add some tests explicitly for this but most likely I would need to add a test requirement on DataStructures.jl to use an `OrderedDict`

quinnj:
I wonder if we even leave off `Associative` here, as to allow for `Vector{Pair}`; then we could add tests w/ `Vector{Pair}` 😉

samoconnor:
I have a WIP experimental branch where I've replaced the internal representation of the headers with a `Vector{Pair}`. This seems to work well and makes dealing with special cases rules (like `Set-Cookie`) easier.
https://github.com/samoconnor/HTTP.jl/commit/bb120bdf0fb36f25496aaa7fb0504cfa4d2c62d4

samoconnor:
@omus, `mkheaders` is now: `mkheaders(h)::Headers = Header[string(k) => string(v) for (k,v) in h]`
Does this cover the intention of your PR?
https://github.com/JuliaWeb/HTTP.jl/blob/e5d43ccec733d8d4afa781af00f7b3a6f8c01ec8/src/Messages.jl#L193

omus:
Looks like this is still an issue:

https://github.com/JuliaWeb/HTTP.jl/blob/f9837eb563187559260a84cdc5d24aaca72b5bed/src/client.jl#L39-L45

samoconnor:
Sorry for the confusion, the `request(client::Client, ...` method is there for compatibility.
See commit b5844b13a2b19a4d546fdc76f44d2c0e9ee85e97

The new doc says:
https://github.com/JuliaWeb/HTTP.jl/blob/f9837eb563187559260a84cdc5d24aaca72b5bed/src/HTTP.jl#L41-L43

The top-level `request` method is now this:
https://github.com/JuliaWeb/HTTP.jl/blob/f9837eb563187559260a84cdc5d24aaca72b5bed/src/HTTP.jl#L293-L301

`mkheaders` is now:
https://github.com/JuliaWeb/HTTP.jl/blob/f9837eb563187559260a84cdc5d24aaca72b5bed/src/Messages.jl#L197

omus:
Ah ok. If the `headers` keyword is now deprecated we should probably add a deprecation

omus:
FYI I managed to make a small test to demonstrate the issue (It appears that not all `PyDict`s cause the issue)
```julia
julia> using HTTP, PyCall

julia> @pyimport requests

julia> r = requests.get("http://httpbin.org/xml")
PyObject <Response [200]>

julia> h = PyDict(r[:request][:headers])
PyCall.PyDict{PyCall.PyAny,PyCall.PyAny,false} with 4 entries:
  "Connection"      => "keep-alive"
  "Accept-Encoding" => "gzip, deflate"
  "Accept"          => "*/*"
  "User-Agent"      => "python-requests/2.18.4"

julia> HTTP.get("http://httpbin.org/xml", headers=h)
ERROR: TypeError: #request: in typeassert, expected Dict, got PyCall.PyDict{PyCall.PyAny,PyCall.PyAny,false}
Stacktrace:
 [1] (::HTTP.#kw##request)(::Array{Any,1}, ::HTTP.#request, ::HTTP.Client, ::String, ::HTTP.URIs.URI) at ./<missing>:0
 [2] #get#23(::Array{Any,1}, ::Function, ::String) at /Users/omus/.julia/v0.6/HTTP/src/client.jl:144
 [3] (::HTTP.#kw##get)(::Array{Any,1}, ::HTTP.#get, ::String) at ./<missing>:0
```
When I use the positional parameter the request works as expected:
```julia
julia> HTTP.get("http://httpbin.org/xml", h)
HTTP.Messages.Response:
"""
HTTP/1.1 200 OK
...
"""
```

samoconnor:
> Ah ok. If the headers keyword is now deprecated we should probably add a deprecation

No, its not deprecated:
https://github.com/JuliaWeb/HTTP.jl/blob/f9837eb563187559260a84cdc5d24aaca72b5bed/src/HTTP.jl#L293-L294

Are you using latest master? This commit ought to fix that problem: https://github.com/JuliaWeb/HTTP.jl/commit/b5844b13a2b19a4d546fdc76f44d2c0e9ee85e97

omus:
The keyword works on the latest master.

omus:
Why do we support headers as a positional and keyword parameter? If this has been discussed elsewhere could you point me to the discussion?

samoconnor:
https://github.com/JuliaWeb/HTTP.jl/issues/141#issuecomment-358511309

PR #135, opened 2017-12-13
WIP: Factorisation of implementation into modules and layers

samoconnor:
_Update: Architecture documentation is now more complete: [HTTP.jl Internal Architecture](https://samoconnor.github.io/HTTP.jl/docs/build/#HTTP.jl-Internal-Architecture-1) and
the new [`HTTP.open`](https://samoconnor.github.io/HTTP.jl/docs/build/#HTTP.open) function provides a more general interface for request and/or response body streaming:_
```julia
HTTP.open("GET", "https://tinyurl.com/bach-cello-suite-1-ogg") do http
    open(`vlc -q --play-and-exit --intf dummy -`, "w") do vlc
        write(vlc, http)
    end
end
```

_Update: Server side is working (but not yet integrated with existing API). See example of using `HTTP.listen` to [create a minimal WebSockets server](https://github.com/samoconnor/HTTP.jl/blob/simplify_parser_branch/src/WebSockets.jl#L119), 
See also example [here](https://github.com/JuliaWeb/WebSockets.jl/issues/84#issuecomment-357078969)_

This branch started as PR #124 (fixes for parser state machine bugs #125 and #126). It evolved from there in an attempt to address some of the other issues I've found (e.g. #108, #115, #120, #122, #130, #131).

The result is an attempt to:
 - split functionality into distinct modules and layers,
 - reduce the size of the interfaces between modules and layers,
 - reduce the degree of type and data interdependancy between modules and layers.

For example the `parse!` function signature before and after:
```julia
function parse!(r::Union{Request, Response}, parser, bytes, len=length(bytes);
        lenient::Bool=true, host::String="", method::Method=GET,
        maxuri::Int64=DEFAULT_MAX_URI, maxheader::Int64=DEFAULT_MAX_HEADER,
        maxbody::Int64=DEFAULT_MAX_BODY, maintask::Task=current_task(),
        canonicalizeheaders::Bool=true)::Tuple{ParsingErrorCode, Bool, Bool, String}
```
```julia
parseheaders(::Parser, bytes) do h::Pair{String,String} ... -> excess
parsebody(::Parser, bytes) -> data, excess
```

Another example is the single consolidated retry loop using `Base.retry`:
```julia
abstract type RetryLayer{Next <: Layer} <: Layer end

isrecoverable(e::Base.UVError) = true
isrecoverable(e::Base.DNSError) = true
isrecoverable(e::Base.EOFError) = true
[...]

PR #142, opened 2018-01-17
Remove MicroLogging

EricForgy:
Servers.jl makes use of `@info`, `@warn` and `@error` macros, which are not supported on v0.6. 

We tried adding some macros to compat.jl, but I was still having trouble on Windows, so I added MicroLogging, which got merged and seemed ok.

However, subsequently, I found MicroLogging was wreaking havoc on stdout to REPL (on Windows) so this PR changes all uses of `@info`, `@warn` and `@error` macros to the v0.6 compatible `info`, `warn` and `error` methods.


EricForgy:
Ack. There is some problem with this PR. There should not be so many files changed. Will see what happened.

samoconnor:
HI @EricForgy, maybe my fault. I un-squashed the commit of #135 to avoid losing granularity with `git blame`.

EricForgy:
Phew 😅 

I thought I did something wrong 😂 

No problem. I'll hope over to the new master now and try again.

PR #143, opened 2018-01-17
Remove MicroLogging

EricForgy:
Servers.jl makes use of `@info`, `@warn` and `@error` macros, which are not supported on v0.6. 

We tried adding some macros to compat.jl, but I was still having trouble on Windows, so I added MicroLogging, which got merged and seemed ok.

However, subsequently, I found MicroLogging was wreaking havoc on stdout to REPL (on Windows) so this PR changes all uses of `@info`, `@warn` and `@error` macros to the v0.6 compatible `info`, `warn` and `error` methods.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -5,7 +5,6 @@ using ..Streams
 using ..Messages
 using ..Parsers
 using ..ConnectionPool
-import ..@info, ..@warn, ..@error, ..@debug, ..@debugshow, ..DEBUG_LEVEL
 using MbedTLS: SSLConfig, SSLContext, setup!, associate!, hostname!, handshake!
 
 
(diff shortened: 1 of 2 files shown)


samoconnor:
A few issues:
 - `@error` logs an error, whereas `error` throws an exception.
 - `warn` is deprecated in 0.7

Best to leave the 0.7 code in Server.jl and add some compatible definitions in compat.jl.
e.g. something like this:

```julia
julia> @warn "hello" x y
W- hello -Warn:Main:REPL[74]:1
|  x = 7
|  y = 8

julia> macro compat_warn(m, args...)
       args = [:(print("|  "); @show $a) for a in args]
       esc(:(println("W- ", $m); $(args...); nothing))
       end
@compat_warn (macro with 1 method)

julia> @compat_warn "hello" x y
W- hello
|  x = 7
|  y = 8
```

PR #144, opened 2018-01-17
Modify testget

EricForgy:
This is (maybe) a slight improvement to `testget`. 

The prior version wrote the output of `curl` to a temp file and then searched the file for `"HTTP/1.1 200 OK"`. This version just extracts the status of `curl` directly from a `Pipe()` so no tmp file is create AND it avoids the use of `bash`, which is problematic on Windows.

The output of the new `testget` is just a string representing the status, e.g. `"200"`. I added a second argument to `testget` for the number of repeated urls you want to call with the single call to `curl`. For example, if `n = 3`, it will call `curl` with the url repeated three times and the output of the function for a successful call would be `"200200200", i.e. the status for each url is appended as strings. 

Not very pretty, but I hope it is ok.

### DIFF
--- a/test/server.jl
+++ b/test/server.jl
@@ -3,16 +3,20 @@ using HTTP.Test
 
 port = rand(8000:8999)
 
-function testget(url)
-    mktempdir() do d
-        cd(d) do
-            cmd = `"curl -v -s $url > tmpout 2>&1"`
-            cmd = `bash -c $cmd`
-            #println(cmd)
-            run(cmd)
-            return String(read(joinpath(d, "tmpout")))
-        end
+function testget(url,n=1)
+    p = Pipe()
+    if Sys.is_windows()
+        null = "NUL"
+    else
+        null = "/dev/null"
     end
+    cmd = `curl $(repeat(["-s","-o",null],outer=n)) -w '%{http_code}' $(repeat([url],outer=n))`
+    println(cmd)
+
+    run(pipeline(cmd,stdout=p))
+    p_txt = String(readavailable(p))
+    close(p)
+    p_txt
 end
 
 @testset "HTTP.Servers.serve" begin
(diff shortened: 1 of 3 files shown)


EricForgy:
Oops. I'm still clueless with GitHub and this PR contains my prior PR (https://github.com/JuliaWeb/HTTP.jl/pull/143 ). I probably should have made a clean separation.

test/server.jl
```diff
@@ -3,16 +3,20 @@ using HTTP.Test
...
+    p = Pipe()
+    if Sys.is_windows()
+        null = "NUL"
+    else
+        null = "/dev/null"
     end
+    cmd = `curl $(repeat(["-s","-o",null],outer=n)) -w '%{http_code}' $(repeat([url],outer=n))`
+    println(cmd)
+
+    run(pipeline(cmd,stdout=p))
```
samoconnor: You can do this with less code:
```julia
open(`curl ...`) do io
    println(readstring(io))
end
```

samoconnor: However, I don't 100% trust pipelining the output of command to Julia. See https://github.com/JuliaLang/julia/issues/14747
That's why I was using bash to dump the output into a file. I trust bash, and I trust Julia to read a file :)

EricForgy: image

I'll be happy to get it to run locally on Windows first and then AppVeyor should be easy 😅

samoconnor: No, windows is not my thing. I think at this point I it might be best to use `@async HTTP.request(...)` on windows and leave a FIXME.

EricForgy: I want to see green on AppVeyor. I'm not giving up so easy 😂

samoconnor: If we used `remote` then at lest the request would come from a seperate process....
```julia
using Distributed # on 0.7
addprocs(1)
remote(u->HTTP.request("GET", u))("http://httpbin.org")
```

samoconnor: I'm just playing with this now... 
```julia
function testget(url, n)
    remote((url, n)-> begin
        r = []
        @sync for i in 1:n
            @async push!(r, HTTP.request("GET", url))
        end
        return r
    end)(url, n)
end
```

EricForgy: Is there a way you can use `curl` on Linux for `testget` that does not require `bash`? I can use `curl` no problem. The problem is `bash`.

samoconnor: In that case, just us `is_windows()` and whatever combination of `open`, `pipeline` works for you on windows.

samoconnor: See https://github.com/JuliaWeb/HTTP.jl/commit/57192e6a6df54d6623ab47d24bd310fea450c710

PR #159, opened 2018-01-21
Remove Response from handler argument and get WS/HTTP working on same port

### LINKED ISSUE #156: HTTP.WebSockets served from the same port
This is more of a feature request that I will try myself, but thought I'd go ahead and create an issue first.

Now, we have `HTTP.Servers.Server.listen` and `HTTP.WebSockets.listen`. It would be nice to have a `listen` that works with both HTTP and WS on the same port.

I think it is as simple as taking the guts from `Server.listen` and `WebSockets.listen` and creating a new `listen` that is dispatched on if given both an HTTP handler and a WS handler (or some other flag).

### LINKED ISSUE #158: Remove Response argument if not needed
A `HandlerFunction` is of the form `(req,res) -> res`.

If the handler does not depend on `res`, we could do something like `(req,Response()) -> res`, but that feels like a waste to construct an object `Response()` that isn't needed. Is the compiler smart enough to optimize that away or does it really construct a `Response()` object?

In Pages, I imported `handle` and made these changes to make it work in order to avoid an empty `Response()`

```
SCHEMES = HTTP.Handlers.SCHEMES
EMPTYVAL = HTTP.Handlers.EMPTYVAL
HTTP.handle(h::HTTP.HandlerFunction,req::HTTP.Request) = h.func(req)
function HTTP.handle(r::HTTP.Router, req::HTTP.Request)
    # get scheme, host, split path into strings and get Vals
    m = Val(Symbol(req.method))
    # get the url/path of the request
    uri = HTTP.URI(req.target)
    s = get(SCHEMES, uri.scheme, EMPTYVAL)
    h = Val(Symbol(uri.host))
    p = uri.path
    segments = split(p, '/'; keep=false)
    # dispatch to the most specific handler, given the path
    vals = (get(r.segments, s, EMPTYVAL) for s in segments)
    handler = r.func(m, s, h, vals...)
    # pass the request & response to the handler and return
    return HTTP.handle(handler, req)
end
```

If there is no penalty for doing `(req,Response()) -> res`, I can just do this, otherwise, would you be open for a PR of the form above?

EricForgy:
I wasn't able to run `Pkg.test` because `addprocs` is broken on my machine, but this works inside Pages.jl.

Closes #158 
Closes #156

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -487,5 +487,14 @@ function HTTP.handle(f::Function, http::Stream)
     return
 end
 
+function HTTP.handle(hf::HandlerFunction, http::Stream)
+    request::HTTP.Request = http.message
+    request.body = read(http)
+    request.response::HTTP.Response = hf.func(request)
+    startwrite(http)
+    write(http, request.response.body)
+    return
+end
+
 
 end # module
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -487,14 +487,6 @@ function HTTP.handle(f::Function, http::Stream)
     return
 end
 
-function HTTP.handle(hf::HandlerFunction, http::Stream)
-    request::HTTP.Request = http.message
-    request.body = read(http)
-    request.response::HTTP.Response = hf.func(request)
-    startwrite(http)
-    write(http, request.response.body)
-    return
-end
-
+HTTP.handle(hf::HandlerFunction, http::Stream) = HTTP.handle(hf.func,http)
 
 end # module
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -455,7 +455,7 @@ function handle_stream(f::Function, http::Stream)
         if applicable(f, http)
             f(http)
         else
-            handle_request(f, http)
+            HTTP.handle(f, http)
         end
     catch e
         if isopen(http) && !iswritable(http)
@@ -478,7 +478,7 @@ end
 Execute Request processing function `f(::HTTP.Request) -> HTTP.Response`.
 """
 
-function handle_request(f::Function, http::Stream)
+function HTTP.handle(f::Function, http::Stream)
     request::HTTP.Request = http.message
     request.body = read(http)
     request.response::HTTP.Response = f(request)
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -163,7 +163,7 @@ function serve(server::Server{T, H}, host, port, verbose) where {T, H}
            ratelimits=Dict{IPAddr, RateLimit}(),
            ratelimit=server.options.ratelimit) do request::HTTP.Request
 
-        handle(server.handler, request, request.response)
+        handle(server.handler, request)
     end
 
     return
--- a/test/server.jl
+++ b/test/server.jl
@@ -49,9 +49,8 @@ sleep(2)
 
 # echo response
 serverlog = HTTP.FIFOBuffer()
-server = HTTP.Servers.Server((req, rep) -> begin
-    rep.body = req.body
-    return rep
+server = HTTP.Servers.Server((req) -> begin
+    return HTTP.Response(200,req.body)
 end, serverlog)
 
 server.options.ratelimit=0
@@ -173,7 +172,7 @@ put!(server.in, HTTP.Servers.KILL)
 
 # keep-alive vs. close: issue #81
 port += 1
-tsk = @async HTTP.Servers.serve(HTTP.Servers.Server((req, res) -> (res.body = "Hello\n"; res), STDOUT), ip"127.0.0.1", port)
+tsk = @async HTTP.Servers.serve(HTTP.Servers.Server((req) -> (HTTP.Response(200,"Hello\n")), STDOUT), ip"127.0.0.1", port)
 sleep(2.0)
 r = HTTP.request("GET", "http://127.0.0.1:$port/", ["Host"=>"127.0.0.1:$port"]; http_version=v"1.0")
 @test r.status == 200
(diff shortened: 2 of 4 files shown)


EricForgy:
Not ready yet. I need to modify the tests too. Working on it 😅

EricForgy:
Removed the args from handler tests, but they appear in some server tests too. I just removed those and waiting for Travis / AppVeyor 🙏

EricForgy:
https://travis-ci.org/JuliaWeb/HTTP.jl/jobs/331411996

https://ci.appveyor.com/project/quinnj/http-jl/build/job/soi8th7cg08t9t41

I think the PR is ok now 👍 🙏 

Looks like there might be a non-deterministic client test that failed, but it passed on the previous two tests and the AppVeyor test passed (on v0.6).

quinnj:
This seems fine to me, but definitely needs some docs. @EricForgy, would you mind writing up some general docs on handlers, and mention that each Request has it's own Response as a field member? That would be awesome.

EricForgy:
Would it be possible to merge this and make a new issue for docs? 😅 

My code relies on these changes and I don't have time to update docs right now. There is a new PR merged and I hope to stay on master if possible.

EricForgy:
There might be an easier way to do this?

What if instead of removing `resp` from all the methods, which is likely to break someone's code who uses HTTP.jl, we just add a few strategic methods like

```
handle(h,req::HTTP.Request) = HTTP(h,req,req.response)
```

?

samoconnor:
Yeah, I'd go for cleaning up first.
If enough people need backwards-compatibity wrappers, that can be another PR. Or, it might be easier to support people through the change. How many existing users are there of this API anyway?

EricForgy:
In order to get WS/HTTP to work, I was using my Pages.jl package, but Pages.jl depends on the original PR here, so I had to just add WS/HTTP changes to this PR. Hope that is alright 😅

EricForgy:
Tests passed on AppVeyor v0.6 now 👍 

One server test failed on Travis v0.6, but I'm guessing it is non-determinstic.

EricForgy:
Tests are passing now on v0.6 for both Travis and AppVeyor 👍 

https://travis-ci.org/JuliaWeb/HTTP.jl/jobs/331892650

https://ci.appveyor.com/project/quinnj/http-jl/build/job/0k09r8afv6duf04b

Sorry I didn't notice the v0.6 tests were not passing before 😅

test/server.jl
```diff
@@ -49,9 +49,8 @@ sleep(2)
 
 # echo response
 serverlog = HTTP.FIFOBuffer()
-server = HTTP.Servers.Server((req, rep) -> begin
-    rep.body = req.body
-    return rep
+server = HTTP.Servers.Server((req) -> begin
+    return HTTP.Response(200,req.body)
```
samoconnor: should be
```julia
req.response.body = req.body
return req.response
```
?

test/server.jl
```diff
@@ -173,7 +172,7 @@ put!(server.in, HTTP.Servers.KILL)
 
 # keep-alive vs. close: issue #81
 port += 1
-tsk = @async HTTP.Servers.serve(HTTP.Servers.Server((req, res) -> (res.body = "Hello\n"; res), STDOUT), ip"127.0.0.1", port)
+tsk = @async HTTP.Servers.serve(HTTP.Servers.Server((req) -> (HTTP.Response(200,"Hello\n")), STDOUT), ip"127.0.0.1", port)
```
samoconnor: Should be `(req.response.body = "Hello\n"; req.response)` ?

src/Servers.jl
```diff
@@ -478,7 +478,7 @@ end
 Execute Request processing function `f(::HTTP.Request) -> HTTP.Response`.
 """
 
-function handle_request(f::Function, http::Stream)
+function HTTP.handle(f::Function, http::Stream)
```
samoconnor: I'd rather not change the name of `handle_request` in Servers.jl.

src/Servers.jl
```diff
@@ -487,5 +487,6 @@ function handle_request(f::Function, http::Stream)
     return
 end
 
+HTTP.handle(hf::HandlerFunction, http::Stream) = HTTP.handle(hf.func,http)
```
samoconnor: I think this should be in handlers.jl.

samoconnor:
Re:
```
            handler = HTTP.gethandler(router,http.message)
            HTTP.handle(handler,http)
```
Why not do this ?
```
            HTTP.handle_request(http) do request::HTTP.Request
                HTTP.handle(router, request)
            end
```

samoconnor:
Hi @EricForgy, I'm going to merge this and tweak a few things per my comments. Please open a new PR if you need any subsequent changes.

samoconnor:
See fd31e706eb402cc3be31deee1b6b5649b46054c0

sairus7:
> Yay! 🎉
> 
> WS/HTTP working on the same port! 😃 🚀
> 
> ```
> function start(p = 8000)
>     global port = p
>     HTTP.listen(ip"127.0.0.1",p) do http
>         if is_upgrade(http.message)
>             HTTP.WebSockets.upgrade(http) do ws
>                 while !eof(ws);
>                     print(String(readavailable(ws)))
>                 end
>             end
>         else
>             handler = HTTP.gethandler(router,http.message)
>             HTTP.handle(handler,http)
>         end
>     end
> end
> ```

An updated version for those who may find this post and want to use it with HTTP.jl v0.9:
```
using Sockets, HTTP, HTTP.WebSockets

function start_server(port = 8081)
    HTTP.listen(ip"127.0.0.1", port) do http
        if WebSockets.is_upgrade(http.message)
            WebSockets.upgrade(http) do ws
                while !eof(ws)
                    print(String(readavailable(ws)))
                end
            end
        else
            handler = HTTP.Handlers.gethandler(router, http.message)
            HTTP.handle(handler, http)
        end
    end
end
```

PR #167, opened 2018-01-23
export handle_request

EricForgy:
Hi @samoconnor 👋 

Thanks for merging my PR 👍 

There was one change you reverted and suggested a neat trick in this comment ( https://github.com/JuliaWeb/HTTP.jl/pull/159#issuecomment-359580010 )

> Re:
```
            handler = HTTP.gethandler(router,http.message)
            HTTP.handle(handler,http)
```
> Why not do this ?
```
            HTTP.handle_request(http) do request::HTTP.Request
                HTTP.handle(router, request)
            end
```

This works for me, but I can't do it precisely like that because `handle_request` is not exported from `Servers`. I would need to make that

```
            HTTP.Servers.handle_request(http) do request::HTTP.Request
                HTTP.handle(router, request)
            end
```

which is not very appealing to me aesthetically so this PR just exports `handle_request` from `Servers` and uses it in `HTTP` so we can call it with `HTTP.handle_request` as suggested.

HOWEVER...

I actually like my original approach better, i.e. rename

`HTTP.handle_request`

to

`HTTP.handle`

This makes sense to me. It is nice to have an easily accessible method that handles streams AND it avoids the need to export `handle_request` because `handle` is already exported.

Then your solution would be

```
            HTTP.handle(http) do request::HTTP.Request
                HTTP.handle(router, request)
            end
```

THAT ☝️  looks very pretty to me. Any reason you don't like it?

If you are ok, I'd prefer to close this PR and submit a new one that changes `HTTP.handle_request` to `HTTP.handle`.

What do you think?

### DIFF
--- a/src/Handlers.jl
+++ b/src/Handlers.jl
@@ -1,6 +1,6 @@
 module Handlers
 
-export handle, gethandler, Handler, HandlerFunction, Router, register!
+export handle, gethandler, handle_request, Handler, HandlerFunction, Router, register!
 
 import ..Nothing, ..Cvoid, ..Val
 
@@ -141,6 +141,6 @@ function handle(r::Router, req)
     return handle(handler, req)
 end
 
-handle(hf::HandlerFunction, http::HTTP.Stream) = HTTP.handle_request(hf.func,http)
+handle_request(f::Function, http::HTTP.Stream) = HTTP.Servers.handle_request(f, http)
 
 end # module


samoconnor:
> ...rename `HTTP.handle_request` to `HTTP.handle`

Well, strictly there was never as `HTTP.handle_request`.

`HTTP.Servers.handle_request` is a non-exported function that is part of the `HTTP.Servers` internal server request execution stack:
```julia
handle_connection(f, ::Connection, ...)
handle_transaction(f, ::Transaction, ...)
handle_stream(f, ::Stream, ...)
handle_request(f, ::Stream, ...)
```

These functions are dispatched by name rather than by type to make the request stack code unambiguous and easy to understand. Note that the last two functions have the same argument type signature.

The proposal to make `HTTP.Servers.handle_request` a method of the `HTTP.Handlers.handle` function seems undesirable because:
 - It does not do what the documentation for `HTTP.Handlers.handle` says:
_"`handle(handler::Handler, request) => Response`."_
(`handle_request` takes a `Stream` argument and returns nothing).
 - `HTTP.Servers` is a lower-layer module than `HTTP.Handlers`, so it doesn't make sense for `HTTP.Servers` to be dependant on a function from `HTTP.Handlers`.

I don't have any objection to the `HTTP.Handlers` module having a function, with whatever name makes sense in that module, that does exactly what `HTTP.Servers.handle_request` does (and is implemented as a 1-liner that calls `HTTP.Servers.handle_request`.

However, I don't like having a function at the `HTTP` level called `HTTP.handle(f, ::Stream)`. That is too ambiguous. Handle a client stream? a server stream? handle the request? handle the response? 

I suggest that if you want to hide the implementation detail (i.e. the need to call `HTTP.Servers.handle_request`) from external packages, you should consider adding a method to `HTTP.Handlers.register!` a websockets handler, then put the code that decides to `upgrade` or `handle` inside the `HTTP.Handlers` module.

EricForgy:
Thanks @samoconnor 🙌 

That sounds like a better plan 👍  I think I'll keep this open for now and close it when I figure out how to implement the `HTTP.Handlers.register!` idea.

EricForgy:
Did this get merged by accident?

https://github.com/JuliaWeb/HTTP.jl/blob/master/src/Handlers.jl#L144

EricForgy:
Yay! 🎉  First time I saw travis pass 😃

EricForgy:
No need for this :)

PR #173, opened 2018-01-24
Fix and improve Dates.format calls

iamed2:
1. Escaped 'Z' as it has a special meaning when TimeZones is loaded (and the code errors)
2. Used the `dateformat` string macro to avoid compiling the date format every time the function is called


iamed2:
Whoops, the two things I did interact poorly

PR #178, opened 2018-01-28
Client and server websockets

EricForgy:
Websocket server and client.

mwe.jl
```
using HTTP

@async HTTP.listen(ip"127.0.0.1",8000) do http
    if HTTP.method(http.message) == "GET" && HTTP.WebSockets.is_upgrade(http.message)
        HTTP.WebSockets.upgrade(http) do client
            while !eof(client);
                msg = String(readavailable(client))
                println(msg)
                write(client, "Hello JavaScript! From Julia")
            end
        end
    else
        HTTP.Servers.handle_request(http) do req::HTTP.Request
            HTTP.Response(200,readstring(joinpath(dirname(@__FILE__),"mwe.html")))
        end
    end
end

@static if is_apple()
    launch(x) = run(`open $x`)
elseif is_linux()
    launch(x) = run(`xdg-open $x`)
elseif is_windows()
    launch(x) = run(`cmd /C start $x`)
end

sleep(2.0)
launch("http://127.0.0.1:8000/examples/mwe")
```

mwe.html
```
<!DOCTYPE html>
<html>
<body>

<p>Check the dev console.</p>

<button type="button" onclick="sayhello()">Say "Hello" to Julia</button>

<script>

var href = window.location.href;
var sock = new WebSocket("ws"+href.substr(4,href.length));
sock.onmessage = function( message ){
    var msg = message.data;
    console.log(msg);
}

sock.onopen = function () {
    sock.send("Hello Julia! From JavaScript")
};

window.onbeforeunload = function() {
    sock.onclose = function () {};
    sock.close()
};

function sayhello() {
    sock.send("Hello Julia! From JavaScript")
}

</script>

</body>
</html>
```

This 

1. Starts an HTTP / WS server, 
2. Launches a browser,
3. Serves mwe.html via HTTP
4. mwe.html connects back to the WebSocket server and sends a message "Hello Julia! From JavaScript"
5. Server receives WS handshake and upgrades the connection
6. WS server receives the message from JS, prints it to the REPL and sends a WS message back to the browser
7. Browser receives the WS message, "Hello JavaScript! From Julia" and prints it to the console.

From the REPL, you can also connect to the WS server as a Julia client

```
julia> HTTP.WebSockets.open("ws://127.0.0.1:8000/examples/mwe") do ws::HTTP.WebSockets.WebSocket
       HTTP.WebSockets.write(ws,"Hello")
       end;
I- Accept:  �    0↑     0↓    127.0.0.1:8000:8000 ≣16
Hello
```

Suggestions / feedback welcome 😊

### DIFF
--- a/src/WebSockets.jl
+++ b/src/WebSockets.jl
@@ -56,17 +56,27 @@ end
 
 # Handshake
 
+function is_upgrade(req::HTTP.Request)
+    is_get = req.method == "GET"
+    # "upgrade" for Chrome and "keep-alive, upgrade" for Firefox.
+    is_upgrade = HTTP.hasheader(req, "Connection", "upgrade")
+    is_websockets = HTTP.hasheader(req, "Upgrade", "websocket")
+    return is_get && is_upgrade && is_websockets
+end
+
+
+function is_upgrade(res::HTTP.Response)
+    is_101 = res.status == 101
+    # "upgrade" for Chrome and "keep-alive, upgrade" for Firefox.
+    is_upgrade = HTTP.hasheader(res, "Connection", "upgrade")
+    is_websockets = HTTP.hasheader(res, "Upgrade", "websocket")
+    return is_101 && is_upgrade && is_websockets
+end
+
 
 function check_upgrade(http)
-
-    if !hasheader(http, "Upgrade", "websocket")
-        throw(WebSocketError(0, "Expected \"Upgrade: websocket\"!\n" *
-                                "$(http.message)"))
-    end
-
-    if !hasheader(http, "Connection", "upgrade")
-        throw(WebSocketError(0, "Expected \"Connection: upgrade\"!\n" *
-                                "$(http.message)"))
+    if !is_upgrade(http.message)
+        throw(WebSocketError(0, "Invalid WebSocket upgrade:\n $(http.message)"))
     end
 end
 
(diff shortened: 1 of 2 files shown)


EricForgy:
Note: I'm working on the tests now and have 4 of 6 WebSocket tests passing and will commit again once all 6 are passing.

EricForgy:
Any thoughts / suggestions? 😊

EricForgy:
@samoconnor / @quinnj 

Any ideas? I don't understand why WS tests pass but WSS tests fail.

I thought maybe one of my `read` / `write` was not calling the `SSLContext` method so inserted `println`'s at every instance and even insert some `println`s in ssl.jl. Each call is hitting the correct `SSLContext` method.

So why does it work with `TCPSocket` and not with `SSLContext`?

samoconnor:
See https://github.com/JuliaWeb/HTTP.jl/commit/45f4e2205b7fb407fae8fa94569674ff1fb93db7

EricForgy:
Thanks @samoconnor 👍 

This https://github.com/JuliaWeb/HTTP.jl/commit/45f4e2205b7fb407fae8fa94569674ff1fb93db7 helps if we want to launch a secure server, but my tests are failing on the client side trying to connect to a WSS server 🤔 

Are there any tricks needed on client side to connect WebSockets to a WSS server? I think not because the travis tests were passing before 🤔 

And regarding server masking, I took care of that in this PR 😊  It checks if a message is being sent by a client or server and masks accordingly 😊

EricForgy:
Although I was happy with my effort here, I'll chalk it up to a good lesson in WebSockets protocol.

PR #182, opened 2018-01-30
Few cleanups of Servers.jl

quinnj:


src/Servers.jl
```diff
@@ -401,6 +396,9 @@ function handle_transaction(f::Function, t::Transaction;
     catch e
         if e isa EOFError && !messagestarted(http.parser)
             return
+        elseif !isopen(http)
```
quinnj: @samoconnor, I added this check in; I was running into _very_ noisy logs when dealing w/ clients that didn't re-use connections (i.e, they make a single request and close the socket).

samoconnor: What is the error? Can you post the logs?
I'm very nervous about ignoring errors without checking the error type.
What if the error is an ArgumentError due to a change in a Base API, or a bug?
What if it is an AssertionError?
This code silently ignores any errors that coincide with the connection being closed.

quinnj: Why does it matter? At this point in the code, we tried to `startread` a closed socket (which we know because we're `catch e` and `!isopen(http)`), there isn't anything else we can do, so why throw a big long stacktrace?

If the remote client closes, I just feel we just print a warning and move on.

samoconnor: It matters because knowledge about the existence of bugs is a very valuable thing; and that `e` that got caught, then ignored because `!isopen(http)` might contain a clue about the existence of a bug; it might be true that most of the time the `e` is about something related to the client dropping the connection, but rare and hard to find bugs are rare, so most-of-the-time is irrelevant, it is the corner cases that we need to think about most.

quinnj: Ok, the error was something like `connection reset by peer`, but I don't know what the actual error type was.

samoconnor: This is especially true in a dynamic language like julia. There could be a if-else-branch inside the header parsing code that is only reached in some rare condition that depends on message fragmentation or timing. This code branch could throw a `UndefVarError` because someone made a typing mistake and the coverage testing is not 100%. But, it could be missed if the `e` is not checked in the catch block.
It is especially important in server code where there are lots of network or resource related exceptions, to be extra careful about bug related exceptions. Otherwise you end up with code that seems ok, but users always feel like it has glitches that they can't pin down, and devs blame it on the network or database lock contention.

samoconnor: OK, here are some steps to take when writing a catch statement:
 - Add `@show typeof(e)` statements or similar to figure out the type of the error.
 - Examine the stack trace to figure out where the error came from.
 - Read the documentation for the code that threw the error.
 - Is there a check that can be done before calling that code that would prevent the error?
 - If the documentation doesn't mention the error that you saw, report that to the owner of the error throwing code and leave a #FIXME next to the catch block with a link to the upstream issue ticket.
 - Look at the fields of the error struct. Is there a `code` field or a `cause` field or something that will help to write a narrow filter. e.g. `typeof(e) == FooError && e.code = :BAR_HAPPENED` ?
 - Write a filter for the error.
 - Review all the other code that is reachable from the try block. Is there any chance that an error that matches the filter could be thrown from somewhere else? If so, would the semantic meaning be equivalent?
 - Before putting the filter in the first place you think of, consider: would it be better to put it closer to the source of the error because that might reduce the chance of catching something unintended?; or  would it be better to go the other way and catch at the top level where an entire process can be restarted. (If an error has happened, then some piece of machinery could be in an inconsistent state, if you're not 100% sure about the side effects of the error, it might be better to re-start a whole module/process/transaction than to carry on).

PR #184, opened 2018-01-30
Update cookies.jl interface to work with new Request/Headers machiner…

quinnj:
…y. Also provide HTTP.cookies to get a Vector{Cookie} from a Request


samoconnor:
aaa6b20a49523aa20ae7b48eea1291920bd20948

PR #194, opened 2018-02-11
Add ReadyState, fix is_websocket_upgrade and add tests

EricForgy:
Sorry! My bad 😅 

I copy / pasted your suggestion (including a typo) without checking 😅

### DIFF
--- a/src/WebSockets.jl
+++ b/src/WebSockets.jl
@@ -10,7 +10,6 @@ using HTTP: header
 import ..@debug, ..DEBUG_LEVEL, ..@require, ..precondition_error
 
 
-
 const WS_FINAL = 0x80
 const WS_CONTINUATION = 0x00
 const WS_TEXT = 0x01
@@ -18,7 +17,6 @@ const WS_BINARY = 0x02
 const WS_CLOSE = 0x08
 const WS_PING = 0x09
 const WS_PONG = 0x0A
-
 const WS_MASK = 0x80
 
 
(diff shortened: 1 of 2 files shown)


EricForgy:
I'll try to add some tests...

EricForgy:
I can't get the test to pass. This might be beyond my skill level 😭 

The first two tests pass on the local WS server, but when I try to write from a WS to IOBuffer, it hangs. Even `close` hangs. I can't figure it out, so I rebased my PR so maybe you can have some suggestions.

EricForgy:
I don't see any place in the code where a socket is ever actually closed...

EricForgy:
I'm playing with the idea of changing:

```
txclosed::Bool -> txstate::ReadyState
```

and

```
rxclosed::Bool -> rxstate::ReadyState
```

where

```
@enum ReadyState CONNECTED=0x1 CLOSING=0x2 CLOSED=0x3
```

This is from the old WebSockets.jl. Since I (we?) are trying to support both client and server, we need some way to avoid bouncing close messages back and forth. If one side is in a `CLOSING` state (meaning it has sent a `close` request) and receives a close message, it is safe to terminate the connection. 

Any thoughts?

EricForgy:
I made some progress in the latest commit. 2 of 3 tests pass, but I'm still having some difficulty getting `
```
write(to::IO, from::HTTP.WebSockets.WebSocket)
```

to work properly with the local WS server.

I'm also getting

```
ArgumentError("stream is closed or unusable")
```

and haven't been able to track it down yet.

EricForgy:
https://github.com/JuliaWeb/HTTP.jl/pull/197 would make this moot 🙏

src/WebSockets.jl
```diff
@@ -37,29 +35,33 @@ struct WebSocketHeader
...
 
 
 # Handshake
 
+
 is_websocket_upgrade(r::HTTP.Message) =
     (r isa HTTP.Request && r.method == "GET" || r.status == 101) &&
     HTTP.hasheader(r, "Connection", "upgrade") &&
-    HTTP.hasheader(r, "Upgrade", "webscoket")
+    HTTP.hasheader(r, "Upgrade", "websocket")
```
samoconnor: Whoops!
Thankyou :)

test/WebSockets.jl
```diff
@@ -5,24 +5,39 @@ using HTTP.IOExtras
...
+        close(ws)
+    end
+end
 
-        buf = IOBuffer()
-        write(buf, io)
-        @test String(take!(buf)) == "Hello There World!"
 
-        close(io)
+p = 8000
```
samoconnor: I've added these tests https://github.com/JuliaWeb/HTTP.jl/commit/2cea2f7e093cf3025d4516523a225ba225611021

EricForgy:
👍

PR #198, opened 2018-02-13
Jq/0.7

quinnj:


test/messages.jl
```diff
@@ -107,7 +107,7 @@ using JSON
 
 #=
     @sync begin
-        io = BufferStream()
+        io = Base.BufferStream()
```
samoconnor: That's annoying ;)
See: https://github.com/JuliaLang/julia/pull/25802#issuecomment-365448930

src/ConnectionRequest.jl
```diff
@@ -39,9 +37,6 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
     end
 end
 
-
-sockettype(url::URI, default) = url.scheme in ("wss", "https") ? SSLContext :
-                                                                 default
-
+sockettype(url::URI, default) = url.scheme in ("wss", "https") ? SSLContext : default
```
samoconnor: This might seem like weird formatting on a wide terminal, but I have taken some care to keep lines of code less than 80 columns where possible.

src/Parsers.jl
```diff
@@ -190,16 +167,12 @@ function parse_status_line!(bytes::AbstractString, response)::SubString{String}
...
 """
 Parse HTTP header-field.
 Return `Pair(field-name => field-value)` and
 a `SubString` containing the remaining header-field lines.
 """
-
-function parse_header_field(bytes::SubString{String})::Tuple{Header,
-                                                             SubString{String}}
-
+function parse_header_field(bytes::SubString{String})::Tuple{Header,SubString{String}}
```
samoconnor: Was wrapped to be < 80 cols.

samoconnor:
Hi @quinnj,
The removal of the blank line between a docstring and the statement that it documents seems like a good change. I don't know why I started doing that, but I seem to have done a lot of it :)
However, most of the other removed whitespace is in places that I intended it to communicate separation of concerns in the source text. e.g.
 - [Two blank lines following the group of `using`, `import` etc statements at the top of a file.](https://github.com/JuliaWeb/HTTP.jl/pull/198/files#diff-aeb305f8cdce4ed0ac6f4e37dd41707eL20)
 - [Two blank lines preceding `end #module`.](https://github.com/JuliaWeb/HTTP.jl/pull/198/files#diff-aeb305f8cdce4ed0ac6f4e37dd41707eL314)
 - [Two blank lines between methods where the methods are not closely-related](https://github.com/JuliaWeb/HTTP.jl/pull/198/files#diff-aeb305f8cdce4ed0ac6f4e37dd41707eL191). I guess that by closely related I mean things like two methods for the same function, or a method and its helper method, which is only called in that one place.
 - [Three blank lines preceding a single-line comment that serves as a "section heading" for a group of related methods:](https://github.com/JuliaWeb/HTTP.jl/pull/198/files#diff-aeb305f8cdce4ed0ac6f4e37dd41707eL146)

The intention is to provide visual cues to the reader, that while not consciously noticed, will help the skimming eye to pick out the first words of a heading, or the start of a paragraph (group of methods).  
 I realise that this is all arguably subjective and down to aesthetic preference. If you feel strongly that to your eye the code was looking messy with the extra whitespace, then lets leave it as is.

PR #200, opened 2018-02-16
WIP: Add more context to server logs

quinnj:


src/Servers.jl
```diff
@@ -382,14 +386,16 @@ Otherwise, execute stream processing function `f`.
...
+                            verbose::Bool=false,
+                            conn::Int=1, kw...)
 
     request = HTTP.Request()
     http = Streams.Stream(request, t)
 
     try
         startread(http)
     catch e
+        @show dump(e)
```
samoconnor: Doing console IO before `rethrow` will currently corrupt the backtrace because of https://github.com/JuliaLang/julia/issues/19979#issuecomment-344453118
Maybe better to only `@show` if `DEBUG_LEVEL` is set? Or leave it there but commented out like the two below it?

src/Servers.jl
```diff
@@ -291,6 +291,7 @@ function listen(f::Function,
     tcpserver = Base.listen(getaddrinfo(host), port)
 
     tcpref[] = tcpserver
+    conn = 1
```
samoconnor: Maybe call this `count`?

samoconnor: But, it seems that instead of passing a connection count down through the stack just for log identification, maybe it's better to use the new `with_logger` feature to keep track of logs per connection:

https://github.com/JuliaLang/julia/pull/26005/files#diff-3e96068c2d524c60e771965c63dca20aR125

https://github.com/JuliaLang/julia/issues/25109#issuecomment-363608019

PR #206, opened 2018-02-21
Fix is_upgrade when checking a POST request

EricForgy:
I was getting ready to tag a new version of Pages and noticed that POST request was failing.

This line

```julia
 (r isa HTTP.Request && r.method == "GET" || r.status == 101)
```
checks to see if a `Requet` is a `GET` and if not, it checks whether the `status` is 101, but `status` is not a property of a `Request`. It is only a property of a `Response`, so if this line receives a POST request, it fails because their is no property `status` in the POST `Request`.

### DIFF
--- a/src/WebSockets.jl
+++ b/src/WebSockets.jl
@@ -56,10 +56,11 @@ end
 
 # Handshake
 
-is_upgrade(r::HTTP.Message) =
-    (r isa HTTP.Request && r.method == "GET" || r.status == 101) &&
+function is_upgrade(r::HTTP.Message)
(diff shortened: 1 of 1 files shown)


EricForgy:
Oh man! The latest tag did not pickup this PR so now I cannot tag a new version of package. 

@quinnj / @samoconnor Please consider this before you tag the next version please 🙏

EricForgy:
POST Requests are broken in HTTP.jl if you mix Http and Websocket handlers because is_upgrade will fail.

EricForgy:
Thank you 🙏 😊

PR #211, opened 2018-03-02
Try to reduce risk of EOF error with POST requests

samoconnor:
- Add a txcount field to the Request struct
- Increment txcount after the body is written
- Modify RetryRequest.isrecoverable to allow retry of non-idempotent
  requests if txcount is zero.
- Modify StreamRequest to Wait for other pipelined reads to complete
  before sending a non-idempotent request body. This should decrease
  the chance of sending a POST body that ends up failing and can't be
  retried.
- Remove yeild() after @async writebody. This should increase the chance
  that the startread() realises that the connection is dead before the
  body is written.
  Depends on: https://github.com/JuliaWeb/MbedTLS.jl/pull/124

### DIFF
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -177,6 +177,8 @@ Represents a HTTP Request Message.
 
 - `response`, the `Response` to this `Request`
 
+- `txcount`, number of times this `Request` has been sent (see RetryRequest.jl).
+
 - `parent`, the `Response` (if any) that led to this request
   (e.g. in the case of a redirect).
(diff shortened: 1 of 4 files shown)


samoconnor:
> 1560 passed, 0 failed, 2 errored, 0 broken.

These are the two https POST tests that fail because of JuliaWeb/MbedTLS.jl#124

quinnj:
0.6 tests all pass w/ the new MbedTLS tag; 0.7 failures are unrelated (moving Sockets out of Base)

PR #212, opened 2018-03-02
Add an experimental keepalive keyword arg for setting keepalive on tc…

quinnj:
…p sockets in connections

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -451,14 +451,24 @@ function getconnection(::Type{Transaction{T}},
     end
 end
 
+function keepalive!(tcp)
+    @debug 2 "setting keepalive on tcp socket"
+    err = ccall(:uv_tcp_keepalive, Cint, (Ptr{Cvoid}, Cint, Cuint),
+                                          tcp.handle, 1, 1)
+    err != 0 && error("error setting keepalive on socket")
+    return
+end
+
 function getconnection(::Type{TCPSocket},
                        host::AbstractString,
                        port::AbstractString;
+                       keepalive::Bool=false,
                        kw...)::TCPSocket
-
     p::UInt = isempty(port) ? UInt(80) : parse(UInt, port)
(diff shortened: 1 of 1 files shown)


samoconnor:
Nice one. This is worth trying for sure!

samoconnor:
I was just about to say, go ahead and merge this because it should have no effect when not enabled.
I think I'd be happier to merge this with the default being disabled, and then test Femtocleaner with #211 with and without this setting.

quinnj:
That sounds good; I wanted to enable it and have the test suite run just to see if it actually affected anything, but then we can merge w/ it off by default until we feel more comfortable.

src/ConnectionPool.jl
```diff
@@ -451,14 +451,24 @@ function getconnection(::Type{Transaction{T}},
     end
 end
 
+function keepalive!(tcp)
+    @debug 2 "setting keepalive on tcp socket"
+    err = ccall(:uv_tcp_keepalive, Cint, (Ptr{Cvoid}, Cint, Cuint),
+                                          tcp.handle, 1, 1)
```
quinnj: @samoconnor @Keno, the 2nd argument here is an "initial delay", which I assume is how long it waits before sending the first keepalive packet (it's unclear how frequently it sends it after that, but whatever). I just set this to 1 for now, but if you guys have other opinions or concerns, I'm happy to try whatever here.

samoconnor: Every time I've looked into the libuv docs I find it very hard to figure out what the exact behaviour is supposed to be. They often just say "see the linux man page for more" without explaining how their behaviour differs. (Sometimes I wonder if we'd be better off using BSD sockets directly and using some kind of BSD/Winsock layer to make it work on windows. libuv is a weird beast).

I think in this case it would make sense to observe what goes on with wireshark.
Probably OSX, Linux and Windows will behave differently too.

Perhaps you can find some place in libcurl, or in python's requests library where they set keepalive options and see if there are any clues there...

PR #213, opened 2018-03-03
0.7 compat

quinnj:


src/ConnectionPool.jl
```diff
@@ -11,7 +11,7 @@ struct to manage pipelining and connection reuse and a
 pipelined request. Methods are provided for `eof`, `readavailable`,
 `unsafe_write` and `close`.
 This allows the `Transaction` object to act as a proxy for the
-`TCPSocket` or `SSLContext` that it wraps.
+`Sockets.TCPSocket` or `SSLContext` that it wraps.
```
samoconnor: I would use the shorthand `TCPSocket` in docstrings, but I'm not too fussed.

src/ConnectionPool.jl
```diff
@@ -459,14 +459,14 @@ function keepalive!(tcp)
     return
 end
 
-function getconnection(::Type{TCPSocket},
+function getconnection(::Type{Sockets.TCPSocket},
```
samoconnor: Can we just import Sockets into the module rather than spelling out out everywhere?

src/DebugRequest.jl
```diff
@@ -1,16 +1,13 @@
 module DebugRequest
 
-import ..Layer, ..request
-using ..IOExtras, ..compat_stdout
+import ..Layer, ..request, ..compat_stdout
+using ..IOExtras
 import ..ConnectionPool: ByteView, byteview
 
-
```
samoconnor: why?

src/RetryRequest.jl
```diff
@@ -7,6 +7,13 @@ using ..MessageRequest
 using ..Messages
 import ..@debug, ..DEBUG_LEVEL
 
+@static if isdefined(Base, :DNSError)
```
samoconnor: I would rather this be based on `VERSION < ` so that it stands out more as a compat hack that should be removed later (as opposed to being some kind of cleverness used to dynamically handle optional dependancies or some such). I've found recently when I come to review compat stuff that is done based on isdefined or isapplicable that it takes some time to figure out which is the old way and which is the new way. An explicit `VERSION < ` is ambiguous.

src/WebSockets.jl
```diff
@@ -8,8 +8,7 @@ using ..Streams
 import ..ConnectionPool
 using HTTP: header
 import ..@debug, ..DEBUG_LEVEL, ..@require, ..precondition_error
-
-
```
samoconnor: why?

src/compat.jl
```diff
@@ -22,6 +22,7 @@ __init__() = supported() || compat_warn()
     const compat_contains = Base.contains
```
samoconnor: need to update the `v07 = v"0.7.0-DEV.4366"` just above here?

src/compat.jl
```diff
@@ -57,6 +59,18 @@ else
         args = [:(print("|  "); @show $a) for a in args]
         esc(:(println("E- ", $m); $(args...); nothing))
     end
+    Base.fetch(t::Task) = wait(t)
+
+end
+
+@static if isdefined(Base, :_sockname)
```
samoconnor: Can you merge this into the `if VERSION >= v07` above?

src/ConnectionPool.jl
quinnj: Either way; I kind of like spelling it out just for future reference of where this thing is actually defined; "oh, in the Sockets stdlib package", as opposed to assuming it's in Base, or having to hunt through imports.

src/RetryRequest.jl
quinnj: Meh, this is such a minor compat case and I don't want to go digging through commit history to find what version it was moved out on. This "autodetects" the compat based on actual, current definitions and we'll just take it out in a few weeks anyway when 0.7 is released.

src/compat.jl
quinnj: done

src/compat.jl
quinnj: done

src/ConnectionPool.jl
samoconnor: I'd agree if it was `Sockets.TCP`. But as it is, I don't think this is an improvement, and a small diff is a good diff.

src/RetryRequest.jl
samoconnor: > don't want to go digging through commit history to find what version it was moved out on

I'm not suggesting that.

We should defined one version of 0.7 that we support and use that for all the compat decisions: 
https://github.com/JuliaWeb/HTTP.jl/blob/a1c6654fcb45f342bf7152b6aa4c7bc5aa25a832/src/compat.jl#L3

e.g. If you've just made things work on 0.7, update `v07 =` to whatever `VERSION` says at the time.

quinnj: addressed

src/ConnectionPool.jl
quinnj: I like `Sockets.TCP`; implemented

samoconnor: > agree if it was Sockets.TCP

... but it isn't.

I'm not sure that evaling a type abbreviation into a stdlib module will be seen as a good thing.
Now if people go looking for `TCP` in the `Sockets` package they won't find it, and they'll have to look in HTTP.jl/src/compat.jl to find it.
Are you intending a PR against stdlib/Sockets to change it there too?

To me, the fact that `TCPSocket` is exported by the stdlib/Sockets package indicates that `TCPSocket` is intended to be a symbol in the global namespace for modules that are using Sockets. I really think we should just import the symbol and leave the code that uses it as-is.

Things like `HTTP.Message` make sense because they are not exported, and you can imagine having an `SMTP.Message` and an `SMS.Message`. But I just don't see that `Sockets.TCP` fits that pattern. I would expect a type called `TCP` (even in the contact of a `Sockets` module) to be a type that represents the TransferControlProtocol, not an instance of a Socket that implements that protocol. e.g. `TCP <: AbstractProtocol`, `TCPSocket <: AbstractSocket`. `Internet.TCP.Socket <: AbstractSocket` would make sense, but `Sockets.TCP` really doesn't.

Stuff like `Sockets.getsockname` also just seems more verbose for not gain.

One day, if someone refactors the `Sockets` module to use terser names that are intended to be prefixed by the module name, then lets apply that pattern, but for now, in a PR titled "0.7 compat", It really seems best to put `using Sockets` in `compat.jl` and `using ..Sockets` in modules that need it.

quinnj: I don't have the time or care to keep worrying about this; to me, it's clearer to preface everything w/ `Sockets` because it would be easy for a naive user to assume that `HTTP` itself defines a `getsockname` or `IPAddr` type address. At the end of the day, it's what, another 8 characters or so? Just not worth worrying about IMO. Feel free to change as you'd like, I really don't mind with whatever you'd like to go w/, but I'm going to merge this for now and you can change to whatever you'd like since you seem to have stronger feelings on the matter.

s2maki:
I believe there is a bug in this change w.r.t. Julia 0.6.  On line 121 of IOExtras.jl, `Base._sockname` was changed to `Sockets._sockname`.  But there is no `Sockets` module before 0.7.

s2maki:
In addition, and this is probably unrelated to this PR, I am actually seeing an `UndefVar` error related to this particular call of `_sockname`.  It is in a try/catch block, so that should not be possible.  But I have otherwise seen issues in Julia where something in a try/catch seems to be thrown anyhow.

samoconnor:
Nice pickup. On my system the catch was resulting in `0` socket e.g. `httpbin.org:80:0`.
After 7f8e4ba I now see `httpbin.org:80:59882`

samoconnor:
On what platform/version are you seeing the undef var error escaping try/catch? Do you think it is worth filing an issue at https://github.com/JuliaLang/julia ?

s2maki:
I see it in both Linux and Mac. But it’s not consistent. The more sockets I have open at the same time, the more likely it seems to be. But I can’t tell if it’s linear or not.

I recall seeing an ignored catch when doing other AWSCore.jl based work several months ago. I thought it was related to the break/finally bug but we’re running a version that was patched with the fix for that.

It would be worth reporting, but I don’t know how to create a reliable reproduction. I suspect it is a cross-task (reentrancy?) bug.

Issue #214, opened 2018-03-03
Analysis of EOF Errors with GitHub API (Seems Resolved)


samoconnor:
Using `verbose=3` and `DEBUG_LEVEL=3`.
Using the `retry_post_requests` branch (no keepalive) _-- note: this branch (and the MbedTLS bug fix the goes with it) is now merged into master_

Note that my POST request to GitHub fails with a 404, but this does not effect the test.
The test is trying to observer the behaviour what a POST request is sent on a connection that has been sitting idle for a while. The 404 result is not important.

First request to create a gist, get a response without issue.
```julia
julia> using GitHub ; token = "XXX" ; myauth = GitHub.authenticate(token)

DEBUG: 581e SSL connect: api.github.com:443...
DEBUG: 581e TCP connect: api.github.com:443...
DEBUG: 581e 🔗  New:            ⏸    0↑     0↓    api.github.com:443:63915 ≣16 RawFD(23)
DEBUG: 581e 👁  Start write:T0  ⏸    0↑     0↓    api.github.com:443:63915 ≣16 RawFD(23)
DEBUG: 581e 👁  Start read: T0  ⏸    0↑🔒    0↓    api.github.com:443:63915 ≣16 RawFD(23)
DEBUG: e0ca 🗣  Write done: T0  ⏸    1↑     0↓    api.github.com:443:63915 ≣16 RawFD(23)
DEBUG: 581e ✉️  Read done:  T0  ⏸    1↑     1↓    api.github.com:443:63915 ≣16 RawFD(23)
GitHub.OAuth2(1e2b86**********************************)

julia> GitHub.create_gist(auth=myauth; params= Dict(["description" => "foo", "public" => false, "files" => Dict([ "f1" => "foo" ])]))
DEBUG: 581e ♻️  Idle:           ⏸    1↑     1↓    api.github.com:443:63915 ≣16 RawFD(23)
DEBUG: 581e 👁  Start write:T1  ⏸    1↑     1↓    api.github.com:443:63915 ≣16 RawFD(23)
➡️   "POST /gists HTTP/1.1\r\n"
➡️   "Authorization: token 1e2b86bXX32e7415f590c196b7\r\n"
➡️   "User-Agent: GitHub-jl\r\n"
➡️   "Host: api.github.com\r\n"
➡️   "Content-Length: 57\r\n"
➡️   "\r\n"
[...]

PR #215, opened 2018-03-03
Add ConnectionPool connect_timeout and idle_timeout

samoconnor:


src/ConnectionPool.jl
```diff
@@ -385,7 +388,11 @@ end
 Remove closed connections from `pool`.
 """
 function purge()
-    isdeletable(c) = !isopen(c.io) && (@debug 1 "🗑  Deleted:        $c"; true)
+    isdeletable(c) = (!isopen(c.io) || (c.idle_timeout > 0 &&
+                                       !c.readbusy &&
+                                       !c.writebusy &&
+                                       time()-c.timestamp > c.idle_timeout)) &&
```
samoconnor: need to call close(c.io) here!

quinnj:
Do you think we could somehow keep our own list of known host timeouts and use those defaults per-host? I'm just thinking about the excellent sleuthing you did in #214, like, we could have a `github.com` default idle/connect timeout, maybe if we detect a specific server type in response headers. Hmmm....maybe too much work to be worth it, but you'd know a little better w/ all the research you've done on this lately. I'm just thinking it might be nice for users to not have to worry about it in some common cases.

src/ConnectionPool.jl
```diff
@@ -459,16 +476,45 @@ function keepalive!(tcp)
...
+    if connect_timeout == 0
+        tcp = Sockets.connect(Sockets.getaddrinfo(host), p)
+        keepalive && keepalive!(tcp)
+        return tcp
+    end
+
+    # FIXME this currently causes the Julia runtime to hang on exit...
+    result = Ref{TCPSocket}()
+    error = Ref{Any}()
+    @schedule try
```
samoconnor: @vtjnash  I'd like your opinion on this.
Should we include a best-attempt connection timeout feature. Or should we just say for now that underlying infrastructure currently makes this hard to implement, and that users should do their own @async timeout thing if they really need a connect timeout?
https://github.com/JuliaCloud/AWSCore.jl/issues/24#issuecomment-369815567

samoconnor:
> Re: list of known host timeouts ...

Doesn't feel right to me.
It might make sense for API packages like GitHub.jl, AWSSQS.jl to be tuned for the behaviour of their particular servers.
I think a sensible default is to have no timeout (assuming that the current scheme of retrying automatically when the connection is found to be dead works in general). Even with a well tuned timeout, the first connection after the timeout still has to pay the reconnection penalty. I think anyone who's application depends on guaranteed worst-case latency limits will need to have some higher level scheme of application level heartbeats, pre-warming of spare connections etc.

What I've read about TCP level keepalive is suggests that in practice many CDNs, ISP NATs, firewalls, proxies etc either don't forward keepalives, or don't count them as activity for timeout purposes.

The general idea of a metadata directory for HTTP endpoints is an interesting one. I don't think building the database into the Julia HTTP library is right, but I can imagine a web service where you do `GET http://httpmeta.net/info?url=foobar.com` and get back a JSON containing a bunch of info about standards compliance of that server, known issues, tuning parameters, etc etc. Then again, maybe it wouldn't be much use in practice because these days when you go to foobar.com, there are many layers of geo-aware CDNs, local ISP caches/proxies, etc between you and the "real server", so the tuning that works for the thing that "foobar.com" connects you to, might not work for me.

src/ConnectionPool.jl
vtjnash: It seems reasonable to add, I just recommended using `close` to do it "properly" (I don't know why the libuv developers claimed in the linked issue that you couldn't cause ECANCELED, since I clearly can):
```julia
julia> s = Sockets.TCPSocket()
Main.Sockets.TCPSocket(RawFD(0xffffffff) init, 0 bytes waiting)

julia> Sockets.connect!(s, Sockets.ip"192.168.101.1", 80)

julia> @async @show Sockets.wait_connected(s)

julia> close(s)

julia>
ERROR (unhandled task failure): connect: operation canceled (ECANCELED)
```

Around this, you just need to add a bit of logic to coordinate closing the socket if the timeout was reached and the socket is still `s.state == StatusConnecting`. And possible also to loop until the timeout is reached (as mentioned in the linked issue, the kernel has its own idea of what the default timeout should be), if you think 2 minutes is too short.

This also uncovers a minor bug in Base, since this assert is wrong:
```diff
diff --git a/stdlib/Sockets/src/Sockets.jl b/stdlib/Sockets/src/Sockets.jl
index ec7876a917..e947da8fd4 100644
--- a/stdlib/Sockets/src/Sockets.jl
+++ b/stdlib/Sockets/src/Sockets.jl
@@ -371,12 +371,13 @@ end
 function uv_connectcb(conn::Ptr{Cvoid}, status::Cint)
     hand = ccall(:jl_uv_connect_handle, Ptr{Cvoid}, (Ptr{Cvoid},), conn)
     sock = @handle_as hand LibuvStream
-    @assert sock.status == StatusConnecting
     if status >= 0
-        sock.status = StatusOpen
+        if !(sock.status == StatusClosed || sock.status == StatusClosing)
+            sock.status = StatusOpen
+        end
         notify(sock.connectnotify)
     else
-        sock.status = StatusInit
+        ccall(:jl_forceclose_uv, Cvoid, (Ptr{Cvoid},), hand)
         err = UVError("connect", status)
         notify_error(sock.connectnotify, err)
     end
```

samoconnor: See new commit.
When I tried to use just `close(tcp)` the close call didn't return. If I did `@schedule close(tcp)` the REPL hung on exit. So, for now I'm calling `ccall(:jl_forceclose_uv`, which seems to work.

Is is worth making a PR for stdlib/Socket for this? (It seems like it might be easier to make it robust without being confined to the official public interface) or should it stay in HTTP.jl for now?

src/ConnectionPool.jl
```diff
@@ -459,16 +476,48 @@ function keepalive!(tcp)
...
+        keepalive && keepalive!(tcp)
+        return tcp
+    end
+
+    tcp = Sockets.TCPSocket()
+    Base.connect!(tcp, Sockets.getaddrinfo(host), p)
+    yield()
+
+    # sum(delays) ~= connect_timeout
+    delays = ExponentialBackOff(n=connect_timeout + 10,
```
vtjnash: Why are you busy-polling this? We have a callback for it (wait_connected)

samoconnor: I couldn't see how to simultaneously wait for connection success and a timeout.
I can call wait_connected in a background task, and notify a condition when it is done, but there is no timeout parameter wait(::Condition). Or I could wait for the task to be done, but `fetch(::Task)` has no timeout parameter.

Are you thinking along the lines of `wait_connected` in the main task and a background task that calls close after the timeout to wake up the `wait_connected` with an error? I used that pattern over here: https://github.com/JuliaWeb/HTTP.jl/blob/master/src/TimeoutRequest.jl#L20
Don't know why it didn't occur to me in this case :)

vtjnash: I think you answered your own question...yes.

src/ConnectionPool.jl
```diff
@@ -459,14 +476,49 @@ function keepalive!(tcp)
...
+    tcp = Sockets.TCPSocket()
+    Base.connect!(tcp, Sockets.getaddrinfo(host), p)
+
+    timeout = Ref{Bool}(false)
+    @schedule begin
+        sleep(connect_timeout)
+        if tcp.status == Base.StatusConnecting
+            timeout[] = true
+            ccall(:jl_forceclose_uv, Void, (Ptr{Void},), tcp.handle)
+            #close(tcp)
```
samoconnor: @vtjnash close didn't work here (0.6.2).
But I assume that's some small detail of handling status flags in Base.close because closing while opening is not a common operation.

src/ConnectionPool.jl
```diff
@@ -459,14 +476,49 @@ function keepalive!(tcp)
...
+        end
+    end
+    try
+        Base.wait_connected(tcp)
+    catch e
+        if timeout[]
+            throw(ConnectTimeout(host, port))
```
vtjnash: Why do you want to make sure it throws this exception?

src/ConnectionPool.jl
vtjnash: Yeah, maybe fixed on master? This is right, but should also set the state to StatusClosing

src/ConnectionPool.jl
samoconnor: Because I don't think `UVError` should ever be thrown from a user-level API unless it means there is a bug. See https://github.com/JuliaLang/julia/pull/15879 etc.

I think application level code will want to handle "I set a short timeout and it wasn't met" differently from "something bad happened deep in the network stack.

Elsewhere in HTTP.jl we wrap things like `UVError` and `ArgumentError "stream is closed or unusable"` in an `IOError` type:
https://github.com/JuliaWeb/HTTP.jl/blob/master/src/ConnectionRequest.jl#L36

In higher level code it is important to handle some specific errors by retrying but make sure everything else finds its way up to create a crash and a stack trace and an issue report and a fix. e.g. https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/http.jl#L47-L51

samoconnor: `ArgumentError "stream is closed or unusable"` seems pretty wrong because it looks like a precondition error, but there is now way to check the precondition before calling the API because the stream be `isopen()` before the call but be found to be "closed or unusable" while the call is running.

samoconnor: Hi @vtjnash, I just noticed that `poll_fd` has a `timeout_s` option. Would that work here?

vtjnash: I'm not sure how that's relevant here, since you don't have a `fd`. But yes, that does show another pattern that can be used to implement timeouts (with some more difficulty ... in fact, you've just inspired me to rewrite that file better ... you may see a new PR to base tomorrow sometime 🙂)

samoconnor: I was assuming that I could get the `fd` out of the socket with `uv_fileno` or something like that.
Is there a gotcha along the lines of windows socket handles won't work with `poll_fd`?

> new PR

👍🙃

vtjnash: (a) the socket isn't open, so you can't poll it (b) it's invalid to pass libuv-derived fd handles to libuv

samoconnor: thx for taking the time to explain :)

samoconnor:
@quinnj would you mind doing a quick review of this and the merging?
It might be handy to have it available whenever someone has time to dbug GitHub.jl more.

samoconnor:
No problem. And thanks!

Issue #218, opened 2018-03-07
Is proxy support useful?


samoconnor:
@ignaciomartin asked about proxy support a while back: https://github.com/JuliaWeb/HTTP.jl/issues/4#issuecomment-311885194

I remember the importance of getting my proxy settings right in Netscape Navigator and tuning my SLIP MTUs in Trumpet Winsock in 1994. But, does anyone need to explicitly configure a proxy in 2018?
In my experience ISP and corporate proxies are usually transparent these days.

Please excuse my ignorance if there is still a use case for configuring a proxy.
If you need this, please describe what you need here and/or submit a PR.

![](http://www.metronet.lib.mi.us/help/OtherHelpTopics/DNS/trumpet1.gif)
![](http://startupbeat.hkej.com/wp-content/uploads/2016/08/internet-24aug.jpg)

amellnik:
Npm, Atom, Bower, Julia (the package manager), Anaconda/pip, wget, curl, docker, rust and many others would all not work on my system if I didn't have `HTTP_PROXY` defined (or provided the settings in a dotfile somewhere).  Judging from the prevalence of StackOverflow and similar questions about proxy configurations I think this is still fairly common.

samoconnor:
See : https://github.com/JuliaWeb/HTTP.jl/pull/219/files

Example below of `proxy="http://127.0.0.1:8888"` option.

This supports HTTP only (not HTTPS) at present.

Is tunneling the a HTTPS connection through a proxy [via the `CONNECT` method](https://github.com/JuliaWeb/Requests.jl/blob/master/src/streaming.jl#L190-L202) a useful feature too?

```julia
julia> HTTP.request("GET", "http://httpbin.org/ip"; verbose=3, proxy="http://127.0.0.1:8888")
DEBUG: a7be 🗑  Deleted:        💀    1↑     1↓    127.0.0.1:8888:0 ≣16 RawFD(-1)
DEBUG: a7be TCP connect: 127.0.0.1:8888...
DEBUG: a7be 🚷  No Retry: GET http://httpbin.org/ip HTTP/1.1, connect: connection refused (ECONNREFUSED) not recoverable,
ERROR: connect: connection refused (ECONNREFUSED)

julia> run(`brew install tinyproxy`)

julia> run(`tinyproxy`)

julia> HTTP.request("GET", "http://httpbin.org/ip"; verbose=3, proxy="http://127.0.0.1:8888")
[...]

PR #219, opened 2018-03-08
proxy= option

samoconnor:
This supports plain old proxying of HTTP requests, and CONNECT tunnelling of HTTPS requests.
See #218 
Needs tests.

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -488,6 +488,15 @@ end
 function getconnection(::Type{SSLContext},
                        host::AbstractString,
                        port::AbstractString;
+                       kw...)::SSLContext
+
+    port = isempty(port) ? "443" : port
+    @debug 2 "SSL connect: $host:$port..."
+    tcp = getconnection(TCPSocket, host, port; kw...)
+    return sslconnection(tcp, host; kw...)
+end
+
+function sslconnection(tcp::TCPSocket, host::AbstractString;
                        require_ssl_verification::Bool=true,
                        sslconfig::SSLConfig=nosslconfig,
                        kw...)::SSLContext
@@ -496,14 +505,19 @@ function getconnection(::Type{SSLContext},
         sslconfig = global_sslconfig(require_ssl_verification)
     end
 
-    port = isempty(port) ? "443" : port
-    @debug 2 "SSL connect: $host:$port..."
     io = SSLContext()
     setup!(io, sslconfig)
-    associate!(io, getconnection(TCPSocket, host, port; kw...))
+    associate!(io, tcp)
     hostname!(io, host)
     handshake!(io)
     return io
+end
+
+function sslupgrade(t::Transaction{TCPSocket},
+                    host::AbstractString; kw...)::Transaction{SSLContext}
+    tls = sslconnection(t.c.io, host; kw...)
+    c = Connection(tls)
+    return client_transaction(c)
 end
 
 function Base.show(io::IO, c::Connection)
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -23,18 +23,47 @@ abstract type ConnectionPoolLayer{Next <: Layer} <: Layer end
 export ConnectionPoolLayer
 
 function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
+                 proxy=nothing,
                  socket_type::Type=TCPSocket, kw...) where Next
+
+    if proxy != nothing
+        target_url = url
+        url = URI(proxy)
+        if target_url.scheme == "http"
+            req.target = string(target_url)
+        end
+    end
 
     IOType = ConnectionPool.Transaction{sockettype(url, socket_type)}
     io = getconnection(IOType, url.host, url.port; kw...)
 
     try
+
+        if proxy != nothing && target_url.scheme == "https"
+            target = "$(URIs.hoststring(target_url.host)):$(target_url.port)"
+            @debug 1 "📡  CONNECT HTTPS tunnel to $target"
+            r = Request("CONNECT", target)
+            writeheaders(io, r)
+            startread(io)
+            readheaders(io, r.response)
+            if r.response.status != 200
+                return r
+            end
+            io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
+        end
+
         return request(Next, io, req, body; kw...)
+
     catch e
         @debug 1 "❗️  ConnectionLayer $e. Closing: $io"
         close(io)
         rethrow(isioerror(e) ? IOError(e, "during request($url)") : e)
+    finally
+        if proxy != nothing && target_url.scheme == "https"
+            close(io)
+        end
     end
+
 end
 
 sockettype(url::URI, default) = url.scheme in ("wss", "https") ? SSLContext : default
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -38,18 +38,8 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
     io = getconnection(IOType, url.host, url.port; kw...)
 
     try
-
         if proxy != nothing && target_url.scheme == "https"
-            target = "$(URIs.hoststring(target_url.host)):$(target_url.port)"
-            @debug 1 "📡  CONNECT HTTPS tunnel to $target"
-            r = Request("CONNECT", target)
-            writeheaders(io, r)
-            startread(io)
-            readheaders(io, r.response)
-            if r.response.status != 200
-                return r
-            end
-            io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
+            return tunnel_request(Next, io, target_url, req, body; kw...)
         end
 
         return request(Next, io, req, body; kw...)
@@ -66,6 +56,27 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
 
 end
 
-sockettype(url::URI, default) = url.scheme in ("wss", "https") ? SSLContext : default
+sockettype(url::URI, default) = url.scheme in ("wss", "https") ? SSLContext :
+                                                                 default
+
+function tunnel_request(Next, io, target_url, req, body; kw...)
+    r = connect_tunnel(io, target_url)
+    if r.status != 200
+        return r
+    end
+    io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
+    return request(Next, io, req, body; kw...)
+end
+
+function connect_tunnel(io, target_url)
+    target = "$(URIs.hoststring(target_url.host)):$(target_url.port)"
+    @debug 1 "📡  CONNECT HTTPS tunnel to $target"
+    request = Request("CONNECT", target)
+    writeheaders(io, request)
+    startread(io)
+    readheaders(io, request.response)
+    return request.response
+end
+
 
 end # module ConnectionRequest
--- a/src/MessageRequest.jl
+++ b/src/MessageRequest.jl
@@ -23,7 +23,6 @@ function request(::Type{MessageLayer{Next}},
                  method::String, url::URI, headers::Headers, body;
                  http_version=v"1.1",
                  target=resource(url),
-                 proxy=nothing,
                  parent=nothing, iofunction=nothing, kw...) where Next
 
     defaultheader(headers, "Host" => url.host)
@@ -37,11 +36,6 @@ function request(::Type{MessageLayer{Next}},
         elseif method == "GET" && iofunction isa Function
             setheader(headers, "Content-Length" => "0")
         end
-    end
-
-    if proxy != nothing
-        target = string(url)
-        url = URI(proxy)
     end
 
     req = Request(method, target, headers, bodybytes(body);
--- a/src/MessageRequest.jl
+++ b/src/MessageRequest.jl
@@ -23,6 +23,7 @@ function request(::Type{MessageLayer{Next}},
                  method::String, url::URI, headers::Headers, body;
                  http_version=v"1.1",
                  target=resource(url),
+                 proxy=nothing,
                  parent=nothing, iofunction=nothing, kw...) where Next
 
     defaultheader(headers, "Host" => url.host)
@@ -36,6 +37,11 @@ function request(::Type{MessageLayer{Next}},
         elseif method == "GET" && iofunction isa Function
             setheader(headers, "Content-Length" => "0")
         end
+    end
+
+    if proxy != nothing
+        target = string(url)
+        url = URI(proxy)
     end
 
     req = Request(method, target, headers, bodybytes(body);


src/ConnectionPool.jl
```diff
@@ -488,6 +488,15 @@ end
 function getconnection(::Type{SSLContext},
                        host::AbstractString,
                        port::AbstractString;
+                       kw...)::SSLContext
+
+    port = isempty(port) ? "443" : port
+    @debug 2 "SSL connect: $host:$port..."
+    tcp = getconnection(TCPSocket, host, port; kw...)
+    return sslconnection(tcp, host; kw...)
+end
```
samoconnor: Here I've split the second part of `getconnection` into a new `sslconnection` function that does the MbedTLS hookup. This is needed so that `sslupgrade` (below) can hook MbedTLS to an existing TCP connection after a `CONNECT` to a proxy.
Aside from being split in two, `getconnection` should do exactly what it did before.

src/ConnectionPool.jl
```diff
@@ -496,16 +505,21 @@ function getconnection(::Type{SSLContext},
...
     handshake!(io)
     return io
 end
 
+function sslupgrade(t::Transaction{TCPSocket},
+                    host::AbstractString; kw...)::Transaction{SSLContext}
+    tls = sslconnection(t.c.io, host; kw...)
+    c = Connection(tls)
+    return client_transaction(c)
+end
```
samoconnor: We reach down through the `Transaction` and the `Connection` to get to the raw TCP socket. (Same as what is done when upgrading to WebSockets).
We then create an SSLContext by calling `sslconnection`.
Since the caller was working with a `Transaction` object, they need the return value to be a `Transaction` object as well, so we wrap the new SSLContext in a `Connection`, then in an `Transaction`. At this point it looks exactly like a real HTTPS connection that might have come from the pool. Except that it is not in the pool. It is intended to be used once then GCed.

src/ConnectionRequest.jl
```diff
@@ -23,18 +23,47 @@ abstract type ConnectionPoolLayer{Next <: Layer} <: Layer end
...
         return request(Next, io, req, body; kw...)
+
     catch e
         @debug 1 "❗️  ConnectionLayer $e. Closing: $io"
         close(io)
         rethrow(isioerror(e) ? IOError(e, "during request($url)") : e)
+    finally
+        if proxy != nothing && target_url.scheme == "https"
+            close(io)
+        end
```
samoconnor: Avoid reusing the connection to the proxy because it got upgraded to be a HTTPS tunnel, so we can't use it to talk to the proxy anymore.

src/ConnectionRequest.jl
```diff
@@ -23,18 +23,47 @@ abstract type ConnectionPoolLayer{Next <: Layer} <: Layer end
...
+            @debug 1 "📡  CONNECT HTTPS tunnel to $target"
+            r = Request("CONNECT", target)
+            writeheaders(io, r)
+            startread(io)
+            readheaders(io, r.response)
+            if r.response.status != 200
+                return r
+            end
+            io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
+        end
```
samoconnor: To proxy HTTPS, we set the target to "host:port" and send a `CONNECT` request to establish a tunnel to the target. If that fails, just return the non-200 Response object. It will be converted into an exception by the ExceptionLayer, retried, etc as usual. If we get a 200 response, call `sslupgrade` to change the `Transaction{TCPSocket}` into a `Transaction{SSLContext}`.

src/ConnectionRequest.jl
```diff
@@ -23,20 +23,60 @@ abstract type ConnectionPoolLayer{Next <: Layer} <: Layer end
...
+end
+
+function connect_tunnel(io, target_url)
+    target = "$(URIs.hoststring(target_url.host)):$(target_url.port)"
+    @debug 1 "📡  CONNECT HTTPS tunnel to $target"
+    request = Request("CONNECT", target)
+    writeheaders(io, request)
+    startread(io)
+    readheaders(io, request.response)
+    return request.response
```
samoconnor: To proxy HTTPS, we set the target to "host:port" and send a `CONNECT` request to establish a tunnel to the target. If that fails, just return the non-200 Response object. It will be converted into an exception by the ExceptionLayer, retried, etc as usual. If we get a 200 response, call `sslupgrade` to change the `Transaction{TCPSocket}` into a `Transaction{SSLContext}`.

src/ConnectionRequest.jl
```diff
@@ -23,20 +23,60 @@ abstract type ConnectionPoolLayer{Next <: Layer} <: Layer end
...
+                                                                 default
+
+function tunnel_request(Next, io, target_url, req, body; kw...)
+    r = connect_tunnel(io, target_url)
+    if r.status != 200
+        return r
+    end
+    io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
+    return request(Next, io, req, body; kw...)
+end
```
samoconnor: To proxy HTTPS, we set the target to "host:port" and send a `CONNECT` request to establish a tunnel to the target. If that fails, just return the non-200 Response object. It will be converted into an exception by the ExceptionLayer, retried, etc as usual. If we get a 200 response, call `sslupgrade` to change the `Transaction{TCPSocket}` into a `Transaction{SSLContext}`.

samoconnor:
@amellnik, it would be great if you could try this out when you get a chance.

amellnik:
This doesn't work for me -- I get timeouts on http or https requests even if I specify the proxy.  I would also suggest that `proxy` should default to the `HTTP_PROXY` environment variable unless otherwise specified.

samoconnor:
Thanks for trying it out @amellnik,

Please verify that you're on the latest commit on this branch:
```bash
$ git log -1
commit f03a793a00f417ddef984993be87b5b8e34f850b (HEAD -> refs/heads/http_proxy, refs/remotes/origin/http_proxy)
Author: Sam O'Connor <sam@octech.com.au>
Date:   Fri Mar 9 09:08:47 2018 +1100

    clean up CONNECT tunnel code
```

Please set `DEBUG_LEVEL=2` in ` src/HTTP.jl` and kw `verbose=3` in your request call.

Please post the request call you are making, showing the `proxy=` setting that you're using and a log of what you see. See example here: https://github.com/JuliaWeb/HTTP.jl/issues/218#issuecomment-371428829

Does it work if you run a local `tinyproxy` ?

amellnik:
I think it's crucial that HTTP.jl uses any defined `HTTP_PROXY` by default.  People will eventually build a lot of packages on top of HTTP.jl, and if proxies are not handled automatically then every package will have to collect the proxy information from the end user and pass it down to HTTP.jl.

This is also consistent with how other similar packages handle it.  See for example [Requests (python)](http://docs.python-requests.org/en/master/user/advanced/#proxies), [R](https://support.rstudio.com/hc/en-us/articles/200488488-Configuring-R-to-Use-an-HTTP-or-HTTPS-Proxy), [cURL](https://stackoverflow.com/questions/9445489/performing-http-requests-with-curl-using-proxy), [Requests (js)](https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables), [Go](https://stackoverflow.com/questions/14661511/setting-up-proxy-for-http-client), [Ruby](https://docs.ruby-lang.org/en/2.0.0/Net/HTTP.html#class-Net::HTTP-label-Proxies) and so on.

amellnik:
I am on that commit.  After modifying DEBUG_LEVEL to 
```julia
using HTTP
HTTP.get("https://httpbin.org/ip"; proxy="<our proxy>:<port>", verbose=3)
```


```DEBUG: b937 🗑  Deleted:        💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 TCP connect: <our proxy>:<port>...
DEBUG: b937 🔗  New:            🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000414)
DEBUG: b937 👁  Start write:T0  🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000414)
DEBUG: b937 📡  CONNECT HTTPS tunnel to httpbin.org:
DEBUG: b937 👁  Start read: T0  🔗    0↑🔒    0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000414)
DEBUG: b937 ❗️  ConnectionLayer EOFError(). Closing: T0  💀    0↑🔒    0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🗣  Write done: T0  💀    1↑     0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 ✉️  Read done:  T0  💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🔄  Retry IOError(EOFError() during request(http://<our proxy>:<port>))
: GET /ip HTTP/1.1
DEBUG: b937 🗑  Deleted:        💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 TCP connect: <our proxy>:<port>...
DEBUG: b937 🔗  New:            🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000418)
DEBUG: b937 👁  Start write:T0  🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000418)
DEBUG: b937 📡  CONNECT HTTPS tunnel to httpbin.org:
DEBUG: b937 👁  Start read: T0  🔗    0↑🔒    0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000418)
DEBUG: b937 ❗️  ConnectionLayer EOFError(). Closing: T0  💀    0↑🔒    0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🗣  Write done: T0  💀    1↑     0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 ✉️  Read done:  T0  💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🔄  Retry IOError(EOFError() during request(http://<our proxy>:<port>))
: GET /ip HTTP/1.1
DEBUG: b937 🗑  Deleted:        💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 TCP connect: <our proxy>:<port>...
DEBUG: b937 🔗  New:            🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000424)
DEBUG: b937 👁  Start write:T0  🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000424)
DEBUG: b937 📡  CONNECT HTTPS tunnel to httpbin.org:
DEBUG: b937 👁  Start read: T0  🔗    0↑🔒    0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000424)
DEBUG: b937 ❗️  ConnectionLayer EOFError(). Closing: T0  💀    0↑🔒    0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🗣  Write done: T0  💀    1↑     0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 ✉️  Read done:  T0  💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🔄  Retry IOError(EOFError() during request(http://<our proxy>:<port>))
: GET /ip HTTP/1.1
DEBUG: b937 🗑  Deleted:        💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 TCP connect: <our proxy>:<port>...
DEBUG: b937 🔗  New:            🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000414)
DEBUG: b937 👁  Start write:T0  🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000414)
DEBUG: b937 📡  CONNECT HTTPS tunnel to httpbin.org:
DEBUG: b937 👁  Start read: T0  🔗    0↑🔒    0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000414)
DEBUG: b937 ❗️  ConnectionLayer EOFError(). Closing: T0  💀    0↑🔒    0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🗣  Write done: T0  💀    1↑     0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 ✉️  Read done:  T0  💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🔄  Retry IOError(EOFError() during request(http://<our proxy>:<port>))
: GET /ip HTTP/1.1
DEBUG: b937 🗑  Deleted:        💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 TCP connect: <our proxy>:<port>...
DEBUG: b937 🔗  New:            🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000418)
DEBUG: b937 👁  Start write:T0  🔗    0↑     0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000418)
DEBUG: b937 📡  CONNECT HTTPS tunnel to httpbin.org:
DEBUG: b937 👁  Start read: T0  🔗    0↑🔒    0↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0x0000000000000418)
DEBUG: b937 ❗️  ConnectionLayer EOFError(). Closing: T0  💀    0↑🔒    0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 🗣  Write done: T0  💀    1↑     0↓🔒   <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
DEBUG: b937 ✉️  Read done:  T0  💀    1↑     1↓    <our proxy>:<port>:0 ≣16 Base.Libc.WindowsRawSocket(Ptr{Void} @0xffffffffffffffff)
IOError(EOFError() during request(http://<our proxy>:<port>))
```

samoconnor:
> `proxy="<our proxy>:<port>"`

Are you actually saying `proxy="http://<our proxy>:<port>"` ?

What do you get if you try to get `http://httpbin.org/ip` (not http**s** ?) ?

(I'm assuming your proxy does not require authentication.)

samoconnor:
Unfortunately, due to the way verbose=3 works, it does not capture the conversation with the proxy before the tunnel is established. So, to see what is going on, please try the patch below. Hopefully it will show why your proxy doesn't like the request. My guess is that we're missing a header it requires.

```patch
--- a/src/IOExtras.jl
+++ b/src/IOExtras.jl
@@ -148,6 +148,7 @@ function Base.readuntil(io::IO,
     # Fast path, buffer already contains delimiter...
     if !eof(io)
         bytes = readavailable(io)
+@show String(bytes)
         if (l = find_delimiter(bytes)) > 0
             if l < length(bytes)
                 unread!(io, view(bytes, l+1:length(bytes)))
@@ -157,6 +158,7 @@ function Base.readuntil(io::IO,

         # Otherwise, wait for delimiter...
         buf = Vector{UInt8}(bytes)
+@show String(buf)
         while !eof(io)
             append!(buf, readavailable(io))
             if (l = find_delimiter(buf)) > 0
```

amellnik:
I must have made a typo when I tried this before -- this is working for http requests but not https requests.  I'm replacing our actual proxy information with the stuff in angle brackets when pasting the results in.

samoconnor:
Regarding the HTTP_PROXY env var, I realise that there are many high level interface libraries that support the env var. e.g. in python, Lib/urlib supports the env var but Lib/http does not.
I guess the question is: Is HTTP.jl trying to be a high level URL downloader or just a HTTP protocol implementation. There are uses cases where high-level client stuff like cookies, default user agent strings, proxies, basic auth, are not needed. e.g. interfacing with REST APIs. I think it would be a good thing for Julia to have a seperate package for that stuff.
That being said, I guess that there's no harm in adding HTTP_PROXY env var support now and splitting it into a higher level package (along with cookies, etc) later on.

samoconnor:
@amellnik, it would be interesting to know what the `@show` patch above yields for your https request if you get a chance.

samoconnor:
@quinnj Given that this works in some cases (http & https with tinyproxy, but only http with @amellnik's real world proxy). I'm inclined to merge this PR as-is and real with remaining issues as bugs.
I'm worried that the alternative is that it may sit and go stale and end up needing rebasing given that proxying is not a high priority use case for you or me.

samoconnor:
It wouldn't be too hard to make a travis compatible test that grabs a binary from here https://github.com/tinyproxy/tinyproxy/releases.
... not to hard, but not a priority right now.

PR #228, opened 2018-04-02
Fix documentation to say HTTP.setstatus/HTTP.setheader

tk3369:
I was unable to run the example from the manual, and it turns out that setheader/setstatus are not exported.   This PR just adds these methods to the export statement.

From the manual:
```
    HTTP.listen() do http::HTTP.Stream
        @show http.message
        @show HTTP.header(http, "Content-Type")
        while !eof(http)
            println("body data: ", String(readavailable(http)))
        end
        setstatus(http, 404)
        setheader(http, "Foo-Header" => "bar")
        startwrite(http)
        write(http, "response body")
        write(http, "more response body")
    end
```

Error message without this PR:
```
I- Accept:  🔗    0↑     0↓    127.0.0.1:8081:0 ≣16
http.message = HTTP.Messages.Request:
"""
GET / HTTP/1.1
Host: localhost:8081
User-Agent: curl/7.58.0
Accept: */*

"""
HTTP.header(http, "Content-Type") = ""
E- UndefVarError(:setstatus)
|  catch_stacktrace() = StackFrame[(::##1#2)(::HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{TCPSocket}}) at REPL[2]:7, handle_stream(::##1#2, ::HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{TCPSocket}}) at Servers.jl:446, (::HTTP.Servers.##46#47{##1#2,HTTP.ConnectionPool.Transaction{TCPSocket},HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{TCPSocket}}})() at task.jl:335]
I- Closed:  💀    1↑     1↓🔒   127.0.0.1:8081:0 ≣16
```

### DIFF
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -1,7 +1,7 @@
 __precompile__()
 module HTTP
 
-export startwrite, startread, closewrite, closeread
+export startwrite, startread, closewrite, closeread, setstatus, setheader
 
 using MbedTLS
 import MbedTLS: SSLContext


quinnj:
I think I'd prefer we just update the docs to fully qualify the references, e.g. `HTTP.setstatus`. I also don't feel super strongly, so if the more people would rather export, I'd be ok. I think for consistency we haven't exported things in general though.

samoconnor:
I think I agree with @quinnj at this point.
I was thinking that it might be better to use `get|setproperty` ? but I'm not sure.
I vote for updating the doc and leaving this PR open as a reminder.

I think a proper decision requires a review of all the structs/types and their fields and get/set methods. Some should follow a pattern of complete-transparency (you are supposed to just access the fields e.g. URI), some should probably follow a pattern of complete hiding, some of the latter should use probably only pre-existing generic interfaces, some maybe should use `get|setproperty` field emulation. 

For now, requiring fully qualified `HTTP.setstatus` is unambiguous and makes possible later deprecation easier.

tk3369:
I'm good with that as well as long as it's consistent.  

BTW, it would be great if the docs can be updated with more examples - or perhaps there should be a tutorial/blog about how to properly build a web api using HTTP.jl.

PR #230, opened 2018-04-10
fix eval issue with precompile

SimonDanisch:


SimonDanisch:
If we could change `Expr[:(::Val{:next}), :(::Val{:path}), :(::Val{:to}), :(::Val{:greatness})]` to something like:
`Val{(:next, :path, :to, :greatness)}`, we could also get rid of the eval in register.

samoconnor:
@quinnj any thoughts on this?

PR #231, opened 2018-04-10
Add handle(::Router,::Stream)

EricForgy:
This PR will allow me to do the following:

```julia
HTTP.listen(ip"127.0.0.1",8000) do http
    if HTTP.WebSockets.is_upgrade(http.message)
        # Do websocket stuff
    else
        HTTP.handle(router[],http)
    end
end
```

### DIFF
--- a/src/Handlers.jl
+++ b/src/Handlers.jl
@@ -135,12 +135,12 @@ function gethandler(r::Router, req)
     return r.func(m, s, h, vals...)
 end
 
-function handle(r::Router, req)
+function handle(r::Router, req::HTTP.Request)
     handler = gethandler(r,req)
     # pass the request to the handler and return
     return handle(handler, req)
 end
 
-handle(hf::HandlerFunction, http::HTTP.Stream) = handle_request(hf.func,http)
+handle(h::Handler, http::HTTP.Stream) = HTTP.Servers.handle_request(req->handle(h, req), http) # FIXME: Consider moving
 
 end # module


src/Handlers.jl
```diff
@@ -141,6 +141,16 @@ function handle(r::Router, req)
...
+    request::HTTP.Request = http.message
+    request.body = read(http)
+    hf = gethandler(r,request)
+    request.response::HTTP.Response = hf.func(request)
+    startwrite(http)
+    write(http, request.response.body)
+    return
+end
+
+handle(hf::HandlerFunction, http::HTTP.Stream) = HTTP.Servers.handle_request(hf.func,http)
```
quinnj: This is the wrong dependency order; `Servers.jl` depends on `Handlers.jl`, not vice-versa. So we should define `handle_request` to be in `Handlers.jl` instead and have `Servers.jl` import/call it.

src/Handlers.jl
```diff
@@ -141,6 +141,16 @@ function handle(r::Router, req)
     return handle(handler, req)
 end
 
-handle(hf::HandlerFunction, http::HTTP.Stream) = handle_request(hf.func,http)
+function handle(r::Router, http::HTTP.Stream)
+    request::HTTP.Request = http.message
+    request.body = read(http)
+    hf = gethandler(r,request)
+    request.response::HTTP.Response = hf.func(request)
```
quinnj: This isn't quite right because it assumes the handler for a route is a `HandlerFunction`, whereas it could be any (even custom) Handler. More correct would be `... = handle(hf, request)`

src/Handlers.jl
EricForgy: Sounds good 👍 I can modify the PR accordingly.

EricForgy:
Thanks for the suggestion 👍 

The single method `handle(::Handler,::Stream)` is much cleaner and satisfies what I needed 😊

EricForgy:
I found an issue on the plane. Will commit a small revision when I get to my hotel :blush:

EricForgy:
Hi @samoconnor and @quinnj 👋 

I have updated my package (Pages) and now it is little more than some sugar on top of HTTP.jl. I am fully using HTTP.jl's router now, but I need this PR before I can tag a new version.

Any thoughts?

samoconnor:
LGTM, sorry about the delay

samoconnor:
Would this do the same thing?
```julia
handle(h::Handler, http::HTTP.Stream) = handle_request(req->handle(h, req), http)
```
(i.e., the new `handle(h::Handler, http::HTTP.Stream)` method is copy/pasted from `Servers.jl/handle_request(f::Function, http::Stream)`, but we could call it with an anon fn instead)

EricForgy:
@samoconnor, you asked:

> Would this do the same thing?

```julia
handle(h::Handler, http::HTTP.Stream) = handle_request(req->handle(h, req), http)
```

That is similar to what I had in an earlier version, but @quinnj made this comment above:

> This is the wrong dependency order; Servers.jl depends on Handlers.jl, not vice-versa. So we should define handle_request to be in Handlers.jl instead and have Servers.jl import/call it.

I like your idea about the anonymous function, but should we move `handle_request` to Handlers.jl? If so, should move the other `handle_*` methods?

EricForgy:
This might be too big of a change, but I went ahead and moved the `handle_*` methods from Servers.jl to Handlers.jl and then with `handle_request` in Handlers, defined the suggested:

```julia
handle(h::Handler, http::HTTP.Stream) = handle_request(req->handle(h, req), http)
```

This works for me in Pages.

samoconnor:
Sorry, I can't agree to moving `handle_connection`, `handle_transaction`, `handle_stream` or `handle_request` from Servers.jl to Handlers.jl. None of these functions depends on anything from Handlers.jl

My intention is that Servers.jl is a lower-level layer than Handlers.jl.

If anything I would move `struct Server` and `serve(server::Server, ...)` to Handlers.jl because these two things do depend on Handlers.jl. The `serve(server::Server, ...)` function is a thin wrapper that joins `listen` up to the Handlers.jl stuff. 

The core function in Servers.jl is `listen(f::Function, host, port)` (which calls `handle_connection`, `handle_transaction`, `handle_stream` and `handle_request`). All of this works without Handlers.jl and is designed to be able to be plugged into any request handling framework via the `f::Function` argument. i.e. When we split HTTP.jl into pieces, the `listen(f::Function, host, port)` stuff will stay in HTTP.jl (or maybe HTTPServer.jl) but the Handlers.jl stuff might end up as a stand-alone package, and might be a peer with other request handling/routing frameworks.

samoconnor:
Note: I have said that I don't want the `listen(f::Function, ...)` part of Servers.jl to depend on Handlers.jl.
It might also be quite reasonable for Jacob to not want Handlers.jl to depend on Servers.jl, i.e. if Handlers.jl is intended to be able to work with any source of requests, not just HTTP.Servers.
In this case there should probably be a HandlerServers.jl that depends on both and glues them together.

EricForgy:
No problem :)

I do need a handle(::Handler, ::Stream).

What are your thoughts on @quinnj's concern about having Handlers depend on Servers? Should we just accept that for now?

EricForgy:
It seems the path of least resistant is your suggested handle(::Handler, ::Steam) for now :)

samoconnor:
Also note the conversation here: https://github.com/JuliaWeb/HTTP.jl/pull/234

It seems that design work and revision will continue for a while in the handler-framework / middleware space, and that there are a bunch of people thinking about different ways to solve the problems.

To me this suggests that we should split `struct Server`, `serve(server::Server, ...)` and `Handlers.jl` into a seperate package sooner rather than later so it can evolve quickly while at the same time we try to avoid architectural changes to HTTP.jl and concentrate on making it do the lower level simple things in a bullet-proof way.

EricForgy:
Sounds good. If there is anything I can do to help with that, I can use some direction, but happy to try.

samoconnor:
Timing may depend a bit on where @quinnj ends up with the logging enhancements in https://github.com/JuliaWeb/HTTP.jl/pull/200. i.e does the logging stuff end up fitting nicely with the `listen` stuff, so it stays in HTTP.jl, or does it end up intertwined with the Handlers stuff and fit more naturally there. The evolution of the new logging infrastructure in Base may influence this too.

EricForgy:
Ok. This commit should be fine I hope 🙏😊

EricForgy:
Any thoughts? 😊

EricForgy:
**bump** 😅

EricForgy:
🙏

EricForgy:
Thank you 😊

Issue #192, opened 2018-02-09
Error setting tlsconfig


josefsachsconning:
This used to work (not sure with which versions of Julia and packages), but no longer does.
```
AWS-Sachs-Ubuntu$ julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.1-pre.94 (2017-10-11 17:18 UTC)
 _/ |\__'_|_|_|\__'_|  |  conning-0.6/3a6ac23b89 (fork: 2 commits, 125 days)
|__/                   |  x86_64-linux-gnu

julia> using HTTP, MbedTLS

julia> client = HTTP.Client(tlsconfig=MbedTLS.SSLConfig(false))
ERROR: MethodError: Cannot `convert` an object of type Tuple{Symbol,MbedTLS.SSLConfig} to an object of type Pair{Symbol,Any}
This may have arisen from a call to the constructor Pair{Symbol,Any}(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] copy!(::IndexLinear, ::Array{Pair{Symbol,Any},1}, ::IndexLinear, ::Array{Any,1}) at ./abstractarray.jl:656
 [2] convert(::Type{Array{Pair{Symbol,Any},1}}, ::Array{Any,1}) at ./array.jl:373
 [3] HTTP.Client(::Dict{String,Set{HTTP.Cookies.Cookie}}, ::Array{Any,1}) at /home/sachs/.julia/v0.6/HTTP/src/client.jl:26
 [4] (::Core.#kw#Type)(::Array{Any,1}, ::Type{HTTP.Client}) at ./<missing>:0
```

samoconnor:
Hi @josefsachsconning,

Can you try using the new kw args to `request` instead of using the `Client` type?
e.g.
```julia
HTTP.request("GET", "https://foobar.com"; sslconfig = MbedTLS.SSLConfig(false))
or
HTTP.get("https://foobar.com"; require_ssl_verification = false)
```

or you can put your favourite options in a variable like this:
```julia
httpconf = [:require_ssl_verification = false,
            :verbose => 1,
            :readtimeout => 120]

HTTP.get("https://foobar.com/bar", httpconf...)
[...]

PR #232, opened 2018-04-10
make http client based get requests work again

amanica:
hi, 
after struggling quite a bit with the new version of HTTP.jl (and julia 0.6.2) I finally figured out that the query parameters gets ignored when calling via the Client api. (I think a similar issue exists with HTTP.get function and friends)

With this fix it is working for me again. Let me know if you are happy with this and if I should attempt to add a test..

thanks
marius

### DIFF
--- a/src/client.jl
+++ b/src/client.jl
@@ -39,6 +39,7 @@ global const DEFAULT_CLIENT = Client()
 function request(client::Client, method, url::URI;
                  headers=Header[],
                  body="",
+                 query=nothing,
                  enablechunked::Bool=true,
                  stream::Bool=false,
                  verbose=false,
@@ -47,6 +48,9 @@ function request(client::Client, method, url::URI;
     # Add default values from client options to args...
     if VERSION > v"0.7.0-DEV.2338"
     args = merge(client.options, args)
+    if query != nothing
+       url = merge(url, query=query)
+    end
     getarg = Base.get
     else
     for option in client.options


quinnj:
Looks good to me, but @samoconnor implemented this code path, so let's let him look at it.

samoconnor:
This change looks good to me in the context of `client.jl`.

However, I think it might be time to deprecate `client.jl` and encourage people to use kw args like this: https://github.com/JuliaWeb/HTTP.jl/issues/192#issuecomment-364599326

As it stands, the `request(client::Client ...` method exists as a place to implement all the deprecation messages for kw args that changed in v0.6. The only thing remaining in `struct Client` (aside form the kw args `NamedTuple`) is the cookiejar. But `cookiejar=` is really just a kw option like all the others, there isn't a compelling reason for it to have its own special field.

Note: [`default_cookie_jar`](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/CookieRequest.jl#L10) handles the common case where the user just wants a single cookiejar per Julia session, so `DEFAULT_CLIENT.cookies` is redundant.

Also, `HTTP.get(url)` being routed through `client.jl` is a bug/oversight (a multiple dispatch gotcha).
It is intended to go to here: https://github.com/JuliaWeb/HTTP.jl/blob/master/src/HTTP.jl#L337

samoconnor:
This commit addresses the routing of `HTTP.get(url)` though `client.jl`: https://github.com/JuliaWeb/HTTP.jl/commit/5d5614c9b65a4bfa66c730b373b8a160b535f012
(@quinnj can you do a quick sanity check?)

amanica:
> This commit addresses the routing of HTTP.get(url) though client.jl: 5d5614c
(@quinnj can you do a quick sanity check?)

I gave it a spin and it seems to work file, thanks!

PR #234, opened 2018-04-17
Add a context field to requests

amellnik:
As per discussions with @JockLawrie and @quinnj, I suggest that we add an additional `context` field to `HTTP.Request` of type `Dict{String, Any}`.  The goal of this is to hold additional information related to the request that doesn't fit in the structure of the current `Request` type, such as authentication and session information.  

Many web frameworks in other languages allow something like this: [Go](https://joeshaw.org/revisiting-context-and-http-handler-for-go-17/), Node/Express, [Python/Flask](https://pythonhosted.org/Flask-Session/), [Ruby](http://www.rubydoc.info/docs/rails/4.1.7/ActionDispatch/Request/Session), etc.  It complicates the `Request` object slightly, but makes it easier for downstream packages.

### DIFF
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -182,6 +182,9 @@ Represents a HTTP Request Message.
 - `parent`, the `Response` (if any) that led to this request
   (e.g. in the case of a redirect).
    [RFC7230 6.4](https://tools.ietf.org/html/rfc7231#section-6.4)
+
+- `context::Dict{String, Any}`, allows storage of additional information
+  related to the request.
 """
 mutable struct Request <: Message
     method::String
@@ -192,12 +195,14 @@ mutable struct Request <: Message
     response::Response
     txcount::Int
     parent
+    context::Dict{String, Any}
 end
 
 Request() = Request("", "")
 
 function Request(method::String, target, headers=[], body=UInt8[];
-                 version=v"1.1", parent=nothing)
+                 version=v"1.1", parent=nothing,
+                 context=Dict{String, Any}())
     r = Request(method,
                 target == "" ? "/" : target,
                 version,
@@ -205,7 +210,8 @@ function Request(method::String, target, headers=[], body=UInt8[];
                 bytes(body),
                 Response(0),
                 0,
-                parent)
+                parent,
+                context)
     r.response.request = r
     return r
 end


samoconnor:
At first glance this looks like something I'd be quite strongly opposed to.
It looks like: _"Add a `context` field to `struct X` to hold additional information related to `X`"_.

Perhaps I just don't understand the use case well enough, but it seems that this proposal could be applied to any type `X`. In a language like JavaScript, you can attache arbitrary dynamic attributes to any object, but Julia is attempting to take types more seriously than JavaScript. Why should `HTTP.Request` have a handy `context` scratchpad any more than `TCPSocket`, or `Array`?

If middleware web framework layer code needs extra information about a request, this should be handled by composition, i.e. it should declare a `struct WebMiddleware.Request` to hold whatever information is needed. The `HTTP.Request` could be a field of that struct (or the relationship could be separately mapped) there are plenty of design patterns for extending the functionality of a lower layer without having to redefine the lower layer's data structures. Pushing ad-hoc application state down into the `HTTP.Request` layer is likely to eventually result in something akin to the fragile base class problem. 

I certainly don't think a generic `context` field belongs at the `HTTP.Request` layer.
It might make sense to have a `struct HTTP.Handlers.Request` with a `context` field, but there would have to be a pretty good argument as to why the application layer shouldn't define its own specialised type which extends HTTP.Handlers.Request by composition, or though an abstract interface type.

Keep in mind that `struct HTTP.Request` is simply intended to modal a "HTTP Request Message" per https://tools.ietf.org/html/rfc7230#section-2.1, it is not intended to model the application specific relationships that refer to the message.
https://github.com/JuliaWeb/HTTP.jl/blob/5d5614c9b65a4bfa66c730b373b8a160b535f012/src/HTTP.jl#L487-L500

JockLawrie:
Agreed Sam. Put another way, I think it makes sense to keep separate what the client knows (`Request`) from what the server knows (context, such as session info). That is, keep these 2 pieces of data in 2 separate objects. I'm not sure where in the stack they should be brought together.

amellnik:
These are all fair points.  

My primary concern is figuring out how to build an ecosystem of reusable server-side components (like [these in express](http://expressjs.com/en/resources/middleware.html) or [these in go](https://github.com/urfave/negroni#third-party-middleware)) without a bunch of competing interfaces, and adding an optional field to this type in HTTP itself seemed like a low-cost way to enable it.   

Like Sam said, I can define something like

```julia
# in Joseki
mutable struct Req_Con
    request::HTTP.Request
    context::Dict{String, Any}
end
```

and I can always rewrite all my middleware layers that don't need `context` and similar to use this type (and even also work with a normal `HTTP.Request`).  Now suppose I want to create some middleware in a separate package that does some authentication and looks up a user's group memberships.  To make it compatible with the approach above it would need to look like

```julia
function lookup_ad_groups!(rc::Joseki.Req_Con)
    # Look up groups and store them in rc.context
end
```

Any project that wants to use this would then need to buy in to Joseki.  This is what I was hoping to avoid since HTTP.jl already includes most of the tools people need to construct servers.  

Would would be the recommended method for creating reusable components in this case?

JockLawrie:
Is it sufficient to simply make the server-side data/code (such as your AD lookup function) visible to the handlers by importing it into the module in which the handlers are defined?

This approach requires a little care to set up, but is only required for middleware and is therefore a one-off effort.

I'm currently doing this for a simple service and it's fine for my project. Before introducing new types, perhaps we should see how this approach fares with other projects?

samoconnor:
> Would would be the recommended method for creating reusable components in this case?

I'd recommend:
 - defining interfaces in terms of abstract types and informal protocols
 - thinking about functions as a more important abstraction than struct fields
 - avoid overuse of specialised methods with concrete argument types, rather use them as a fallback for cases where duck-typing and informal protocols is not enough
 - figuring what vertical layers your framework will have and clearly defining the purpose of each and the interfaces up and down
 - if you're taking inspiration from other frameworks, try to figure out what mechanisms they're using for polymorphism/genericity/plugability/composition/inheritance/reuse. Try to figure out how the implementation language has influenced these patterns and how the patterns could best be mapped to the patterns that Julia is best at.
 - if the interface of a dependancy doesn't fit your model, don't change it, wrap it with your preferred interface. (If the model turns out to be awesome in practice, it will be an easy sell to push the change into the dependancy later).
 - build an example system with several components at each layer
 - write the user documentation that explains how to add a new component at each layer.
 - review all of the above and redo anything that doesn't make sense now that you've learned more stuff.
 - repeat until you no longer have the urge to redo anything.

JockLawrie:
I agree we should just try some ideas and see what works best.

samoconnor:
Oh, and another thing:
 - When you find the need for composition, (or inheritance-by-burte-force, i.e. adding a field to a base type), consider just passing around two objects, or a `Tuple`, or even a `NamedTuple`.
 - In Julia it is fine to return more than one value from a function (see the Base iteration protocol).
 - This can get out of hand, you don't want functions that take dozens of positional arguments, but passing around request and context might be fine, or having `context=` as an optional kw arg.

amellnik:
I'll give it a shot.

PR #238, opened 2018-04-19
0.7 compat

quinnj:


quinnj:
Ok, this is almost ready to go; I'm still seeing a WebSocket error like
```julia
WebSockets: Error During Test at /home/chronos/user/.julia/dev/HTTP/test/WebSockets.jl:5
  Got exception HTTP.WebSockets.WebSocketError(1002, "") outside of a @test
  HTTP.WebSockets.WebSocketError(1002, "")
filesize("result_file") = 7440
```
@EricForgy, does that mean anything to you? I'm seeing that error on latest 0.7 master (alpha tag). If you have a chance, could you look into it? Otherwise I can try and dig in to figure out what's going on.

PR #239, opened 2018-04-27
Servers: support all IOServer providers

vtjnash:
This enables support for PipeServer, rather than type-asserting that it can only be used with TCPServer.

I didn't yet add support for this on the connection side, to be able to test it. But, for example, nodejs can connect to this:
```
julia> HTTP.Servers.serve("testsocket")

> http.get({socketPath:"testsocket"}, (res) => {
... console.log(res.statusCode);
... res.setEncoding('utf8');
... res.on('data', (chunk) => { console.log(chunk) })
... })
```

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -70,8 +70,8 @@ mutable struct Connection{T <: IO}
     port::String
     pipeline_limit::Int
     idle_timeout::Int
-    peerport::UInt16
-    localport::UInt16
+    peerport::UInt16 # debug only
+    localport::UInt16 # debug only
     io::T
     excess::ByteView
     sequence::Int
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -140,6 +140,7 @@ end
 @enum Signals KILL
 
 serve(server::Server, host::IPAddr, port::Integer, verbose::Bool) = serve(server, Sockets.InetAddr(host, port), verbose)
+serve(server::Server, host::AbstractString, port::Integer, verbose::Bool) = serve(server, parse(IPAddr, host), port, verbose)
 serve(server::Server{T, H}, host::AbstractString, verbose::Bool) where {T, H} = serve(server, String(host), verbose)
 function serve(server::Server{T, H}, host::Union{Sockets.InetAddr, String}, verbose::Bool) where {T, H}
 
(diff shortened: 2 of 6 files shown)


samoconnor:
I haven't tried this out, but it looks good to me.
The only point I have to raise is that we loose `HTTP.listen(s)` where `s` is a string accepted by `getaddrinfo` (because `s` is now assumed to be a named socket).
Probably not a big deal, but maybe a bit surprising if you wrote `HTTP.listen("www1.octech.com.au")` and get no error, and a named pipe in `pwd`.

`Sockets.listen(path::AbstractString)` is consistent with other places in Base where a string is assumed to mean a file path, but when it comes to networking it wouldn't be unreasonable to assume that a string would be interpreted as a DNS name.

Perhaps there should be a `Sockets.SocketName` type (peer of `Sockets.InetAddr`) and no `listen(::AbstractString)` method so that the user is forced to be explicit about what type of server they want to create.

vtjnash:
It would be weird to ask a DNS server for what static IP address you told it you have – and it won't usually even return the right answer.

samoconnor:
> weird to ask a DNS server for what static IP address you told it you have

If guess, if "you" told it.
But if "you" are a short-lived container instance, you might only know that you are supposed to listen on "replica2.myapp.com", "publicapi.myapp.com" and "intranetcallback.myco.com", maybe the virtualisation platform and its DHCP/DNS server are the only place that knows the IPs.

I can't imagine anyone would ever write code like `HTTP.listen(ip"192.68.0.1")` except in an example or a throw-away hack. Any real code that listens on a specific IP address will get that address from a config file, or an environment variable, or a command line argument or similar. All these things are likely to be strings.

i.e. It seems likely that this change will result in almost every non-trivial call to `HTTP.listen` changing from `HTTP.listen(ENV["MY_LISTEN_ADDRESS"])` to `HTTP.listen(getaddrinfo(ENV["MY_LISTEN_ADDRESS"]))`. This is fine in itself, but the simlutaneous existence of a `HTTP.listen(::String)` method that expects a named pipe/socket seems likely to cause confusion.

vtjnash:
If you don't have a stable IP address, I expect typical usage in that case is to call `HTTP.listen(ip"0.0.0.0")` – why would you want to peer on a snapshot of one dynamic address chosen at random, rather than on all addresses?

The other typical usage is to call `HTTP.listen(Sockets.localhost)` (aka `HTTP.listen()`)

samoconnor:
I guess I'm imagining a node that is connected to multiple seperate IP networks with different addresses (e.g. a storage network, an internal private network and the public internet).

Re `HTTP.listen(ip"0.0.0.0")`, maybe this is a better default than `HTTP.listen(Sockets.localhost)`.

vtjnash:
> Re HTTP.listen(ip"0.0.0.0"), maybe this is a better default than HTTP.listen(Sockets.localhost).

Secure is a better default than globally internet accessible.

I added another method to the PR, such that if it gets a pair, it'll always interpret it as ip-address: `("ipaddr", 80)`

samoconnor:
@vtjnash Looks good. I'm happy for this to be merged.

It still feels not quite right for `[Sockets|HTTP].listen(::AbstractString)` to assume that strings are named pipes. However, that is really a seperate issue and HTTP should follow whatever pattern Base.Sockets sets.

A couple of options would be:
 - Remove `Sockets.listen(::AbstractString)` and add a new type: `Sockets.SocketName`.
 - Keep `Sockets.listen(::AbstractString)` and interpret the string as a URI (`net.pipe://localhost/mypipe`, `file:///var/mypipe`, `http://0.0.0.0:8080`, `http://127.0.0.1:80`)

Issue #246, opened 2018-06-04
WebSocket test failure on 0.7


quinnj:
As documented here: https://github.com/JuliaWeb/HTTP.jl/pull/238

quinnj:
I believe this may have been due to a change in the server we're pinging for websocket tests, since I checked out 5d5614c9b65a4bfa66c730b373b8a160b535f012, which had known passing tests (according to travis logs), but which currently has the same test failures we're seeing on PRs/master.

quinnj:
Note that the failing test has been commented out for now to ensure passing package tests; given this functionality is newer and not extremely robust, I think we're ok allowing failures on master for the moment, but it would be great if @EricForgy or @samoconnor could look into why this test is now failing on 0.7 and how to fix.

samoconnor:
Tried this just now on 1.0, seems to pass:
```julia
$ julia10 ./test/WebSockets.jl
writing
testing
writing again
closewrite
reading response
writing
testing
writing again
closewrite
reading response
[ Info: Listening on: Sockets.InetAddr{IPv4}(ip"127.0.0.1", 0x1f95)
Testing local server...
[ Info: Accept:  🔗    0↑     0↓    0s 127.0.0.1:8085:8085 ≣16
[ Info: Closed:  💀    1↑     1↓🔒   0s 127.0.0.1:8085:8085 ≣16
Test Summary: | Pass  Total
WebSockets    |    8      8
```

samoconnor:
6f454ba3a7e1b654f02baa65b5f16256978ed04f

samoconnor:
Travis passing: https://travis-ci.org/JuliaWeb/HTTP.jl/jobs/422319666

PR #249, opened 2018-06-05
Add option to allow servers to listen on the same port (as long as th…

quinnj:
…e system allows it)


quinnj:
Alright, the only failures here are https://github.com/JuliaWeb/HTTP.jl/issues/246. Any concerns here @samoconnor, @vtjnash? I'd also request a special review from @tanmaykm since he did some of the original work w/ this in Base.

tanmaykm:
LGTM! :+1:

Issue #263, opened 2018-06-30
Feature request: parse mulitpart form data into a dict


piever:
Julia 0.6.3
HTTP.jl 0.6.12
MbedTLS.jl 0.5.11

This has been mentioned in #59, but it seems like the issue was closed without a solution for the thing I'd like to do. What I gather from that issue is that if one is sending a request in Julia to a Julia server, then the recommended way is for the request body to be a JSON string which will be sent as is. However, if the request comes from an actual form in a website, then the body will definitely be a multipart form. 

I am trying to understand what is the best way to analyze that kind of data in Julia once it reaches the server, meaning are there some tools to decompose a multipart form back into a dictionary so that one can work with it in Julia?

If this doesn't exist and nobody is working on it, I'm happy to give it a go (though I have very little experience with string processing, so I may take a while). What would the easiest way to do this be? Just get the boundary from the header, then split the body by the boundary and use regexes to get all the relevant information?

fredrikekre:
#514, #745

PR #264, opened 2018-07-01
WIP/RFC: added parser for multipart form

piever:
This is an attempt to solve #263. I've tested with a simple form with a textbox and a file input (I updated a PNG) and it seems to work fine (still need to add tests though).

The "content" of each section add a newline and a couple of dashes at the end that I'm removing (`\r\n--`, though the type of newline may vary I guess), though maybe I'm not 100% correct in how I do so (on whether they should be removed at all): I'm taking out the match of  `r"[\r\n]*-*$"`

### DIFF
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -3,7 +3,7 @@ function parse_multipart_chunk!(chunk, d)
     Parsers.exec(re, chunk) || return nothing
     i, j = re.ovec
     description = String(chunk[1:i])
-    content = chunk[nextind(chunk, j):end]
+    content = Parsers.nextbytes(re, chunk)
 
     v = match(r"Content-Disposition: form-data; name=\"(.*)\"; filename=\"(.*)\"[\r\n]+Content-Type: (\S*)", description)
     if v !== nothing
@@ -23,7 +23,11 @@ function parse_multipart_chunk!(chunk, d)
     push!(d["content"], content)
 end
 
+parse_multipart_body(body::AbstractArray{UInt8}, boundary) =
+    parse_multipart_body(String(body), boundary)
+
 function parse_multipart_body(body, boundary)
+    body = String(body)
     dict = Dict(
         "name" => String[],
         "filename" => Union{String, Void}[],
@@ -32,11 +36,11 @@ function parse_multipart_body(body, boundary)
     )
     re = Regex(boundary)
     Parsers.exec(re, body) || return nothing
-    str = body[nextind(body, re.ovec[2]):endof(body)]
+    str = Parsers.nextbytes(re, body)
     while Parsers.exec(re, str)
         chunk = str[1:re.ovec[1]]
         parse_multipart_chunk!(chunk, dict)
-        str = str[nextind(str, re.ovec[2]):endof(str)]
+        str = Parsers.nextbytes(re, str)
     end
     return dict
 end
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -1,6 +1,6 @@
-function find_boundary(bytes::Vector{UInt8}, str, dashes)
+function find_boundary(bytes::AbstractVector{UInt8}, str, dashes; start::Int = 1)
     l = length(bytes)
-    i = 1
+    i = start
     cons_dash = 0
     while i+length(str) <= l
         @inbounds cons_dash = (bytes[i] == 0x2d) ? cons_dash + 1 : 0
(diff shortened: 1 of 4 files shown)


piever:
I'm also not sure whether I can be certain that `req.body::Array{UInt8}` or if it could also happen to a be a `String` or something else.

piever:
Actually, it seems that it's fine to work with strings, the continuation byte is only a problem in Julia 0.6

```julia
julia> v = [0x89, 0x00]
2-element Array{UInt8,1}:
 0x89
 0x00

julia> s = String(v)
2-byte String of invalid UTF-8 data:
 0x89
 0x00

julia> SubString(s, 1, 2)
ERROR: ArgumentError: invalid SubString index
```

whereas in Julia 0.7 this works fine,

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+end
+
+parse_multipart_body(body::AbstractArray{UInt8}, boundary) =
+    parse_multipart_body(String(body), boundary)
+
+function parse_multipart_body(body, boundary)
+    body = String(body)
+    dict = Dict(
+        "name" => String[],
+        "filename" => Union{String, Void}[],
```
quinnj: Use `Nothing` instead of `Void`

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+    while Parsers.exec(re, str)
+        chunk = str[1:re.ovec[1]]
+        isempty(chunk) || parse_multipart_chunk!(chunk, dict)
+        str = Parsers.nextbytes(re, str)
+    end
+    return dict
+end
+
+function parse_multipart_form(req::Request)
+    m = match(r"multipart/form-data; boundary=(.*)$", req["Content-Type"])
```
quinnj: This will throw a KeyError if the `Request` doesn't have a `Content-Type` header; it seems like we'd rather return `nothing` instead?

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+        push!(d["contenttype"], nothing)
+    end
+
+    re = r"(?:[\r\n]*)(?:-*)$"
+    Parsers.exec(re, content) && (content = content[1:re.ovec[1]])
+    push!(d["content"], content)
+end
+
+parse_multipart_body(body::AbstractArray{UInt8}, boundary) =
+    parse_multipart_body(String(body), boundary)
```
quinnj: Note that in 0.7 this is destructive; i.e. w/ `body` as a `Vector{UInt8}`, it will be turned into a `Vector{UInt8}(0)` after the String is created. Not saying that's necessarily bad, but if someone tried to access `body` again after calling this, it would not be what they expected. We should at least have docs that clearly spelled this behavior out.

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+
+function parse_multipart_body(body, boundary)
+    body = String(body)
+    dict = Dict(
```
quinnj: I feel like my personal preference would be to return a `Vector{Multipart}` where `Multipart` was a simple struct like:
```julia
struct Multipart
    name::String
    filename::Union{Nothing, String}
    contenttype::Union{Nothing, String}
    content::Vector{UInt8}
end
```
It just seems like it would be a bit simpler on the server side if I could just call this function and iterate over the multparts, dealing w/ each as needed, instead of needing to do something like:
```julia
for i in 1:length(multiparts)
    nm = multiparts["name"][i]
    filename = multiparts["filename"][i]
    # etc....
end
```

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
+function parse_multipart_chunk!(chunk, d)
+    re = r"(\n){2,}|(\r){2,}|[\r\n]{4,}"
```
quinnj: It might be a little clearer to put all the regexes as top-level `const BOUNDARY_REGEX = r"(\n){2,}...`, so it's clear they're const and they have more descriptive names.

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+    content = Parsers.nextbytes(re, chunk)
+
+    v = match(r"Content-Disposition: form-data; name=\"(.*)\"; filename=\"(.*)\"[\r\n]+Content-Type: (\S*)", description)
+    if v !== nothing
+        push!(d["name"], v[1])
+        push!(d["filename"], v[2])
+        push!(d["contenttype"], v[3])
+    else
+        v = match(r"Content-Disposition: form-data; name=\"(.*)\"", description)
+        v == nothing && return
```
quinnj: This `v == nothing && return` makes me uncomfortable because we've basically gotten into a bad state and we're just returning. I think we probably need to read up on the spec to see exactly what is and isn't required in terms of the `Content-Disposition: form-data;` tags and make sure we're following that. Is `name` _always required? What if just `name` and `Content-Type` are present? Seems like we want to make sure we're accounting for what the spec allows here.

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+        "filename" => Union{String, Void}[],
+        "contenttype" => Union{String, Void}[],
+        "content" => Any[]
+    )
+    re = Regex(boundary)
+    Parsers.exec(re, body) || return nothing
+    str = Parsers.nextbytes(re, body)
+    while Parsers.exec(re, str)
+        chunk = str[1:re.ovec[1]]
+        isempty(chunk) || parse_multipart_chunk!(chunk, dict)
```
quinnj: What's the scenario where `isempty(chunk)` is true? It seems like if `Parsers.exec(re, str)` returns true, then it indeed matched a boundary, and yet this is saying that we didn't find the boundary? Maybe the regex needs a tweak here?

src/parsemultipart.jl
piever: I see, so I should be coding with just Julia 0.7 in mind?

src/parsemultipart.jl
piever: Agreed, at least the name should have a `!`, will try to run this code in 0.7 (I still haven't transitioned)

src/parsemultipart.jl
piever: If `re` matches the beginning of the string, then I guess we'd have `str[1:re.ovec[1]] == ""` ? But I agree that this should actually not happen in practice and `parse_multipart_chunk!` should probably still work when given an empty string.

src/parsemultipart.jl
quinnj: Definitely. HTTP.jl has stayed master-compatible since 0.6 branched.

src/parsemultipart.jl
```diff
@@ -0,0 +1,52 @@
...
+function parse_multipart_body(body, boundary)
+    body = String(body)
+    dict = Dict(
+        "name" => String[],
+        "filename" => Union{String, Void}[],
+        "contenttype" => Union{String, Void}[],
+        "content" => Any[]
+    )
+    re = Regex(boundary)
+    Parsers.exec(re, body) || return nothing
```
samoconnor: If the`Parsers` module is only being used for `exec`, `nextbytes` etc then this file should just do `include("parseutils.jl")` instead of saying `Parsers.exec(...`.

samoconnor:
A few things it would be good to have:

- [ ] Description (in docstrings) of what this PR implements in terms of the RFC(s).
- [ ] Description of what parts of the RFC(s) that the implementation does not cover (if any).
- [ ] Tests written to cover whatever behaviour is specified in the RFC(s).
- [ ] User API docstrings and examples.
- [ ] Quick comparison with how other HTTP libraries do this API.

src/parsemultipart.jl
samoconnor: I would say that at this point, with 0.7 beta out, feature addition PRs for HTTP.jl could(should?) be 0.7-only. i.e. we should put `julia 0.7` in `REQUIRE` soon anyway.

piever:
Yes, I still need to add tests and documentation.

I'm definitely not an expert on internet standards, is there some good source where I could see what I should expect to be present or not in a multipart form body?

samoconnor:
Some starting points for specifications:
https://tools.ietf.org/html/rfc7231#section-3.1.1.4
https://tools.ietf.org/html/rfc2046#section-5.1
https://tools.ietf.org/html/rfc7578

src/parsemultipart.jl
piever: It seems that `req[str]` returns `""` if the key is missing, so this should work fine.

src/parsemultipart.jl
piever: What's the alternative? `String(copy(body))` ?

piever:
It's probably better to develop this only for 0.7 also because string processing seems to have change in ways I'm not sure I understand. When using:

`HTTP.Parsers.exec(re, body)`

I keep getting a `ERROR: LoadError: PCRE.exec error: UTF-8 error: isolated byte with 0x80 bit set` which I'm not sure how to get rid of. If needed I can upload a file with a saved body of an example request that generates this error. Is `exec(re, body)` the preferred way to separate the body in chunks according to the boundary or am I missing some better alternative?

src/parsemultipart.jl
piever: There already is a `Multipart` struct which I'd like to reuse, except it stores the content as `IO` object, whereas for me a `Array` (actually `SubArray`) of `UInt8` would be better.  Should I simply write my `SubArray` to a `IOBuffer` and put that instead?

piever:
Still needs tests and docs but should have addressed most things. Now I no longer convert to strings before parsing but work directly with vectors of bytes, as otherwise it screws up if the `UInt8` encoding of the file has invalid characters.

In terms of header handling, I though the following seemed best according to the RFC linked above (especially https://tools.ietf.org/html/rfc7578):

- The header has to say `Content-Disposition: form-data`
- The `name` field is necessary
- `Content-Type` is not necessary: omitting it defaults to `text/plain`
- `filename` is not necessary

The main difference with the `Multipart` type as implemented originally in this package are:

- `Multipart` has no `name` field: I've added it and I think it should be considered in `writemultipartheader` function
- `Multipart` assumes that data is of type `IO`: so far I've simply put my `SubArray{UInt8}` in a `IOBuffer`
- `Multipart` assumes that `filename` is necessary whereas I don't think it should be, especially because some fields of the form may not be files
- there is some inconsistency about how to handle missing fields: I've been using `nothing` and HTTP uses `""`, maybe it's simpler to always use `""` esp. for backward compatibility
- unlike `Multipart` struct I don't really have the field `Content-Transfer-Encoding:` and looking at https://tools.ietf.org/html/rfc7578 it seems like it's deprecated, so I'm not sure what to do about that

I'm curious to know what's the best way to proceed on the above points.

acdupont:
Is this PR abandoned?

quinnj:
Superseded by https://github.com/JuliaWeb/HTTP.jl/pull/427

Issue #261, opened 2018-06-26
Errors in subsequent requests caused by connection reuse, 204 responses combined with chunked Transfer Encoding


bauglir:
I initially encountered this issue in my testing framework based on HTTP.jl when testing `DELETE` requests where the server returns a `204` response. These specific test cases would pass properly, but the subsequent test would fail.

From further investigation I believe I have narrowed it down to being unrelated to `DELETE` requests and the issue having to do with the way `204` responses are send over the pipe when using chunked transfer encoding. It could be this problem would arise with any response without a body, I haven't been able to verify that at this point in time.

When a request receives a `204` response it appears to get handled properly. However, an `excess` is left on the connection which is parsed and and causes a crash on the subsequent request on the same connection. This is reported as being due to an invalid status line being encountered. Instead the sequence `0\r\n\r\n` is parsed from the excess, which I believe indicates this is related to chunked transfer encoding.

The issue is easily reproduced by starting a server similar to the standard `serve` implementation including `handle_request`. The following code reproduces the issue in its most dressed down form I've been able to get it
```julia
using HTTP

@async HTTP.listen() do http::HTTP.Stream
  request::HTTP.Messages.Request = http.message
  request.response = HTTP.Messages.Response(204)

  HTTP.startwrite(http)
  write(http, request.response.body)
end

sleep(1)

response = HTTP.request("GET", "http://localhost:8081")
println(response)
response = HTTP.request("GET", "http://localhost:8081")
println(response)
```

This results in
```bash
$ julia --color=yes ./204_errors.jl                                                                                                                                                                                                  130 ↵
I- Listening on: Base.InetAddr{IPv4}(ip"127.0.0.1", 0x1f91)
I- Accept:  🔗    0↑     0↓    0s 127.0.0.1:8081:8081 ≣16
HTTP.Messages.Response:
"""
HTTP/1.1 204 No Content
Transfer-Encoding: chunked

"""
ERROR: LoadError: W- HTTP.Parsers.ParseError(:INVALID_STATUS_LINE, "0\r")write: broken pipe (EPIPE)
while loading /.../204_errors.jl, in expression starting on line 15
```

Forcing the requests to not reuse existing connections works fine, as expected.

[...]

PR #265, opened 2018-07-02
Various fixes for #261 as suggested by @samoconnor

quinnj:
cc: @samoconnor, @bauglir


quinnj:
@samoconnor, check out the test failure [here](https://travis-ci.org/JuliaWeb/HTTP.jl/jobs/399204489); something is causing a test failure in the loopback.jl code and causing the rest of the tests to hang.

PR #266, opened 2018-07-02
Add ability to set a default User-Agent string to be used w/ each req…

### LINKED ISSUE #161: Default User-Agent ?
The Github API requires that the User-Agent be set.
https://developer.github.com/v3/#user-agent-required

It was working for me in 0.5.7 but it is not in master.
Ergo, I suspect something changed.

quinnj:
…uest. Fixes #161 

cc: @oxinabox, @samoconnor

### DIFF
--- a/test/server.jl
+++ b/test/server.jl
@@ -61,7 +61,7 @@ tsk = @async HTTP.Servers.serve(server, Sockets.localhost, port)
 sleep(1.0)
 
 
-r = testget("http://127.0.0.1:$port/")
+r = testget("http://127.0.0.1:$port")
 @test HTTP.compat_occursin(r"HTTP/1.1 200 OK", r)
 
 rv = []
--- a/test/server.jl
+++ b/test/server.jl
@@ -37,7 +37,7 @@ end
 
 @testset "HTTP.Servers.serve" begin
 
-port = rand(8000:8999)
+port = 8086 # rand(8000:8999)
 
 # test kill switch
 server = HTTP.Servers.Server()
--- a/test/server.jl
+++ b/test/server.jl
@@ -58,7 +58,7 @@ end, stdout)
 
 server.options.ratelimit=0
 tsk = @async HTTP.Servers.serve(server, Sockets.localhost, port)
-sleep(1.0)
+sleep(5.0)
 
 
 r = testget("http://127.0.0.1:$port")
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -176,6 +176,7 @@ HTTP.Messages.setheader
 HTTP.Messages.defaultheader
 HTTP.Messages.appendheader
 HTTP.Messages.readheaders
+HTTP.MessageRequest.setuseragent!
 HTTP.Messages.readchunksize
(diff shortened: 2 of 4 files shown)


test/server.jl
```diff
@@ -37,7 +37,7 @@ end
 
 @testset "HTTP.Servers.serve" begin
 
-port = rand(8000:8999)
+port = 8086 # rand(8000:8999)
```
oxinabox: Delete commented bit?

samoconnor:
@oxinabox, I'd say that being a good net citizen means:
 - Not defining User-Agent in software that is not a user agent, i.e. A library that uses HTTP.jl to implement webdav is not a user agent; and a library that uses the webdav library to implement some acme.com file sharing API is not a user agent either. However, if I build "Sams Super Stats App" in Julia and I have an option to save stats .CSV files to acem.com, then I should set the User-Agent to "Sams Super Stats...".
 - Not setting User-Agent to something generic that ends up being duplicated across many user agents. If someone adds a User-Agent rule to a proxy to deal with a single badly behave Julia app, we don't want every other Julia app to experience weird behaviour just because they all say "User-Agent: Juila HTTP". User-Agent is only useful if it allows people to implement fine-grained work-arounds for bugs in particular apps.
- If you are going to set User-Agent, set it to something well defined and so that consumers can reliably interpret the content. e.g. Mozilla has something like `User-Agent: Mozilla/<version> (<system-information>) <platform> (<platform-details>) <extensions>`

samoconnor:
Re curl:
curl is a user agent (it's main use is as a command line tool) so it makes sense for it to have a User-Agent header. When curl is used as part of some larger application the `-A` option is used so set User-Agent to something more specific.

oxinabox:
@samoconnor  sounds like you've got half the doc I was suggesting written there :-)

oxinabox:
> If you are going to set User-Agent, set it to something well defined and so that consumers can reliably interpret the content. e.g. Mozilla has something like User-Agent: Mozilla/<version> (<system-information>) <platform> (<platform-details>) <extensions>

It was on that notion,
that I though it might be good for all calls to `setuseragent!` to append.
Which would basically ensure that like Mozilla has every extension listed,
this would mean every loaded package that uses `HTTP` and `setuseragent!`.
But that is not standard practice, so it is prob not a good idea to be the first.

oxinabox:
This just sounds plain more convenient  than declaring a wrapper function(/s) to use everywhere in your program.
And declaring such a wrapper function wouldn't propagate to calls that are made from libraries you've used, which is probably bad.

Lets say someone is made a package called GoogleDrive.jl built on using Swagger.jl which is inturn (in updated to be) built on HTTP.jl
All the `get` requests are in  Swagger.jl
But they should be use the user-agent "GoogleDrive.jl v0.1.1".

So rather than wrapping `HTTP.get` we have to make a bunch of Swagger.jl's functions take a USER-AGENT arguement, and then we have to wrap them.

Then say someone uses GoogleDrive.jl in a package, now we have to wrap ...
etc.

This seems intensely inelegant.

PR #273, opened 2018-07-06
Replace HTTP.download

oxinabox:
rather than deleting it as in #271  
this replace it with a new and better version.

It is very verbose, but I think the 0.7 way of handling that is to tell your logger to be quieter.
Also to tell your logger to display a progress bar instead, that should be possible with the new system.

Locally this was working when I had it as a separate package but it is screwing up now that I've merge it in and I'm at a lose as to why.

```
ERROR: DNSError: , unknown node or service (EAI_NONAME)
```

### DIFF
--- a/src/download.jl
+++ b/src/download.jl
@@ -2,15 +2,15 @@ using .Pairs
 
 """
     safer_joinpath(basepart, parts...)
-A variation on `joinpath`, that is more resistant to directory traveral attack
+A variation on `joinpath`, that is more resistant to directory traversal attacks.
 The parts to be joined (excluding the `basepart`),
 are not allowed to contain `..`, or begin with a `/`.
 If they do then this throws an `DomainError`.
 """
 function safer_joinpath(basepart, parts...)
-    explain =  "Possible Directory Traversal Attack detected."
+    explain =  "Possible directory traversal attack detected."
     for part in parts
-        occursin("..", part) && throw(DomainError(part, "contains illegal string \"..\". $explain"))
+        occursin("..", part) && throw(DomainError(part, "contains \"..\". $explain"))
         startswith(part, '/') && throw(DomainError(part, "begins with \"/\". $explain"))
     end
     joinpath(basepart, parts...)
--- a/src/download.jl
+++ b/src/download.jl
@@ -72,6 +72,7 @@ from the rules of the HTTP.
 
 """
 function download(url::AbstractString, local_path=nothing; update_period=0.5)
+    @debug 1 "downloading $url"
     local file
     HTTP.open("GET", url) do stream
         resp = startread(stream)
--- a/src/download.jl
+++ b/src/download.jl
@@ -1,0 +1,114 @@
+using .Pairs
+
+function try_get_filename_from_headers(headers)
+    content_disp = getkv(headers, "Content-Disposition")
+    if content_disp != nothing
+        # extract out of Content-Disposition line
+        # rough version of what is needed in https://github.com/JuliaWeb/HTTP.jl/issues/179
+        filename_part = match(r"filename\s*=\s*(.*)", content_disp)
+        if filename_part != nothing
+            filename = filename_part[1]
+            quoted_filename = match(r"\"(.*)\"", filename)
+            if quoted_filename != nothing
+                # It was in quotes, so it will be double escaped
+                filename = unescape_string(quoted_filename[1])
+            end
+            return filename
+        end
+    end
+    return nothing
+end
+
+function try_get_filename_from_remote_path(target)
+    target == "" && return nothing
+    filename = basename(target)
+    if filename == ""
+        try_get_filename_from_remote_path(dirname(target))
+    else
+        filename
+    end
+end
+
+
+determine_file(::Nothing, resp) = determine_file(tempdir(), resp)
+# ^ We want to the filename if possible because extension is useful for FileIO.jl
+
+function determine_file(path, resp)
+    # get the name
+    name = if isdir(path)
+        # got to to workout what file to put there
+        filename = something(
+                        try_get_filename_from_headers(resp.headers),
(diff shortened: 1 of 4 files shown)


oxinabox:
So this is failing on 0.6, because of usign the new `@info`.
with lots of kwargs.
I think using `@info` that way is good, because it will play nice with a Logger that implements progress bars.

oxinabox:
Demo of use (with [OhMyLog.jl](https://github.com/oxinabox/OhMyLog.jl))
[HTTP Demo](https://asciinema.org/a/PDw1Ekwsf1U6mbKkP5lglQi03?t=40)

samoconnor:
Regarding 0.6 compatibility. I would just hold of merging this until we update `REQUIRE` to say `julia 0..7`. 

See comment here: https://github.com/JuliaWeb/HTTP.jl/pull/264#discussion_r199653839
> "I would say that at this point, with 0.7 beta out, feature addition PRs for HTTP.jl could(should?) be 0.7-only. i.e. we should put julia 0.7 in REQUIRE soon anyway."

oxinabox:
Fair, I am curious what it would take for 0.6 compat so I just made a commit (also I want to retrigger 0.7 CI which is being flaky).

oxinabox:
While I remember: this should take a headers argument.
because that is required to work with OATH.

oxinabox:
Ahhhh,
this has a Directory Traversal vulnerability I think.
Or at least needs to be checked carefully for one. 

Consider that it lets the file set its own filepath.
And that 
```
julia> joinpath("alocaldir", "/etc/passwd")
"/etc/passwd"
```

So I think we need to blacklist any filepath that starts with a `/`, or contains any `..`
and probably some other things.
Does anyone know of a set of blacklist rules for this?

samoconnor:
As a double-check, how about adding an assertion that the resulting path is actually under the download directory.
```julia
path = joinpath(download_dir, file_path)
@assert startswith(normpath(path), normpath(download_dir))
```

oxinabox:
Ah, I didn't realize that `normpath` existed.
I was thinking about doing this using `realpath`,
but that has a variety of problems since symlinks make it not nesc true.

On the other hand, `normpath` literally just looked for `..`  and removes them.
Given we have banned `..` outright (because you never want this in a downloaded file)
idk, it would be an assertion, and it isn't like it would take long per check.

oxinabox:
> I would expect something like HTTP.download to just download a specified file to some specified location.

It does. But the question is what does a specific location mean?
`HTTP.download` should work a lot like clicking download in a browser.
Only instead of a dialog appearing, it uses the second argument to give the specific location

 - if a specific location is path to a file* then yes, send it there. Easy Done.
       - (* or  technicall to a nonexistant location, which this PR currently assumes to be a file)
 - if a specific location is a directory, then we need to workout the filename to put in that directory.
        - We need to get the filename right, because 
               - it is part of the file being downloaded -- intent of the file creator
               - FileIO.jl (etc) need the file extension to know how to open it.
       - If it is coming from a very simple HTTP server, then the file name is generally just the last part of the URL
       - But HTTP lets it be specified in the header.
                    - A surfiently evil header on a compromised server might say that the filename of this is say `../../.shh/accept/evilkey.pub`, and then break into your system

EricForgy:
Why allow the case where the filename is located in the header? It would be reasonable, I think, to reject the download if the filename is not in the url :thinking_face:

oxinabox:
There are a lot of files I want to download (millions)
That do not have it in the URL.
Anything on any modern filehosting service like Dropbox or google drive, anything on data.gov etc etc
I want to download them and I want there original filenames

KristofferC:
Should this also accept headers?

quinnj:
@oxinabox, what's the status here? ready to review? still things you want to cleanup? just let me know. I'm going to target one final 0.6 release in the next few days and then we can rip out all the 0.6 compat stuff.

oxinabox:
I want to add everything you suggested in 
https://github.com/JuliaWeb/HTTP.jl/pull/273#issuecomment-403803409
And I was kinda not rushing on it since can wait for 0.7

It won't take much to finish that off

oxinabox:
Right, added headers and kwargs.
I didn't add body, because that doesn't seem reasonable.
I could be wrong, maybe someone uses it for auth.
But a GET request with a body; particularly for a download doesn't make sense to me.

I've not added human units, as I am of many minds and bikesheds over that.
So it seems better to add that later in another PR.

So this is ready for review, I hope.

The compat stuff can be stripped out, though I am not sure the best way to do that right now.
Can I just delete the compat.jl file?

oxinabox:
Ok, this is all rebased for 0.7/1.0
and working

oxinabox:
Assuming windows passes, I am done with this now.

oxinabox:
Bump

oxinabox:
New video, with OhMyLog.

[asciicast](https://asciinema.org/a/00hoDMBZqtNxwHxKThDg5j9cC)

src/download.jl
```diff
@@ -0,0 +1,142 @@
...
+When saving into a directory, the filename is determined (where possible),
+from the rules of the HTTP.
+
+ - `update_period` controls how often (in seconds) to report the progress.
+    - set to `Inf` to disable reporting
+ - `headers` specifies headers to be used for the HTTP GET request
+ - any additional keyword args (`kw...`) are passed on to the HTTP request.
+"""
+function download(url::AbstractString, local_path=nothing, headers=Header[]; update_period=1, kw...)
+    format_progress(x) = "$(round(100x, digits=2))%"
```
oxinabox: Possibly this should not be formatted.
idk what Juno is doing with `progress`.
OhMyLog doesn't mind if it is given a string or a float.

oxinabox:
Can we have a release tagged?
I want to Switch over to using this in my packages

PR #277, opened 2018-08-03
Fix julia 0.7 deprecations

davidanthoff:


quinnj:
I think the osx nightly error is just because the build is out-dated.

Issue #250, opened 2018-06-06
Remove client.jl


quinnj:
We've had the deprecations in for several releases now, so I think it's time to remove the client.jl file. We need to refactor the `download` function, but we could also just deprecate it as well. @samoconnor, do you want to tackle this? Or I could take a stab at it sometime soon.

Or if anyone else is interested, I think it'd be a pretty easy PR.

samoconnor:
I believe the download function isn't tested or documented.
I'm in favour of just removing it.
Maybe put an example like this at the top of the README:
```julia
HTTP.open("GET", url) do io
    write("/tmp/foo", io)
end
```

samoconnor:
I'm assuming that at some point soon after 0.7 is released, we'll make a HTTP version that drops 0.6.3 from the REQURE file. Maybe that's a good time to remove the client.jl (and do a general pass through the code other deprecated stuff that we might want to remove).
My thinking is that this will avoid inconvenience for people who may have old working 0.6 code and no immediate plans to move to 0.7. Whereas people who are taking the plunge to upgrade their code to 0.7 will be expecting to deal with a deprecations/changes.

samoconnor:
Re dropping 0.6.3 from REQUIRE: We can always back-port anything critical to a 0.6 branch if there is sufficient demand; and dropping 0.6.3 would allow cleanup of various `if VERSION...` cruft.

quinnj:
Yeah, that's a good idea to do this w/ the dropping of 0.6.

Nosferican:
Is this still relevant or should it be closed?

PR #278, opened 2018-08-11
Update for 0.7/1.0

oxinabox:


oxinabox:
I just checked over the `if VERSION ...` blocks.
to my surprise i actually got them all correct AFAICT 

All the conditions and stuff should be solved.
This does need femtocleaning to get ride of a deprecation I introduced,

I am note sure what is with the failure.

samoconnor:
Probably should delete `client.jl` at the same time as dropping 0.6 support: https://github.com/JuliaWeb/HTTP.jl/issues/250#issuecomment-395248642

samoconnor:
This should also cover #283

oxinabox:
#283 seems like a separate problem.
I just want to get the 0.7 transition done, and compat cruft removed.
Isn't that a bug present in 0.6?

oxinabox:
Can someone with a bit more experience at the workings of this package take a look at the remaining errors. (@quinnj, @samoconnor).
I don't think they are directly related to any deprecations, but maybe are related to something I changed.

Also can Femtocleaner be waved over this?
It has a deprecation here that I think it can fix.

quinnj:
Closed in favor of https://github.com/JuliaWeb/HTTP.jl/pull/290 (since I can't push to @oxinabox 's fork)

oxinabox:
Sorry, idk why you can't push to my fork.
Thanks

PR #282, opened 2018-08-19
add HTTP.openraw method to get raw socket

erikedin:
This is the PR that was discussed in the closed PR #279 . This adds a new client method that returns
an unmanaged raw socket. An expected use case is upgrading from HTTP to other protocols, such as
WebSockets.

The method returns a tuple of the socket, the response headers, and any excess bytes read of the body
while parsing the HTTP headers. The test in client.jl tests a WebSocket upgrade, and sending and
receiving a single WebSocket frame.


samoconnor:
@quinnj this looks good to me. Any comments?

erikedin:
FYI, this fails for SSL connections. The fix is simple, and incoming soon. Just have to get my Julia installation under control.

quinnj:
Cool, just ping us again when you have a fix @erikedin (or any questions)

erikedin:
The immediate issue was that I had made the types in `HTTP.openraw` specific to `TCPSocket`, so when SSL was used, then it failed because the socket was a `MbedTLS.SSLContext`.
That is fixed, but SSL still doesn't seem to work. It appears that the socket is EOF as soon as `HTTP.openraw` returns. I don't know if anything else needs to be done for SSL.

Note that the test is modified to make it easier to run it for both non-SSL and SSL, but that it still doesn't run `openraw` for SSL, since it's failing.

I'll resolve the conflicts and keep looking.

erikedin:
It appears the SSL issue has something to do with the `Base.wait_close` call in `HTTP.openraw`, in the bit where I do

```julia
        while(isopen(socket))
            Base.wait_close(socket)
        end
```
If I replace the call to `Base.wait_close` with a sleep, then the SSL test works just fine. Do you have any ideas about why this would happen? Could we replace it with something specific to `MbedTLS` maybe?

erikedin:
Good news is that I could fix it by handling `Base.wait_close` separately for `MbedTLS.SSLContext`. The tests pass now, so it seems fine.
It does depend on `MbedTLS.SSLContext` having the field `bio`, which is the underlying socket. I'm not happy about having to rely on internal details like that, but it's the best I could think of.

samoconnor:
Related: https://github.com/JuliaWeb/HTTP.jl/pull/305
HTTP.jl/src/WebSockets.jl now uses ConnectionPool.Transaction (not the raw socket or tls). This approach means that there is no need for special handling of `excess`.

erikedin:
I tested the PR as is with the latest changes, and it still works. I also tried modifying the `openraw` code to use a `Transaction` instead, but the tests failed, apparently because it failed to read any bytes. The `openraw` code using the Transactions looks like

```julia
function openraw(method::String, url, headers=Header[]; kw...)::Tuple{IO, Response}
    waitforclose(t::ConnectionPool.Transaction{TCPSocket}) = Base.wait_close(t.c.io)
    waitforclose(t::ConnectionPool.Transaction{MbedTLS.SSLContext}) = Base.wait_close(t.c.io.bio)
    socketready = Channel{Tuple{IO, Response}}(0)
    @async HTTP.open(method, url, headers; kw...) do http
        HTTP.startread(http)
        socket = http.stream
        put!(socketready, (socket, http.message))
        while(isopen(socket))
            waitforclose(socket)
        end
    end
    take!(socketready)
end
```

I would personally still prefer the raw TCPSocket, though.

samoconnor:
Here is a slightly tweaked version that works for me:

```julia
      function openraw(method::String, url, headers=Header[]; kw...)::Tuple{IO, Response}
           socketready = Channel{Tuple{IO, Response}}(0)
           @async HTTP.open(method, url, headers; kw...) do http
               HTTP.startread(http)
               socket = http.stream
               put!(socketready, (socket, http.message))
               while(isopen(socket))
                   Base.wait_close(socket)
               end
           end
           take!(socketready)
       end
```

Here is the test script:

```julia
@testset "openraw client method" begin
     for sch in ("ws", "wss")
         println("openraw client method: $sch")

         @testset "can send and receive a WebSocket frame" begin
             # WebSockets require valid headers.
             headers = Dict(
                 "Upgrade" => "websocket",
                 "Connection" => "Upgrade",
                 "Sec-WebSocket-Key" => "dGhlIHNhbXBsZSBub25jZQ==",
                 "Sec-WebSocket-Version" => "13")

             socket, response = openraw("GET", "$sch://echo.websocket.org", headers)

             @test response.status == 101

             # This is an example text frame from RFC 6455, section 5.7. It sends the text "Hello" to the
             # echo server, and so we expect "Hello" back, in an unmasked frame.
             frame = UInt8[0x81, 0x85, 0x37, 0xfa, 0x21, 0x3d, 0x7f, 0x9f, 0x4d, 0x51, 0x58]

             write(socket, frame)

             # The frame we expect back looks like `expectedframe`.
             expectedframe = UInt8[0x81, 0x05, 0x48, 0x65, 0x6c, 0x6c, 0x6f]

             # Note the spec for read says:
             #     read(s::IO, nb=typemax(Int))
             # Read at most nb bytes from s, returning a Vector{UInt8} of the bytes read.
             # ... so read will return less than 7 bytes unless we wait first:
             eof(socket)
             actualframe = read(socket, 7)
             @test expectedframe == actualframe
             
             write(socket, frame)

             # On the other hand `read!` always waits to fill the whole vector.
             actualframe = Base.StringVector(7)
             read!(socket, actualframe)
             @test expectedframe == actualframe
             close(socket)
         end
     end
 end
```

This change will make `Base.wait_open` work for both TCP and TLS sockets:
```patch
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -327,6 +327,7 @@ function monitor_idle_connection(c::Connection{SSLContext})
     # https://github.com/JuliaWeb/MbedTLS.jl/pull/145
 end

+Base.wait_close(t::Transaction) = Base.wait_close(tcpsocket(t.c.io))

 function Base.close(t::Transaction)
     close(t.c)
```

This change makes it all work even with `HTTP.jl/DEBUG_LEVEL = 4`

```patch
--- a/src/IODebug.jl
+++ b/src/IODebug.jl
@@ -27,10 +27,16 @@ else

 end

+Base.wait_close(iod::IODebug) = Base.wait_close(iod.io)
+
 Base.write(iod::IODebug, a...) =
     (logwrite(iod, :write, join(a));
      write(iod.io, a...))

+Base.write(iod::IODebug, a::Array) =
+    (logwrite(iod, :write, join(a));
+     write(iod.io, a))
+
 Base.write(iod::IODebug, x::String) =
     (logwrite(iod, :write, x);
      write(iod.io, x))
```

samoconnor:
The IODebug.jl and ConnectionPool.jl changes mentioned above:
https://github.com/JuliaWeb/HTTP.jl/commit/0e3b053a52b1792a81a06e0d3d25025213405eb5
https://github.com/JuliaWeb/HTTP.jl/commit/967bd971bbe18c7233bbef309e6ae3402bfcce86

erikedin:
I would need control over the raw socket, regardless. For instance, I need to conditionally set the TCP_NODELAY flag on the socket.

samoconnor:
> I would need control over the raw socket, regardless. For instance, I need to conditionally set the TCP_NODELAY flag on the socket.

How about this:
 - Return `socket = http.stream` from `openraw` (a `socket isa Transaction`).
 - Add a method to `tcpsocket` that return a `TCPSocket` for `Transaction` (regardless of TLS).

That way code using `openraw` could use `tcpsocket(socket)` to set TCP flags.

e.g.
```julia
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -123,6 +123,7 @@ function client_transaction(c)
 end

 getrawstream(t::Transaction) = t.c.io
+IOExtras.tcpsocket(t::Transaction) = tcpsocket(getrawstream(t))

 inactiveseconds(t::Transaction) = inactiveseconds(t.c)
```

erikedin:
Yes, that would work for me, as far as I can see. I'll make the changes when I can.

erikedin:
I updated according to your suggestions. This should work for me.

I had an issue with the second part to the test script that you added, so I skipped it. The `read!` method caused an exception `ErrorException("HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext} does not support byte I/O")`.

samoconnor:
Can you open an issue that describes the failing test code?
I'd like to look into that but I don't have time just now.

erikedin:
For reference, I opened issue #329 describing the exception I had in the test code.

Issue #245, opened 2018-05-30
EOFError() during request


Alexander-Barth:
I getting the error below when I want to upload a file to a webdav server.
I am getting this error always after a some time (~ minute) of inactivity, but if I repeat the upload right-away it always succeeds. 

The error occurs when I read the stream. If I replace the stream `f` by `read(f)`), then the request always succeeds (but I have to deal with large files, this is not a good option for me).

The equivalent `curl` request works always:
```
curl -T mytest_curl --user 'username:password' 'https://b2drop.eudat.eu/remote.php/webdav/mytest_curl'
```

I use these versions:

Julia 0.6.2 
HTTP.jl 0.6.9 
MbedTLS.jl 0.5.8

I am getting also the same error with current master (794b3201edbe94f4a761eb449f7c192a9aabf81c).

Unfortunately, I cannot share my credentials for you to reproduce the error :-). 
The webdav server here is an NextCloud instance.

Here is the test code with the error message:

```
userinfo = "username:password"
f = open("mytest","r"); r = HTTP.request("PUT", "https://b2drop.eudat.eu/remote.php/webdav/mytest_julia5", [("Authorization", "Basic $(base64encode(userinfo))")], f; timeout = 120, verbose = 3);
DEBUG: 2018-05-30T16:32:18.828 1967 ➡️  "PUT /remote.php/webdav/mytest_julia5 HTTP/1.1\r\n"
DEBUG: 2018-05-30T16:32:18.828 1967 ➡️  "Authorization: Basic XXXXXXXX=\r\n"
DEBUG: 2018-05-30T16:32:18.828 1967 ➡️  "Host: b2drop.eudat.eu\r\n"
DEBUG: 2018-05-30T16:32:18.829 1967 ➡️  "Transfer-Encoding: chunked\r\n"
DEBUG: 2018-05-30T16:32:18.829 1967 ➡️  "\r\n"
DEBUG: 2018-05-30T16:32:18.829 4f72 ➡️  "a\r\n"ERROR: 
IOError(EOFError() during request(https://b2drop.eudat.eu/remote.php/webdav/mytest_julia5))

Stacktrace:
 [1] readuntil(::HTTP.DebugRequest.IODebug{HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}}, ::HTTP.Parsers.#find_end_of_header) at /home/abarth/.julia/v0.6/HTTP/src/IOExtras.jl:171
[...]

PR #285, opened 2018-08-22
capture async error in writebody(http, request, body)

samoconnor:
See https://github.com/JuliaWeb/HTTP.jl/issues/245#issuecomment-414847061


src/StreamRequest.jl
```diff
@@ -48,12 +48,20 @@ function request(::Type{StreamLayer}, io::IO, request::Request, body;
...
         @sync begin
 
             if iofunction == nothing
-                @async writebody(http, request, body)
+                @async try
+                    writebody(http, request, body)
+                catch e
+                    write_error = e
+                    @warn("Error in @async writebody task",
+                          exception=(write_error, catch_backtrace()))
```
quinnj: I think this should be `stacktrace(catch_backtrace())`

samoconnor: I did it this way because the `@warn` macro seems to do special processing of `Tuple{Exception,BackTrace}`. i.e. it prints out a nice stack-trace.

src/StreamRequest.jl
```diff
@@ -68,8 +76,8 @@ function request(::Type{StreamLayer}, io::IO, request::Request, body;
         end
 
     catch e
-        if e isa CompositeException
-           e = first(e.exceptions).ex
+        if write_error != nothing
+            throw(write_error)
```
quinnj: Is this changing behavior from before? Didn't we pass on to the next `if` block before and check is it's a `isioerror`. Why change that now?

samoconnor: hmm, looking at this again, I think there is more to it...

Assume that `writebody` is sending some huge message body, and the server sends an error with `Connection: "close` (https://tools.ietf.org/html/rfc7230#section-6.5).
`readbody` will return, and `isaborted()` will be true on line 72.
So, close gets called and the `writebody` task stop with an IO error, which is stored in `write_error`.
However, the `catch` block at 79 will never be reached, because there is no exception to catch  in the main task, so `write_error` will not be used at all. This makes sense in this case because the main task has decided to kill the `writebody` task and doesn't care about any subsequent exceptions that it might throw.

I think we can reduce the `catch` block to just:
```julia
    catch e
        if write_error != nothing
            throw(write_error)
        else
            rethrow(e)
        end
    end
```

This makes the intent of the code a bit easier to follow, i.e: If there is an error while reading the response, check if the root-cause was an error sending the request, otherwise just rethrow the reading error.

quinnj:
Rebased.

PR #290, opened 2018-09-07
Remove 0.6 compat

quinnj:
Replaces #278


oxinabox:
Project.toml need be removed before tagging or it breaks juliaCIBot (if you have test time deps)

Just renaming to to `_Project.toml` for now works

quinnj:
Huh.....all green except 0.6 😛

PR #292, opened 2018-09-07
Better connection management

quinnj:


src/ConnectionRequest.jl
```diff
@@ -49,7 +50,7 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
         close(io)
         rethrow(isioerror(e) ? IOError(e, "during request($url)") : e)
     finally
-        if proxy !== nothing && target_url.scheme == "https"
+        if (proxy !== nothing && target_url.scheme == "https") || reuse_limit == 0
```
quinnj: @samoconnor, in this case, if the `reuse_limit=0`, then the connection was still hanging around in the pool until the next request, at which point it realized it couldn't be re-used and closed it. This just ensures that if a connection shouldn't be re-used, it is closed immediately after.

src/Servers.jl
```diff
@@ -394,9 +399,6 @@ function handle_connection(f::Function, c::Connection;
             io = Transaction(c)
             handle_transaction(f, io; final_transaction=(count == reuse_limit),
                                       kw...)
-            if count == reuse_limit
```
quinnj: If `reuse_limit=0`, then the connection was being closed before the handler could respond. Moved the close further down below.

src/ConnectionRequest.jl
samoconnor: Could you please try to maintain 80-col wrapping in files that currently adhere to that.

src/Servers.jl
```diff
@@ -315,12 +316,24 @@ function listen(f::Function,
...
 
+    if host isa Sockets.InetAddr # build debugging info
+        hostname = string(host.host)
+        hostport = string(host.port)
+    else
+        hostname = string(host)
+        hostport = ""
+    end
+    listenloop(f, tcpserver, sslconfig, hostname, hostport, pipeline_limit, require_ssl_verification, tcpisvalid, connectioncounter; kw...)
+end
```
samoconnor: Why is `listenloop` split into a seperate function?
It seems a bit cumbersome to have to pass all those positional parameters across in the right order.
Will `listenloop` be called from somewhere else too?

quinnj: The split was to improve type inference; once the higher-level `listen` function figures out the exact types of `tcpserver`, `sslconfig`, and `tcpisvalid`, it can pass them all to `listenloop` which can then compile solely for the types provided; previously, type inference struggled quite a bit with all the uncertainty.

samoconnor: Fair enough. Was this causing a noticeable performance problem?

quinnj: Yes, it was a 5-10% performance hit from my benchmarks.

test/runtests.jl
```diff
@@ -7,19 +7,19 @@ using HTTP
...
+    println("running uri.jl tests..."); include("uri.jl");
+    println("running url.jl tests..."); include("url.jl");
+    println("running cookies.jl tests..."); include("cookies.jl");
+    println("running parser.jl tests..."); include("parser.jl");
 
-    include("loopback.jl");
-    include("WebSockets.jl");
-    include("messages.jl");
+    println("running loopback.jl tests..."); include("loopback.jl");
+    # println("running WebSockets.jl tests..."); include("WebSockets.jl");
```
samoconnor: ?

samoconnor: passing in #303 CI
https://travis-ci.org/JuliaWeb/HTTP.jl/jobs/429938842#L1158-L1168

PR #296, opened 2018-09-11
Wrap UVError/IOError in getconnection with HTTP.IOError

### LINKED ISSUE #295: No error handling for broken pipe in getconnection
Julia 0.6.4 
HTTP.jl 0.6.14 
MbedTLS.jl 0.5.12 -- Please specify versions

https://github.com/JuliaWeb/HTTP.jl/blob/cd9f60bcce2cd34c20bd87f10f5e30ff3db22744/src/ConnectionRequest.jl#L38
Sometimes this call to `getconnection` throws a `UVError`/`IOError` "connect: broken pipe (EPIPE)" (e.g., https://github.com/invenia/CloudWatchLogs.jl/issues/4). 

As far as I know this will be the same in the latest Julia+HTTP version combo.

Wrapping this in `HTTP.IOExtras.IOError` would allow the [retry code in AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/http.jl#L54) to retry in this case. Is that something we can add+backport?

iamed2:
Fixes #295 

Is there a nice way to test this (e.g., generate a broken pipe)?

### DIFF
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -36,7 +36,12 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
     end
 
     IOType = ConnectionPool.Transaction{sockettype(url, socket_type)}
-    io = getconnection(IOType, url.host, url.port; reuse_limit=reuse_limit, kw...)
+    local io
+    try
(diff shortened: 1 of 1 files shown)


iamed2:
Are we going to wait for Sam for this?

Issue #313, opened 2018-09-21
Multipart Request Issues


dtjohnson:
Julia 1.0.0 
HTTP.jl 0.7.0 

I'm having two issues submitting multipart/form-data. I'm submitting a mixture of JSON data and files to an API that expects the JSON part to come first in order to understand the requests before handling the files. I'm able to form a multipart/form-data body using `HTTP.Form` with a `Dict`, but I've hit two issues.

First, I put the JSON part into the `Dict` as a serialized string, but when it gets written to the API it is URL-encoded. I can see this is intentional on [this line](https://github.com/JuliaWeb/HTTP.jl/blob/master/src/multipart.jl#L72), but I'm not sure why. I don't see this as a requirement in [RFC 7578](https://tools.ietf.org/html/rfc7578) and it's not the behavior I've seen in browsers with [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) nor in Node.js.

The second issue is that the order of the parts is not preserved because `Dict` does not preserve them. The API rejects the request as it's expecting the JSON part first. Section 5.2 of the RFC does state that order should be preserved in form data so it's reasonable for an API to require this. And unfortunately, since the constructor is typed with a concrete `Dict` I can't pass in something like an [`OrderedDict`](http://juliacollections.github.io/DataStructures.jl/latest/ordered_containers.html#OrderedDicts-and-OrderedSets-1) that preserves order. Could we switch this to a more generic type? Or perhaps change to an array of pairs?

PR #319, opened 2018-09-30
Added array of pair constructor for multipart Form. Removed escaping form values.

dtjohnson:
Addresses #313. Keeps in place the existing `Dict` constructor for `Form` but adds a new one with an array of pairs to preserve multipart order. Also, removes escaping form values.

You can now create a form with either syntax:
```julia
HTTP.Form(["foo" => "FOO", "bar" => "BAR"])
# OR
HTTP.Form(Dict("foo" => "FOO", "bar" => "BAR"))
```

### DIFF
--- a/src/multipart.jl
+++ b/src/multipart.jl
@@ -53,7 +53,7 @@ function Base.read(f::Form, n::Integer)
     return result
 end
 
-function Form(d::Dict)
+function Form(d::Vector{Pair{String,T}}) where T<:Any
     boundary = string(rand(UInt128), base=16)
     data = IO[]
     io = IOBuffer()
@@ -69,13 +69,17 @@ function Form(d::Dict)
             io = IOBuffer()
         else
             write(io, "\r\n\r\n")
-            write(io, escapeuri(v))
+            write(io, v)
         end
         i == len && write(io, "\r\n--" * boundary * "--" * "\r\n")
     end
     seekstart(io)
     push!(data, io)
     return Form(data, 1, boundary)
+end
+
+function Form(d::Dict{String,T}) where T<:Any
+    return Form(collect(d))
 end
 
 function writemultipartheader(io::IOBuffer, i::IOStream)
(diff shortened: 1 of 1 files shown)


src/multipart.jl
```diff
@@ -53,7 +53,7 @@ function Base.read(f::Form, n::Integer)
     return result
 end
 
-function Form(d::Dict)
+function Form(d::Vector{Pair{String,T}}) where T<:Any
```
samoconnor: Maybe just allow any type of `d` ?

That way you can pass in anything that makes sense with : `for (i, (k, v)) in enumerate(d)`
e.g. an iterator that produces Pairs, or Tuples, or a Dict ...

dtjohnson:
Done! Would be nice to restrict to just inputs matching that enumeration pattern, but not sure that's possible with Julia...

src/multipart.jl
```diff
@@ -69,7 +69,7 @@ function Form(d::Dict)
             io = IOBuffer()
         else
             write(io, "\r\n\r\n")
-            write(io, escapeuri(v))
```
samoconnor: The `escape` comes from this commit https://github.com/JuliaWeb/HTTP.jl/commit/014d9f56363429568e12545f8d74fca93cc664cc#diff-1f61f32925620a8c37482af6f9a5d108R349.
As you say (#313), it doesn't seem to be called for by the RFC.
@quinnj do you recall the intention here?

quinnj: No, I don't recall; very well could be wrong.

samoconnor:
> restrict to just inputs matching that enumeration pattern
You could add `@require eltype(d) <: Pair` at the top of `Form(d)` to check that the argument produces pairs.

samoconnor:
added `@require`  here https://github.com/JuliaWeb/HTTP.jl/commit/01a2cff816e3cd6095990b80167f216cb26e906c

dtjohnson:
Thanks for merging!

Issue #320, opened 2018-10-02
WebSockets.is_upgrade gives the wrong result for (at least) Firefox clients


CarpeNecopinum:
```
using HTTP
HTTP.listen("0.0.0.0", 9002) do req
    @show HTTP.WebSockets.is_upgrade(req.message)
end
```
prints out `(HTTP.WebSockets).is_upgrade(req.message) = false` when trying to connect via WebSockets from Firefox.

The Problem in the `is_upgrade` function is that the expression `HTTP.hasheader(r, "Connection", "upgrade")` requires `upgrade` as the only element in the `Connection` header-field being sent, but (at least) Firefox sends `keep-alive, Upgrade`.

Julia 1.0.0
HTTP.jl 0.7.1
MbedTLS.jl none
Firefox 62.0.2

PR #321, opened 2018-10-04
Fixed WebSockets not accepting upgrade requests from Firefox

CarpeNecopinum:
Fixes issue #320.

In order to accept upgrade requests from Firefox, which not only sends "connection: upgrade", but "connection: keep-alive, upgrade", a new method `headercontains` was added.

`headercontains` works similar to `hasheader` but will interpret the header value as comma-separated list and test if the expected value is in that list, rather than considering it as a single string.

Looking forward to your feedback.

### DIFF
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -336,7 +336,7 @@ hasheader(m, k::AbstractString, v::AbstractString) =
 Does the header for `key` (interpreted as comma-separated list) contain `value` (both case-insensitive)?
 """
 headercontains(m, k::AbstractString, v::AbstractString) =
-    any(field_name_isequal.(strip.(split(header(m, k), ",")), lowercase(v)))
+    any(field_name_isequal.(strip.(split(header(m, k), ",")), v))
 
 """
     setheader(::Message, key => value)
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -60,7 +60,7 @@ module Messages
 export Message, Request, Response, HeaderSizeError,
        reset!, status, method, headers, uri, body,
        iserror, isredirect, ischunked, issafe, isidempotent,
-       header, hasheader, setheader, defaultheader, appendheader,
+       header, hasheader, headercontains, setheader, defaultheader, appendheader,
        mkheaders, readheaders, headerscomplete,
        readchunksize,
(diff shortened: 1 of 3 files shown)


src/Messages.jl
```diff
@@ -330,6 +330,14 @@ Does header for `key` match `value` (both case-insensitive)?
 hasheader(m, k::AbstractString, v::AbstractString) =
     field_name_isequal(header(m, k), lowercase(v))
 
+"""
+    headercontains(::Message, key, value) -> Bool
+
+Does the header for `key` (interpreted as comma-separated list) contain `value` (both case-insensitive)?
+"""
+headercontains(m, k::AbstractString, v::AbstractString) =
+    any(field_name_isequal.(strip.(split(header(m, k), ",")), lowercase(v)))
```
samoconnor: no need to `lowercase` `v` because `field_name_isequal` already does case insensitive comparison?

Need test cases for this new function.

samoconnor:
Nice one. LGTM. Needs tests. At least a unit test for the headercontains function.

src/Messages.jl
CarpeNecopinum: (then the `lowercase v` should probably go from the `hasheader` above, too)

EricForgy:
Out of curiosity, will there be a performance implication with this?

samoconnor:
@EricForgy I would guess quite tiny overhead.
Also, we could build a lazy zero-allocation `headercontains` for https://github.com/JuliaWeb/HTTP.jl/blob/so/lazyintegrate/src/LazyHTTP.jl

CarpeNecopinum:
Just measured, it takes about 2 times as long as `hasheader`, but we're still talking about single-digit microseconds here and if you're handling several million upgrade requests per second, you'll probably run into other problems first.

(Also, having a version that works with both major browsers is probably better than a version that is fast but only works on one of them.)

CarpeNecopinum:
Noticed that `hasheader` wasn't really tested either, so I squeezed in 2 `@test`s for it, too...

samoconnor:
> Is there still anything preventing this from being merged?

Just lack of prodding :)

Hugo-Trentesaux:
I modified this a little bit in #374

PR #330, opened 2018-10-16
Cosmetic changes

samoconnor:
Remove redundant `using MbedTLS` from toplevel `src/HTTP.jl`

Servers.jl:
 - Line-wrap some docstrings and code
 - More examples in docstrings.
 - Some refinement of wording and punctuation in docstrings.
 - Avoid bringing unwanted symbols in from MbedTLS (`encrypt` etc).
 - Rename `Server2` -> `Server` and fix type params.
 - Call `@async check_readtimeout` rather than having `@async while` inside that function.

test/server.jl
 - Qualify type names to allow test to run stand-alone.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -1,8 +1,9 @@
 """
-The `HTTP.Servers` module provides server-side http functionality in pure Julia.
+The `HTTP.Servers` module provides HTTP server functionality.
 
-The main entry point is `HTTP.listen(f, host, port; kw...)` which takes a `f(::HTTP.Stream)::Nothing` function argument
-a `host` and `port` and optional keyword arguments. For full details, see `?HTTP.listen`.
+The main entry point is `HTTP.listen(f, host, port; kw...)` which takes
+a `f(::HTTP.Stream)::Nothing` function argument, a `host`, a `port` and
+optional keyword arguments.  For full details, see `?HTTP.listen`.
 """
 module Servers
 
@@ -14,7 +15,9 @@ using ..Messages
 using ..Parsers
 using ..ConnectionPool
 using Sockets
-using MbedTLS
+using MbedTLS: SSLContext, SSLConfig
+import MbedTLS
+
 using Dates
 
 # rate limiting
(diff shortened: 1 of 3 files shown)


src/Servers.jl
```diff
@@ -119,15 +135,47 @@ e.g.
...
+
+To run the following HTTP chat example, open two Julia REPL windows and paste
+the example code into both of them. Then in one window run `chat_server()` and
+in the other run `chat_client()`, then type `hello` and press return.
+Whatever you type on the client will be displayed on the server and vis-versa.
+
+```
+using HTTP
+
+function chat(io::HTTP.Stream)
```
quinnj: This is a great example

src/Servers.jl
```diff
@@ -60,18 +64,18 @@ function check_rate_limit(tcp, rate_limit::Rational{Int})
 end
 
 "Convenience object for passing around server details"
-struct Server2{S, I}
-    ssl::S # Union{SSLConfig, Nothing}; Nothing if non-SSL
+struct Server{S <: Union{SSLConfig, Nothing}, I <: Base.IOServer}
```
quinnj: My only concern here was if people had code calling `HTTP.Server` and now it's changed. But I guess that'll be an issue anyway, so I guess this is fine.

samoconnor: But, this `HTTP.Servers.Server` is not exported right?

src/Servers.jl
EricForgy: Chat? Real-time dashboards? 😅

samoconnor: No, I know that WebSockets is really useful for building interactive browser integrated stuff.
It's a rhetorical question really.
WebSockets is necessary because that is what browsers have a javascript API for. But it's really a bit of a historical oddity due to the fact that the bi-directional streaming that is already built into the HTTP/1.1 spec was not initially widely (or well) implemented.
As this chat example demonstrates, it is perfectly possible to use a plain old HTTP/1.1 connection as a bi-directional socket. And in some cases, if browser interaction is not required, it will be more efficient.

PR #333, opened 2018-10-16
Use the General registry uuid.

fredrikekre:
Otherwise HTTP in a checked out state is considered another package.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,5 @@
 name = "HTTP"
-uuid = "132d396a-9d4e-11e8-2bcf-fd61033e480f"
+uuid = "cd3eb016-35fb-5094-929b-558a96fad6f3"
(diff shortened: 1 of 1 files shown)


samoconnor:
LGTM. But I don't really grok Pkg3 yet.
@quinnj unless you have a comment on this I'll assume that @fredrikekre is right that this is the correct thing to do.

fredrikekre:
To give some more context; the METADATA -> General registry converter generates the uuid based on the name, and for now ignores uuid in the repos Project.toml. It should match https://github.com/JuliaRegistries/General/blob/4113ec12bcb4a74ac8ebb212d1069c795d6d48cc/H/HTTP/Package.toml#L2

quinnj:
Thanks @fredrikekre; yeah, the alternative UUID here came from using Pkg3 very early on while developing HTTP.jl.

Issue #337, opened 2018-10-20
support `readbytes!`?


ssfrr:
Julia 1.0.1
HTTP.jl 0.7.1
MbedTLS.jl 0.6.4

I'm not sure if this is the same issue as #329 or if this is more expected behavior, but here's a simpler MWE that throws the same error on the current release:

```julia
using HTTP

const STREAM_URL = "http://doppler.media.mit.edu/impoundment.opus"

readbuf = Vector{UInt8}(undef, 4096)
HTTP.open("GET", STREAM_URL) do stream
    readbytes!(stream, readbuf)
end
```

Is there a workaround to get `readbytes!` working? I have some other code that works fine when streaming files from disk, but I'm now trying to get it working streaming audio from the network. `read!` and `read` also throw the same error.

samoconnor:
One of the reasons for not providing a specialised method for `readbytes!(::HTTP.Transaction, ...) `is that Base has two contradicting definitions of `readbytes!`, so it is difficult to know what behaviour to implement (See https://github.com/JuliaLang/julia/issues/24526)

They both say "at most nb bytes", but the second one says `all=` is `true` by default, which implies "Read ~at most~ _exactly_ nb bytes..."
```
help?> readbytes!
search: readbytes!

  readbytes!(stream::IO, b::AbstractVector{UInt8}, nb=length(b))

  Read at most nb bytes from stream into b, returning the number of bytes
  read. The size of b will be increased if needed (i.e. if nb is greater than
  length(b) and enough bytes could be read), but it will never be decreased.

  ────────────────────────────────────────────────────────────────────────────

  readbytes!(stream::IOStream, b::AbstractVector{UInt8}, nb=length(b); all::Bool=true)

  Read at most nb bytes from stream into b, returning the number of bytes
  read. The size of b will be increased if needed (i.e. if nb is greater than
  length(b) and enough bytes could be read), but it will never be decreased.

  See read for a description of the all option.
```

`read(::IO)` has the same problem:

```
  read(s::IO, nb=typemax(Int))

  Read at most nb bytes from s, returning a Vector{UInt8} of the bytes read.

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  read(s::IOStream, nb::Integer; all=true)

  Read at most nb bytes from s, returning a Vector{UInt8} of the bytes read.

  If all is true (the default), this function will block repeatedly trying to read all requested bytes, until an error or
[...]

PR #338, opened 2018-10-20
Improve streaming interface. Reduce allocation and buffering.

samoconnor:
These changes:
 - Improve the streaming interface (see #337, #339) by implementing more of the Base `IO` interface for `HTTP.Stream`: `close`, `unsafe_read`, `read(io, UInt8)`, `read(io, n)`, `.readbytes!` etc.
 - Change the internal read implementation to use `unsafe_read` (rather than `readavailable`) where possible to reduce allocation and buffering.
 - Replace `Connection.excess::ByteView` with `Connection.buffer::IOBuffer` and remove `IOExtras.unread`. The `buffer` is used to read headers and chunk sizes (who's byte-length can't be determined in advance).

The previous `ByteView` / `unread!` scheme avoided copying data between buffers by twiddling views into the buffer returned by `readavailable`. However, using `readavailable` implies allocating a new buffer on every MbedTLS read. With these changes, the method `Base.readuntil(t::Transaction, f::Function)` reads small blocks of bytes into an `IOBuffer` until the end of the header is found.

https://github.com/JuliaWeb/HTTP.jl/blob/a77beb6dde496a65c49917d9e3480862c9d5384f/src/ConnectionPool.jl#L220-L230

 The `IOBuffer` remains allocated for the lifetime of the `Connection`. The excess bytes after the header are consumed by the next call to one of the `read*` methods. After that, the `read*` methods are passed directly to the underlying `MbedTLS.Context`.

Consider an audio streaming example:
```julia
samples = Vector{UInt8}(undef, 16384)
HTTP.open("GET", url) do http
    while !eof(http)
        n = readbytes!(http, samples)
        play_audio_samples(samples, n)
    end
end
```

The `readbytes!` call will occasionally result in an internal read to `Connection.buffer::IOBuffer` to read HTTP headers or HTTP chunk size. However, most of the time the `ccall` to `mbedtls_ssl_read` will now be reading bytes directly into the `samples` buffer.

Needs Tests

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -3,23 +3,13 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[BinaryProvider]]
 deps = ["Libdl", "Pkg", "SHA", "Test"]
-git-tree-sha1 = "b530fbeb6f41ab5a83fbe3db1fcbe879334bcd2d"
+git-tree-sha1 = "9930c1a6cd49d9fcd7218df6be417e6ae4f1468a"
 uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
-version = "0.4.2"
-
-[[Compat]]
-deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
-git-tree-sha1 = "ae262fa91da6a74e8937add6b613f58cd56cdad4"
-uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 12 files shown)


samoconnor:
Closing because GitHub UI is not showing new commits to this branch.

samoconnor:
Closing because GitHub UI is not showing new commits to this branch.

samoconnor:
Closing because GitHub UI is not showing new commits to this branch.

samoconnor:
Wow. GuitHub is FUBAR 😫

ssfrr:
Looks great, just tested the following on this branch and it seems to be doing great, and now works within my other library that takes an `IO` object and reads from it with `readbytes!`.

```julia
const STREAM_URL = "http://doppler.media.mit.edu/impoundment.opus"
buf = Vector{UInt8}(undef, 10000)
HTTP.open("GET", STREAM_URL) do io
    read!(io, buf)
    close(io)
end
```

Thanks!

samoconnor:
Any objections to merging this now?

ssfrr:
eerie. I was just checking the status on this. great news!

Issue #343, opened 2018-10-27
reuse_limit=0 should send "Connection: close" header


samoconnor:
... as should the last request before the `reuse_limit`.

PR #346, opened 2018-11-01
fix #344 and #343

### LINKED ISSUE #344: closeread(::HTTP.Stream) should check for "close" before reading excess body data
There is no need to do this
```julia
    # Discard body bytes that were not read...
    while !eof(http)
        readavailable(http)
    end
```
If the connection is being closed anyway.
See #339

samoconnor:
fix #344 and #343


samoconnor:
Any objections to merging this now?

PR #349, opened 2018-11-06
Use basename for Content-Disposition filename on multipart form data

christopher-dG:
[MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#Directives): 
> path information should be stripped

[RFC 6266](https://tools.ietf.org/html/rfc6266#section-4.3) is a bit less concrete but seems to recommend it: 
> One strategy to achieve this is to never trust folder name information in the filename parameter, for instance by stripping all but the last path segment and only considering the actual filename

### DIFF
--- a/src/multipart.jl
+++ b/src/multipart.jl
@@ -80,7 +80,7 @@ function Form(d)
 end
 
 function writemultipartheader(io::IOBuffer, i::IOStream)
-    write(io, "; filename=\"$(i.name[7:end-1])\"\r\n")
+    write(io, "; filename=\"$(basename(i.name[7:end-1]))\"\r\n")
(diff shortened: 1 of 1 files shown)


oxinabox:
Content Disposition is one of the few fields of HTTP that has comprehensive test urls  available
There should be a few  URLs from this 
that will allow you to test the behavour

http://test.greenbytes.de/tech/tc2231/

oxinabox:
Oh, but this is upload not download?
In which case sounds like the server's problem

christopher-dG:
Yeah this is just uploads.

oxinabox:
I feel this is legit.

quinnj:
Sorry for the delay here @christopher-dG, thanks for contributing.

PR #352, opened 2018-11-27
Deprecate the t kwarg to timestamp in sign_aws4!

ararslan:
`t` is fine for a positional argument name, since those aren't really user-facing, but `t` is an insufficiently descriptive name for a keyword argument. This PR deprecates it in favor of a `timestamp` keyword argument.


ararslan:
Failure on 32-bit Windows with Julia nightly is unrelated.

PR #354, opened 2018-11-30
Make DEBUG_LEVEL a Ref{Int} instead of an Int

ararslan:
This is a quality of life improvement for local use. One can now simply write `HTTP.DEBUG_LEVEL[] = 2`, for example, to set the appropriate level.

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -667,7 +667,7 @@ function Base.show(io::IO, c::Connection)
         " ≣", c.pipeline_limit,
         length(c.excess) > 0 ? " $(length(c.excess))-byte excess" : "",
         nwaiting > 0 ? " $nwaiting bytes waiting" : "",
-        DEBUG_LEVEL > 1 && applicable(tcpsocket, c.io) ?
+        DEBUG_LEVEL[] > 1 && applicable(tcpsocket, c.io) ?
             " $(Base._fd(tcpsocket(c.io)))" : "")
 end
 
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -2,7 +2,7 @@ module HTTP
 
 export startwrite, startread, closewrite, closeread
 
-const DEBUG_LEVEL = 0
+const DEBUG_LEVEL = Ref(0)
 
 Base.@deprecate escape escapeuri
 Base.@deprecate URL URI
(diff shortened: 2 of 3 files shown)


samoconnor:
How does this interact with preoccupation/JIT compilation?

i.e. If do the debug macros still evaluate to `:()` by default?
If I do `HTTP.DEBUG_LEVEL[] = 2` from the REPL after `using HTTP`, are all the methods that were compiled with the old value automatically recompiled with the new value?

ararslan:
`Ref` is actually what's recommended for this kind of package-global setting particularly because of how it interacts with precompilation. Doing something like `@eval HTTP DEBUG_LEVEL = 2` is both unsavory and does not cause anything to be recompiled.

> do the debug macros still evaluate to `:()` by default?

Yes, since the variable still defaults to 0.

> If I do `HTTP.DEBUG_LEVEL[] = 2` from the REPL after `using HTTP`, are all the methods that were compiled with the old value automatically recompiled with the new value?

They aren't recompiled, they're compiled once using the `Ref`, the value of which is looked up at runtime via the call to `getindex`.

samoconnor:
> They aren't recompiled ...

So doing `HTTP.DEBUG_LEVEL[] = 2` will have no effect unless it is done before any methods are compiled? I'm not sure I understand the use case for this. Is there a guarantee that right after `using HTTP` none of the methods form the HTTP package will have been compiled yet?

ararslan:
No, changing the debug level can be done at any time, regardless of whether things have been compiled, since it's checking the value at runtime.

samoconnor:
But the `@debug` macro queries `DEBUG_LEVEL` at compile time:

```julia
macro debug(n::Int, s)
    DEBUG_LEVEL[] >= n ? :(println(debug_header(), $(esc(s)))) :
                         :()
end
```

The macro either evaluates to a `println()` statement or to an empty expression at compile time.

ararslan:
Oh, I see what you mean. Fixed in https://github.com/JuliaWeb/HTTP.jl/pull/356.

samoconnor:
But #356 seems to defeat the whole purpose of the `@debug` macro, which was to ensure zero overhead when `DEBUG_LEVEL` is `0`. Doing a runtime comparison might have low overhead at runtime, but having the comparison and the calls to `println` and `debug_header` in the AST can change the inlining an type inference behaviour of the compiler.

ararslan:
It will not affect type inference unless you put a call to `@debug` as the last thing in the function, where it might affect the return value. As for inlining, ¯\\\_(ツ)\_/¯

fredrikekre:
You can extract the printing etc to a `@noinline` function.

samoconnor:
We have [Base Logging](https://docs.julialang.org/en/v1/stdlib/Logging/index.html) for places where a runtime logging level check is what we want.
The intention of `HTTP.jl/src/debug.jl` is to allow super-verbose debug messages for use by package developers but have zero overhead when disabled. (we occasionally ask end-users to set `DEBUG_LEVEL` to assist with issue support).

ararslan:
I am not convinced that there will be any noticeable overhead.

Issue #291, opened 2018-09-07
Test suite for `sign_aws4!`


omus:
It would be good to have a test suite for this function. Amazon provides a [test suite](https://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html) which we could use in order to validate the behaviour of this function.

omus:
I did a manual test and I needed to modify what headers were being generated by `sign_aws4!` but it did pass this test:

```julia
using Base.Test, HTTP.AWS4AuthRequest, HTTP.URIs, HTTP.Headers

# Modify internals of `sign_aws4!` so that only these headers are set
# setkv(headers, "x-amz-date",  datetime)
# setkv(headers, "host",  url.host)

# get-vanilla-query-order-key-case
result = HTTP.AWS4AuthRequest.sign_aws4!(
    "GET", 
    URI("https://example.amazonaws.com/?Param2=value2&Param1=value1"), 
    Headers([]), 
    b""; 
    t=DateTime(2015,8,30,12,36),
    aws_service="service",
    aws_region="us-east-1",
    aws_secret_access_key="wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY", 
    aws_access_key_id="AKIDEXAMPLE",
)

expected = "Authorization" => "AWS4-HMAC-SHA256 Credential=AKIDEXAMPLE/20150830/us-east-1/service/aws4_request, SignedHeaders=host;x-amz-date, Signature=b97d918cfa904a5beff61c982a1b6f458b799221646efd99d3219ec94cdf2500"

@test result == expected
```

I'll also note that `sign_aws4!` did not pass the "get-header-key-duplicate" test.

ararslan:
Something to note is that we send more headers (specifically `x-amz-content-sha256` and `Content-MD5`) than Amazon includes in their examples, so we can't necessarily compare our output to theirs 1-1 unless we rethink the headers we want to send. Technically the content hashes are sort of redundant, since the SHA-256 of the content is used in constructing the signature itself, and the request is denied if the sent signature doesn't match the signature that Amazon calculates upon receiving the request.

omus:
I think either we could make sure that our headers are a superset of the ones in the test or we can reevaluate why we have the additional headers.

quinnj:
It's also only required for PUT S3 requests; GET requests don't need it.

ararslan:
Closed by the addition of tests in #357, #359, #364, #366, #368, and #369.

PR #357, opened 2018-12-04
AWS4: Header changes and tests

ararslan:
This PR consists of 4 related but functionally distinct commits:

* Include `host` unconditionally in the headers. This matches all of Amazon's examples, though it is not required.
* Make the `Content-MD5` header optional. This is recommended but not required, and omitting it aids in testing against Amazon's official test cases. The default here is to include it (no change in behavior).
* Make the `x-amz-content-sha256` header optional when it isn't required. It is only required for S3 PUT requests. The default here is to always include it (no change in behavior).
* Add a few basic tests that pass from Amazon's test suite. This is a first step toward #291.

### DIFF
--- a/src/AWS4AuthRequest.jl
+++ b/src/AWS4AuthRequest.jl
@@ -47,6 +47,7 @@ function sign_aws4!(method::String,
                     aws_access_key_id::String=ENV["AWS_ACCESS_KEY_ID"],
                     aws_secret_access_key::String=ENV["AWS_SECRET_ACCESS_KEY"],
                     aws_session_token::String=get(ENV, "AWS_SESSION_TOKEN", ""),
+                    include_md5=true,
                     kw...)
     if t !== nothing
         Base.depwarn("The `t` keyword argument to `sign_aws4!` is deprecated; use " *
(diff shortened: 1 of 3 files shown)


test/aws4.jl
```diff
@@ -0,0 +1,43 @@
...
+using Test
+using HTTP
+using HTTP: Headers, URI
+using HTTP.AWS4AuthRequest: sign_aws4!
+
+# Based on https://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html
+
+@testset "AWS Signature Version 4" begin
+    withenv("AWS_ACCESS_KEY_ID" => "AKIDEXAMPLE",
+            "AWS_SECRET_ACCESS_KEY" => "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY") do
```
omus: Would be good to add a comment mentioning these aren't real credentials. I sure hope they aren't real.

src/AWS4AuthRequest.jl
```diff
@@ -75,9 +77,12 @@ function sign_aws4!(method::String,
 
     # HTTP headers...
     rmkv(headers, "Authorization")
-    setkv(headers, "x-amz-content-sha256",  content_hash)
-    setkv(headers, "x-amz-date",  datetime)
-    setkv(headers, "Content-MD5", base64encode(body_md5))
+    setkv(headers, "host", url.host)
+    setkv(headers, "x-amz-date", datetime)
+    include_md5 && setkv(headers, "Content-MD5", base64encode(body_md5))
+    if (aws_service == "s3" && method == "PUT") || include_sha256
```
omus: It would be good to have a reference to the S3 documentation that says this is required

omus:
LGTM

PR #359, opened 2018-12-05
Fix and test edge cases in handling of AWS4 headers

ararslan:
In particular, we were not correctly handling duplicate and multiline headers.

### DIFF
--- a/src/AWS4AuthRequest.jl
+++ b/src/AWS4AuthRequest.jl
@@ -34,6 +34,19 @@ function request(::Type{AWS4AuthLayer{Next}},
     return request(Next, url, req, body; kw...)
 end
 
+# Normalize whitespace to the form required in the canonical headers.
+# Note that the expected format for multiline headers seems not to be explicitly
+# documented, but Amazon provides a test case for it, so we'll match that behavior.
+# We replace each `\n` with a `,` and remove all whitespace around the newlines,
+# then any remaining contiguous whitespace is replaced with a single space.
+function _normalize_ws(s::AbstractString)
+    if any(isequal('\n'), s)
+        join(map(_normalize_ws, split(s, '\n')), ',')
+    else
+        replace(strip(s), r"\s+" => " ")
+    end
+end
+
 function sign_aws4!(method::String,
                     url::URI,
                     headers::Headers,
@@ -90,10 +103,19 @@ function sign_aws4!(method::String,
     end
 
     # Sort and lowercase() Headers to produce canonical form...
-    canonical_headers = sort!(map(headers) do (k, v)
-        string(lowercase(k), ':', replace(strip(v), r"\s+" => " "))
-    end)
-    signed_headers = join(sort!(map(lowercase∘first, headers)), ";")
+    unique_header_keys = Vector{String}()
+    normalized_headers = Dict{String,Vector{String}}()
+    for (k, v) in sort!([lowercase(k) => v for (k, v) in headers], by=first)
+        if !haskey(normalized_headers, k)
+            normalized_headers[k] = Vector{String}()
+            push!(unique_header_keys, k)
+        end
+        push!(normalized_headers[k], _normalize_ws(v))
+    end
+    canonical_headers = map(unique_header_keys) do k
+        string(k, ':', join(normalized_headers[k], ','))
+    end
+    signed_headers = join(unique_header_keys, ';')
 
     # Sort Query String...
     query = sort!(collect(queryparams(url.query)), by=first)
--- a/test/aws4.jl
+++ b/test/aws4.jl
@@ -6,39 +6,84 @@ using HTTP.AWS4AuthRequest: sign_aws4!
 
 # Based on https://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html
 
-@testset "AWS Signature Version 4" begin
-    # NOTE: These are the example credentials as specified in the AWS docs, they are not real
-    withenv("AWS_ACCESS_KEY_ID" => "AKIDEXAMPLE",
-            "AWS_SECRET_ACCESS_KEY" => "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY") do
+function test_get!(headers, params; opts...)
+    sign_aws4!("GET",
+               URI("https://example.amazonaws.com/" * params),
+               headers,
+               UInt8[];
+               timestamp=DateTime(2015, 8, 30, 12, 36),
+               aws_service="service",
+               aws_region="us-east-1",
+               # NOTE: These are the example credentials as specified in the AWS docs,
+               # they are not real
+               aws_access_key_id="AKIDEXAMPLE",
+               aws_secret_access_key="wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY",
+               include_md5=false,
+               include_sha256=false,
+               opts...)
+    headers
+end
 
-        cases = [
-            ("get-vanilla", "", "5fa00fa31553b73ebf1942676e86291e8372ff2a2260956d9b8aae1d763fbf31"),
-            ("get-vanilla-empty-query-key", "?Param1=value1", "a67d582fa61cc504c4bae71f336f98b97f1ea3c7a6bfe1b6e45aec72011b9aeb"),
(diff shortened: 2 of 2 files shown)


src/AWS4AuthRequest.jl
```diff
@@ -34,6 +34,17 @@ function request(::Type{AWS4AuthLayer{Next}},
     return request(Next, url, req, body; kw...)
 end
 
+# Normalize whitespace to the form required in the canonical headers.
+# Note that the expected format for multiline headers seems not to be explicitly
+# documented, but Amazon provides a test case for it, so we'll match that behavior.
+function _normalize_ws(s::AbstractString)
```
oxinabox: Can we just fully spell that out:
```
_normalize_whitespace
```

ararslan: It's an internal function used in one place and I have explanatory comments at the definition and at the call site, so that seems unnecessary to me.

src/AWS4AuthRequest.jl
```diff
@@ -90,10 +101,14 @@ function sign_aws4!(method::String,
     end
 
     # Sort and lowercase() Headers to produce canonical form...
-    canonical_headers = sort!(map(headers) do (k, v)
-        string(lowercase(k), ':', replace(strip(v), r"\s+" => " "))
-    end)
-    signed_headers = join(sort!(map(lowercase∘first, headers)), ";")
+    unique_headers = sort!(unique!(map(lowercase∘first, headers)))
```
oxinabox: `unique_headers_keys`

src/AWS4AuthRequest.jl
```diff
@@ -90,10 +101,14 @@ function sign_aws4!(method::String,
...
+    canonical_headers = map(unique_headers) do k
+        # All values present for this key (case-insensitive) with whitespace normalized
+        vs = map(i->_normalize_ws(last(headers[i])),
+                 findall(x->lowercase(first(x)) == k, headers))
+        string(lowercase(k), ':', join(vs, ','))
```
oxinabox:
```suggestion
        string(k, ':', join(vs, ','))
```

`k` is alrady lowercase

src/AWS4AuthRequest.jl
```diff
@@ -90,10 +101,14 @@ function sign_aws4!(method::String,
...
     # Sort and lowercase() Headers to produce canonical form...
-    canonical_headers = sort!(map(headers) do (k, v)
-        string(lowercase(k), ':', replace(strip(v), r"\s+" => " "))
-    end)
-    signed_headers = join(sort!(map(lowercase∘first, headers)), ";")
+    unique_headers = sort!(unique!(map(lowercase∘first, headers)))
+    canonical_headers = map(unique_headers) do k
+        # All values present for this key (case-insensitive) with whitespace normalized
+        vs = map(i->_normalize_ws(last(headers[i])),
+                 findall(x->lowercase(first(x)) == k, headers))
```
oxinabox: This map + findall is kind scary
I think just writing out the for loop is cleaner

Something like: 
```
vs = Vector{Header}() # is this eltype right?
for header in headers
	header_key = lowercase(first(header))
	if header_key = k
		header_value
		push!(vs, _normalize_ws(header_value))
	end
end
```

src/AWS4AuthRequest.jl
oxinabox: The comment at the callsite wouldn't be required if it wasn't true that `ws` is a unclear acronym.

src/AWS4AuthRequest.jl
```diff
@@ -34,6 +34,17 @@ function request(::Type{AWS4AuthLayer{Next}},
...
 
+# Normalize whitespace to the form required in the canonical headers.
+# Note that the expected format for multiline headers seems not to be explicitly
+# documented, but Amazon provides a test case for it, so we'll match that behavior.
+function _normalize_ws(s::AbstractString)
+    if any(isequal('\n'), s)
+        join(map(_normalize_ws, split(s, '\n')), ',')
+    else
+        replace(strip(s), r"\s+" => " ")
+    end
```
omus: There seems to be some subtleties that are worth commenting about. So you want to replace each `\n` with a `,` and also remove all whitespace around the newlines. Any remaining contiguous whitespace is replaced with a single space.

src/AWS4AuthRequest.jl
```diff
@@ -90,10 +101,14 @@ function sign_aws4!(method::String,
...
-        string(lowercase(k), ':', replace(strip(v), r"\s+" => " "))
-    end)
-    signed_headers = join(sort!(map(lowercase∘first, headers)), ";")
+    unique_header_keys = sort!(unique!(map(lowercase∘first, headers)))
+    canonical_headers = map(unique_header_keys) do k
+        # All values present for this key (case-insensitive) with whitespace normalized
+        vs = map(i->_normalize_ws(last(headers[i])),
+                 findall(x->lowercase(first(x)) == k, headers))
+        string(k, ':', join(vs, ','))
+    end
```
omus: Alternative:
```julia
using DataStructures: DefaultOrderedDict
normalized_headers = DefaultOrderedDict{String,Vector{String}}(() -> String[])
for k in sort!(collect(keys(headers)))
    ck = lowercase(k)
    push!(canonical_headers[ck], _normalize_ws(header[k]))
end
canonical_headers = [string(k, ':', join(vs, ',')) for (k, vs) in normalized_headers]
unique_header_keys = collect(keys(normalized_headers))
```
or for dependency haters:
```julia
unique_header_keys = Vector{String}()
normalized_headers = Dict{String,Vector{String}}()
for k in sort!(collect(keys(headers)))
    ck = lowercase(k)

    if !haskey(canonical_headers, ck)
        canonical_headers[ck] = Vector{String}()
    end

    push!(canonical_headers[ck], _normalize_ws(header[k]))
    push!(normalized_headers, ck)
end
canonical_headers = [string(k, ':', join(vs, ',')) for (k, vs) in normalized_headers]
```
The main advantage to this approach is you avoid some iterations

test/aws4.jl
```diff
@@ -6,38 +6,84 @@ using HTTP.AWS4AuthRequest: sign_aws4!
...
+               include_sha256=false,
+               opts...)
+    headers
+end
+
+test_auth_string(headers, sig) =
+    string("AWS4-HMAC-SHA256 ",
+           "Credential=AKIDEXAMPLE/20150830/us-east-1/service/aws4_request, ",
+           "SignedHeaders=", headers, ", ",
+           "Signature=", sig)
```
omus: Hard to read:
```julia
function test_auth_string(headers, sig) 
    d = [
        "AWS4-HMAC-SHA256 Credential" => "AKIDEXAMPLE/20150830/us-east-1/service/aws4_request",
        "SignedHeaders" => headers,
        "Signature" => sig,
    ]
    join(("$k=$v" for (k, v) in d), ", ")
end
```

test/aws4.jl
```diff
@@ -6,38 +6,84 @@ using HTTP.AWS4AuthRequest: sign_aws4!
...
-                              "Signature=", sig)
-                @test d["Authorization"] == auth
-            end
+    noheaders = [
+        ("get-vanilla", "", "5fa00fa31553b73ebf1942676e86291e8372ff2a2260956d9b8aae1d763fbf31"),
+        ("get-vanilla-empty-query-key", "?Param1=value1", "a67d582fa61cc504c4bae71f336f98b97f1ea3c7a6bfe1b6e45aec72011b9aeb"),
+        ("get-utf8", "ሴ", "8318018e0b0f223aa2bbf98705b62bb787dc9c0e678f255a891fd03141be5d85"),
+    ]
+    for (name, p, sig) in noheaders
+        @testset "$name" begin
```
omus: Note that: `@testset "$name" for (name, p, sig) in noheaders` is a thing

src/AWS4AuthRequest.jl
ararslan: I had to revert this change, as it did not produce the correct result.

omus: All is well now

ararslan:
The hashes produced on 32-bit Windows seem to be different. I am _very_ thoroughly annoyed by this,

ararslan:
I can reproduce in a 32-bit Linux VM. Time to investigate...

ararslan:
64-bit Windows failure with Julia nightly seems to be a network issue.

PR #363, opened 2018-12-10
Add URI path normalization

ararslan:
In particular, removal of dot segments per RFC 3986, section 5.2.4.

### DIFF
--- a/src/URIs.jl
+++ b/src/URIs.jl
@@ -196,14 +196,14 @@ https://tools.ietf.org/html/rfc3986#section-4.3
 isabsolute(uri::URI) =
     !isempty(uri.scheme) &&
      isempty(uri.fragment) &&
-    (isempty(uri.host) || isempty(uri.path) || pathissabsolute(uri))
+    (isempty(uri.host) || isempty(uri.path) || isabspath(uri))
 
 
 """
 https://tools.ietf.org/html/rfc7230#section-5.3.1
 https://tools.ietf.org/html/rfc3986#section-3.3
 """
-pathissabsolute(uri::URI) = startwith(uri.path, "/")
+isabspath(uri::URI) = startswith(uri.path, "/") && !startswith(uri.path, "//")
 
 
 ==(a::URI,b::URI) = a.scheme      == b.scheme      &&
@@ -378,6 +378,45 @@ function splitpath(p::AbstractString)
         (i > len || c in ('?', '#')) && break
     end
     return elems
+end
+
+"""
+    URIs.normpath(url)
+
+Normalize the path portion of a URI by removing dot segments.
+
+Refer to:
+* https://tools.ietf.org/html/rfc3986#section-5.2.4
+"""
+normpath(url::URI) =
+    URI(scheme=url.scheme, userinfo=url.userinfo, host=url.host, port=url.port,
+        path=normpath(url.path), query=url.query, fragment=url.fragment)
+
+function normpath(p::AbstractString)
+    if isempty(p) || p == "/"
+        return p
+    elseif p == "." || p == ".."
+        return "/"
+    end
+    buf = String[]
+    for part in splitpath(p)
+        if part == "."
+            continue
+        elseif part == ".."
+            isempty(buf) || pop!(buf)
+        else
+            push!(buf, part)
+        end
+    end
+    out = join(buf, '/')
+    # Preserve leading and trailing slashes if present, but don't duplicate them
+    if startswith(p, '/') && !startswith(out, '/')
+        out = "/" * out
+    end
+    if (endswith(p, '/') || endswith(p, '.')) && !endswith(out, '/')
+        out *= "/"
+    end
+    out
 end
 
 absuri(u, context) = absuri(URI(u), URI(context))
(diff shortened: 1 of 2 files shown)


test/uri.jl
```diff
@@ -477,4 +478,27 @@ end
...
+        checknp("http://a", "http://a")
+        checknp("http://a/b/c/.", "http://a/b/c/")
+        checknp("http://a/b/c/./", "http://a/b/c/")
+        checknp("http://a/b/c/..", "http://a/b/")
+        checknp("http://a/b/c/../", "http://a/b/")
+        checknp("http://a/b/c/../g", "http://a/b/g")
+        checknp("http://a/b/c/../..", "http://a/")
+        checknp("http://a/b/c/../../", "http://a/")
+        checknp("http://a/b/c/../../g", "http://a/g")
+    end
```
oxinabox: If anyone, like me, was wondering, difference from `Base.normpath`
according to these tests are:

`Base.normpath` tranlates `http://` into `http:/`.

`Base.normpath` translates `http://a/b/c/d/../../../../g` into , "http:/g")

test/uri.jl
```diff
@@ -477,4 +478,27 @@ end
     # Issue 323
     @test string(HTTP.URI(scheme="http", host="example.com")) == "http://example.com"
 
+    @testset "Normalize URI paths" begin
+        # Examples given in https://tools.ietf.org/html/rfc3986#section-5.2.4
+        @test URIs.normpath("/a/b/c/./../../g") == "/a/g"
+        @test URIs.normpath("mid/content=5/../6") == "mid/6"
+
+        checknp = (x, y)->(@test URIs.normpath(URI(x)) == URI(y))
```
oxinabox: why not write ? 
```suggestion
        checknp(x, y) = @test URIs.normpath(URI(x)) == URI(y)
```

(For my education)

src/URIs.jl
```diff
@@ -380,6 +380,45 @@ function splitpath(p::AbstractString)
...
 
+"""
+    URIs.normpath(url)
+
+Normalize the path portion of a URI by removing dot segments.
+
+Refer to:
+* https://tools.ietf.org/html/rfc3986#section-5.2.4
+"""
+normpath(url::URI) =
```
oxinabox: I feel like this should be `Base.normpath(url::URI)`,
but that does mean renaming the `normpath(p::AbstractString)`
so it won't collide.

src/URIs.jl
```diff
@@ -380,6 +380,45 @@ function splitpath(p::AbstractString)
...
+        path=normpath(url.path), query=url.query, fragment=url.fragment)
+
+function normpath(p::AbstractString)
+    if isempty(p) || p == "/"
+        return p
+    elseif p == "." || p == ".."
+        return "/"
+    end
+    buf = String[]
+    for part in splitpath(p)
```
oxinabox: `Base.splitpath` is in julia 1.1 only.
(or is there a `splitpath` on HTTP master that I am not aware of)
Current bounds on this package is julia 0.7

Am I right in saying
`Base.splitpath` has different functionality on windows than linux?
And that will make this wrong on windows?
https://github.com/JuliaLang/julia/pull/28156/files#diff-e0580027daf242717c633ff9cf06a750R139

src/URIs.jl
```diff
@@ -380,6 +380,45 @@ function splitpath(p::AbstractString)
...
+            continue
+        elseif part == ".."
+            isempty(buf) || pop!(buf)
+        else
+            push!(buf, part)
+        end
+    end
+    out = join(buf, '/')
+    # Preserve leading and trailing slashes if present, but don't duplicate them
+    if startswith(p, '/') && !startswith(out, '/')
```
oxinabox: I am not sure how one can get a leading `/` in the `out`.
Can you add a test case that causes this? (or comment one that is already there)

ararslan: Dunno, it's just a sanity check. I'm not sure how it would happen.

src/URIs.jl
ararslan: None of the path-related names from Base are imported/extended in the URIs submodule. In this case, `splitpath` refers to the function defined above this one.

src/URIs.jl
oxinabox: perhaps change it to an `@assert` ?

src/URIs.jl
ararslan: See the comment below regarding `splitpath`; this module uses Base names but contains them in the URIs namespace. This is just following suit with the other such instances.

test/uri.jl
ararslan: That would probably be fine, I'm just never sure of the exact scoping/context rules for defining functions inside of `@testset`s, so I figured it's easier to just assign a closure.

src/URIs.jl
oxinabox: Ah, it was hidden in the fold, and is in master but not in the release branch I am currently on.

test/uri.jl
ararslan: The difference is because `http://a` is not considered part of the path; the path begins with the first `/` after `a`.

src/URIs.jl
oxinabox: (It is wrong there, and it is wrong here. But I guess being wrong consistantly is better ...)

PR #364, opened 2018-12-13
Fix and test path normalization for AWS4 signing

ararslan:
Paths should be normalized but currently we aren't doing that.


ararslan:
AppVeyor failure is network issues.

PR #366, opened 2018-12-13
 Add option to exclude the session token from signing 

ararslan:
This PR consists of two commits, the first of which simply provides tests for AWS4 POST requests. The second adds an option to exclude the session token when calculating the signature. According to Amazon some services want it in the signature and some do not. The header is always included if the token is available, and it's included in signing by default, which means there's no change in default behavior.

### DIFF
--- a/src/AWS4AuthRequest.jl
+++ b/src/AWS4AuthRequest.jl
@@ -60,6 +60,7 @@ function sign_aws4!(method::String,
                     aws_access_key_id::String=ENV["AWS_ACCESS_KEY_ID"],
                     aws_secret_access_key::String=ENV["AWS_SECRET_ACCESS_KEY"],
                     aws_session_token::String=get(ENV, "AWS_SESSION_TOKEN", ""),
+                    token_in_signature=true,
                     include_md5=true,
                     include_sha256=true,
                     kw...)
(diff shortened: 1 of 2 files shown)


src/AWS4AuthRequest.jl
```diff
@@ -60,6 +60,7 @@ function sign_aws4!(method::String,
                     aws_access_key_id::String=ENV["AWS_ACCESS_KEY_ID"],
                     aws_secret_access_key::String=ENV["AWS_SECRET_ACCESS_KEY"],
                     aws_session_token::String=get(ENV, "AWS_SESSION_TOKEN", ""),
+                    include_token_in_sig=true,
```
omus: Rename to `token_in_signature`? The `include_*` keywords below indicate if a header is included.

src/AWS4AuthRequest.jl
```diff
@@ -98,7 +99,8 @@ function sign_aws4!(method::String,
         # https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
         setkv(headers, "x-amz-content-sha256", content_hash)
     end
-    if aws_session_token != ""
+    # Some services don't want the token as part of the signature
```
omus: Comment is more sensible as:

> Some services want to include the security token as part of the signature

PR #368, opened 2018-12-14
Exclude Content-Length from AWS4 signing

ararslan:
This is done in Amazon's examples but does not appear to be addressed in the documentation, so we'll just follow suit with the example.

### DIFF
--- a/src/AWS4AuthRequest.jl
+++ b/src/AWS4AuthRequest.jl
@@ -111,6 +111,11 @@ function sign_aws4!(method::String,
         if k == "x-amz-security-token" && !token_in_signature
             continue
         end
+        # In Amazon's examples, they exclude Content-Length from signing. This does not
+        # appear to be addressed in the documentation, so we'll just mimic the example.
+        if k == "content-length"
+            continue
+        end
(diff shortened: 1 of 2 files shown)


src/AWS4AuthRequest.jl
```diff
@@ -111,6 +111,11 @@ function sign_aws4!(method::String,
         if k == "x-amz-security-token" && !token_in_signature
             continue
         end
+        # In Amazon's examples, they exclude Content-Length from signing. This does not
+        # appear to be addressed in the documentation, so we'll just mimic the example.
+        if k == "content-length"
+            continue
+        end
```
omus: I'd make this a one-liner but that's just me ¯\_(ツ)_/¯

ararslan: I was going to, it just looked a bit funny to me with the long-form `if` above it. I can change it if that's better, I don't feel strongly (or have any particular feelings at all) about it.

ararslan: Yeah, going back and changing it now I think it looks out of place since there are long-form `if`s both above and below it. So I'll opt to keep it this time.

PR #369, opened 2018-12-14
AWS4: Escape S3 paths and add tests

ararslan:
The documentation says that the paths for S3 requests should be escaped, but currently we're neither escaping nor normalizing them. Per text included in the AWS4 test suite, we'll continue not to normalize them, but we should be escaping.


ararslan:
I believe that this, combined with https://github.com/JuliaWeb/HTTP.jl/pull/368, sufficiently closes #291.

PR #391, opened 2019-03-02
Check for headers correctly in download

oxinabox:
There was a problem that @mschauer  found with

```
download("https://www.dropbox.com/s/uy6j8wru1nkqhw5/coal.csv?dl=1")
```
returning the wrong filename.
It was coming back with `"coal.csv?dl=1"`,
but the correct name according to headers is `"coal.csv"`.

I took a look in the headers and realized that they were casing all in lowercase.
Which is allowed, but the download  function was doing a case sensitive check on the list of headers,
rather than using the `header` function on the response.

This PR fixes that.

Could add a test based on the above, but seems a bit fragile.

### DIFF
--- a/src/download.jl
+++ b/src/download.jl
@@ -16,9 +16,9 @@ function safer_joinpath(basepart, parts...)
     joinpath(basepart, parts...)
 end
 
-function try_get_filename_from_headers(headers)
-    content_disp = getkv(headers, "Content-Disposition")
-    if content_disp != nothing
+function try_get_filename_from_headers(resp)
+    content_disp = header(resp, "Content-Disposition")
+    if content_disp != ""
         # extract out of Content-Disposition line
         # rough version of what is needed in https://github.com/JuliaWeb/HTTP.jl/issues/179
         filename_part = match(r"filename\s*=\s*(.*)", content_disp)
@@ -54,9 +54,9 @@ function determine_file(path, resp)
     name = if isdir(path)
         # got to to workout what file to put there
         filename = something(
-                        try_get_filename_from_headers(resp.headers),
+                        try_get_filename_from_headers(resp),
                         try_get_filename_from_remote_path(resp.request.target),
-                        basename(tempname()) # fallback, basically a random string
+                        basename(tempname())  # fallback, basically a random string
                     )
         safer_joinpath(path, filename)
     else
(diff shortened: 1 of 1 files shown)


oxinabox:
On the way past, I added a thing to break  early if the  stream  `eof`  as it seems that sometimes (all the time?) at  first  time  the  stream bit is called, the stream is not actually ready,
which means it fails to read the filename  from the header, so it defaults the file name,
then makes an empty file there, as it has no body to write to it.

In the cases where the filename is part of the URL this doesn't matter as the original blank file is overwritten. But when it has to be found from the header, it means you get an extra incorrectly  named  empty file left behind

mschauer:
Can confirm that this solves the problems with Dropbox links. 👍

oxinabox:
Julia 1.0 tests failing on linux because
https://github.com/JuliaWeb/MbedTLS.jl/issues/193

mschauer:
In the meanwhile I added your branch as dependency, https://github.com/mschauer/PointProcessInference.jl/blob/master/Manifest.toml

But somehow it still can go wrong, as if DataDeps does use its own dependency

For example, when running
```julia
pkg> add https://github.com/mschauer/PointProcessInference.jl

julia> observations, parameters, λinfo = PointProcessInference.loadexample("coal")
```
I got the old behaviour.

fredrikekre:
`PointProcessInference.jl/Manifest.toml` is not used when you add a package, so you will simply get the released version of HTTP.

src/download.jl
```diff
@@ -52,15 +52,16 @@ determine_file(::Nothing, resp) = determine_file(tempdir(), resp)
 function determine_file(path, resp)
     # get the name
     name = if isdir(path)
+        # We have been given a path too a directory
```
mschauer:
```suggestion
        # we have been given a path to a directory
```

mschauer:
Thank you I was not aware of that. To be of help, I went through the code and found no issues (I also checked that the calls to header do the right thing when the key is not found.)

oxinabox:
Til this is merged you can always get back on this branch by making sure it is added in your current active envirioment.

mschauer:
Just keeping this on your screen, as you suggested, oxinabox.

oxinabox:
@quinnj @samoconnor  can we merge this?
The CI failure is not real

src/download.jl
```diff
@@ -99,11 +100,14 @@ function download(url::AbstractString, local_path=nothing, headers=Header[]; upd
...
         resp = startread(stream)
+        eof(stream) && return  # don't do anything for streams we can't read (yet)
+        
         file = determine_file(local_path, resp)
-        total_bytes = parse(Float64, getkv(resp.headers, "Content-Length", "NaN"))
+        total_bytes = parse(Float64, header(resp, "Content-Length", "NaN"))
         downloaded_bytes = 0
         start_time = now()
         prev_time = now()
+
```
oxinabox:
```suggestion
```

Issue #392, opened 2019-03-04
Stream is closed or unusable


pfitzseb:
With Julia 1.1 and HTTP 0.8.0 I'm getting a super weird error once the route is hit:
```
julia> using HTTP, Sockets

julia> ROUTER = HTTP.Router()
HTTP.Handlers.Router{Symbol("##363")}(HTTP.Handlers.RequestHandlerFunction{getfield(HTTP.Handlers, Symbol("##1#2"))}(getfield(HTTP.Handlers, Symbol("##1#2"))()), Dict{HTTP.Handlers.Route,Any}(), ^[[BDict{String,Val}())

julia> function test(req::HTTP.Request, args...)
           @show req
           @show resp =  HTTP.Response(
               400,
               [
                   "Access-Control-Allow-Origin" => "*"
               ]
           )
           return resp
       end
test (generic function with 1 method)

julia> HTTP.@register(ROUTER, "POST", "/test", test)

julia> HTTP.serve(ROUTER, Sockets.localhost, 8181)
┌ Error: error handling request
│   exception =
│    IOError: stream is closed or unusable
│    Stacktrace:
│     [1] check_open at ./stream.jl:323 [inlined]
│     [2] uv_write_async(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:871
│     [3] uv_write(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:845
│     [4] unsafe_write(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:901
│     [5] unsafe_write at /home/pfitzseb/.julia/packages/HTTP/GN0Te/src/ConnectionPool.jl:132 [inlined]
│     [6] macro expansion at ./gcutils.jl:87 [inlined]
│     [7] write at ./strings/io.jl:165 [inlined]
│     [8] writestartline(::HTTP.ConnectionPool.Transaction{TCPSocket}, ::HTTP.Messages.Response) at /home/pfitzseb/.julia/packages/HTTP/GN0Te/src/Messages.jl:428
│     [9] writeheaders(::HTTP.ConnectionPool.Transaction{TCPSocket}, ::HTTP.Messages.Response) at /home/pfitzseb/.julia/packages/HTTP/GN0Te/src/Messages.jl:439
[...]

Issue #376, opened 2019-01-01
How can I keep the websocket alive?


Hugo-Trentesaux:
After around one minute of successful exchange between a Julia websocket server and a browser client, I get a connection timeout. I guess the browser closes the connection (because Julia does not). So the question is: how can I ping the client so he keeps the connection alive?

I tried to write the WS_PING byte in the websocket, but it does not even send the message.

zot:
I'm not positive on this but it appears that check_readtimeout, from the new release, is what is disconnecting the sockets upgrading to a websocket does not appear to cancel the timeout behavior. My code didn't timeout from the previous release but it does now.

Hugo-Trentesaux:
Thanks, it will help me to debug it

samoconnor:
@quinnj I haven't played with WebSockets ping-pong. However I think when #341 is fixed it shouldn't be needed.

quinnj:
@samoconnor, I think we should either remove the websockets code or document that it shouldn't be used as it seems there are no real maintainers of the code. Thoughts?

zot:
Wow, I'm sure using it!

quinnj:
@zot, the main issue is the websockets code in HTTP.jl hasn't gotten nearly enough love since being added. If you're willing to step up and help maintain the code, we would gladly welcome the contributions. As is though, there are several outstanding websocket-related issues and no one working on them. It just makes me uncomfortable having that code in HTTP.jl and users expecting it to be at the same quality/standard as the rest of the package.

zot:
WebSockets.jl seems to be actively maintained and it even imports a version of this package that is known to work with it. I suspect that the connection timeout behavior will still not play well with upgraded web socket connections though, since they reuse the socket and the timeout will close the connection.

I think there might need to be a way to protect an existing connection from timing out so that WebSockets.jl can use that to protect the connection from being closed from outside. Like maybe add a "notimeout" flag to Connection that check_readtimeout() can look at before closing the connection (maybe use it instead of wait_for_timeout?).

samoconnor:
> @samoconnor, I think we should either remove the websockets code or document that it shouldn't be used as it seems there are no real maintainers of the code. Thoughts?

@quinnj are you aware of any current WebSockets issues that are non-#341 related ?

PR #393, opened 2019-03-06
Fix Websocket timeout issues

### LINKED ISSUE #341: listen readtimeout option
A couple of issues with the `readtimeout` option:
 - when no data is received for 60 seconds `HTTP/1.1 408 Request Time-out\r\n\r\n` is sent in the middle of the response data stream even though a response header has already been sent.
 - `check_readtimeout` aborts the connection even after the request has been successfully received and the response data is being streamed.

https://github.com/JuliaWeb/HTTP.jl/blob/df13276971d66a7cd9c54cfb88a5d0000773d419/src/Servers.jl#L304-L318

For now I think we should disable this feature by default. @quinnj ?

To be useful, the timeout should probably only be triggered if we are actually waiting on more data.
e.g., if we haven't yet found the end of the header, or if there was a content-length header and we're still waiting on more body data, or if we're midway through reading a chunk.

Also, if we've already sent a successful response header, then it makes no sense to send status 408.
https://tools.ietf.org/html/rfc7231#section-6.5.7

The timeout should probably be suspended as soon as the handler writes a response header.
If a handler wants to read more body data after sending a response header, and it want's a timeout for that, it should probably implement that itself with semantics that make sense for the particular application.

HBaghdadi1995:
There was an issue where WebSockets would automatically timeout after a minute (#341)

# Changes
## ConnectionPool.jl
- Added `idle_timeout_enabled::Bool` field to connection
- Ensured connection doesn't timeout if `idle_timeout_enabled::Bool == false`
## Servers.jl
- Added `idle_timeout_enabled = true` keyword argument to `Server.listen`
- Added `idle_timeout_enabled = true` argument to `Server.listenloop`
## WebSockets.jl
- `WebSockets.listen` opens up a websocket without causing it to crash after a minute.

Should resolve #341, #376 & #392

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -70,7 +70,6 @@ mutable struct Connection{T <: IO}
     peerport::UInt16 # debug only
     localport::UInt16 # debug only
     io::T
-    idle_timeout_enabled::Bool
     buffer::IOBuffer
     sequence::Int
     writecount::Int
@@ -96,13 +95,13 @@ end
 
 Connection(host::AbstractString, port::AbstractString,
            pipeline_limit::Int, idle_timeout::Int,
-           require_ssl_verification::Bool, io::T, idle_timeout_enabled = true) where T <: IO =
+           require_ssl_verification::Bool, io::T) where T <: IO =
     Connection{T}(host, port,
                   pipeline_limit, idle_timeout,
                   require_ssl_verification,
                   peerport(io), localport(io),
-                  io, idle_timeout_enabled,
-                  PipeBuffer(), -1,
+                  io, PipeBuffer(),
+                  -1,
                   0, false, Condition(),
                   0, false, Condition(),
                   time())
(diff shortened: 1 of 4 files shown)


HBaghdadi1995:
if we're with #394 and then parts of this Pr should be migrated to websockets

samoconnor:
I don't think there should be any need for `idle_timeout_enabled::Bool` in  `ConnectionPool.jl`. The only place it is used is in `inactiveseconds`, which is called by `check_readtimeout` which is never called unless `readtimeout > 0`.

I think the fix should be to change the default of `readtimeout` to `0` here: https://github.com/JuliaWeb/HTTP.jl/blob/master/src/Servers.jl#L198

Also, this should probably be reverted: https://github.com/JuliaWeb/HTTP.jl/commit/086157b9ba343b3df5c0abe46ff726c67ca2fafa

HBaghdadi1995:
Thanks Sam, I've taken your feedback and applied changes based on them.

PR #396, opened 2019-03-09
Expand `Server` convenience type and use in `listen` and `listenloop`

EricForgy:
In this PR, I expanded the definition of the convenience type `Server` to include all input to `listen`.

```julia
Base.@kwdef mutable struct Server{T <: Union{MbedTLS.SSLConfig, Nothing}}
    host::Union{IPAddr, String}=Sockets.localhost
    port::Integer=8081
    ssl::T=nothing
    tcpisvalid::Function=tcp->true
    server::Union{Base.IOServer, Nothing}=nothing
    reuseaddr::Bool=false
    connection_count::Ref{Int}=Ref(0)
    rate_limit::Union{Rational{Int}, Nothing}=nothing
    reuse_limit::Int=nolimit
    readtimeout::Int=60
    verbose::Bool=false
end
```

This will be helpful for some things / simplifications I want to do with WebSockets.jl.

I also exported `Server` from `Servers` so it can be conveniently accessed as

```julia
HTTP.Server()
```

instead of

```julia
HTTP.Servers.Server()
```

With this, we can do things like:

```julia
julia> HTTP.listen(f,HTTP.Server())
````

to kick up a quick server.

On the WebSockets.jl side, we have a similar convenience type: `ServerWS`. 

I'd like to add `Server` to `ServerWS` so we can have easy control of the HTTP server including a nice `Base.close(s::ServerWS)` method.

Generally speaking, I think having a more complete convenience type is... well... convenient 😊 

PS: The minor change to runtests.jl is to make it easier to comment out tests when debugging.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -7,7 +7,7 @@ optional keyword arguments.  For full details, see `?HTTP.listen`.
 """
 module Servers
 
-export listen
+export listen, Server
 
 using ..IOExtras
 using ..Streams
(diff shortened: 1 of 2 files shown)


EricForgy:
Btw...

Note that whatever we do, `HTTP.listen` is always constructing the smaller convenience type anyway:

https://github.com/JuliaWeb/HTTP.jl/blob/master/src/Servers.jl#L223

and feeds this into `HTTP.listenloop` along with several other server parameters. 

This PR simply pulls that convenience type out so it can be passed in (and monitored) rather than always constructing it on the fly.

EricForgy:
I think that test must have been a fluke. It seems unlikely 7 out of 8 environments pass.

EricForgy:
Again, another 7 out of 8 environments passed, but this time, the one that failed was a different environment. These tests seem to not be very deterministic. Could we make some of these tests `@test_skip` so the test suite doesn't fail? 🤔

quinnj:
Closing as stale and I'm not sure this really provides much utility on the server-side

PR #402, opened 2019-03-30
typos fix

tlienart:


src/Streams.jl
```diff
@@ -25,7 +25,7 @@ mutable struct Stream{M <: Message, S <: IO} <: IO
 end
 
 """
-    Stream(::IO, ::Request)
+    Stream(::Request, ::IO)
```
tlienart: Might be nice in the doc to refer to the fact that a request is a message (?)

Issue #405, opened 2019-04-01
Request after server closure is still served


tlienart:
Long story short: I believe there are circumstances in which @async threads remain alive even after closing the server that is being listened on. This means that you can still access pages as if they are still being served while the status of the server is closed (6). This can cause problems when starting and closing a server multiple times in one Julia session. I present a MWE below that reproduces the problem + a possible fix.

Julia 1.2
HTTP.jl 0.8#master
Mac
latest Safari & Firefox

## MWE

**Edit**: Jonas provided a much simpler example [below](https://github.com/JuliaWeb/HTTP.jl/issues/405#issuecomment-479227654).

Here's a simple example with a single websocket:

**index.html**
```html
<!DOCTYPE html>
<html><body>Some text</body><html>
```

**script**
```julia
using HTTP
using Sockets
using Test

const JSWS = """
    <script type="text/javascript">
      var sock = new WebSocket("ws://" + location.host + location.pathname);
      sock.onmessage = function(msg){sock.close();};
    </script>
    """

server = Sockets.listen(8000)

listener = @async HTTP.listen(Sockets.localhost, 8000, server=server) do http::HTTP.Stream
    if HTTP.WebSockets.is_upgrade(http.message)
        println("upgrade")
        HTTP.setstatus(http, 101)
        HTTP.setheader(http, "Upgrade" => "websocket")
        HTTP.setheader(http, "Connection" => "Upgrade")
        key = HTTP.header(http, "Sec-WebSocket-Key")
        HTTP.setheader(http, "Sec-WebSocket-Accept" => HTTP.WebSockets.accept_hash(key))
        HTTP.startwrite(http)

        io = http.stream
        ws = HTTP.WebSockets.WebSocket(io; server=true)
        write(ws, "ping")
        println("pinged")
    elseif http.message.method == "GET" && http.message.target ∈ ("/", "/index.html")
        h = HTTP.Handlers.RequestHandlerFunction() do req::HTTP.Request
            content = read("index.html", String)
            content = replace(content, "</body>" => "$JSWS</body>")
            HTTP.Response(200, content)
         end
        HTTP.Handlers.handle(h, http)
    end
end
```

What this does:

* serves `index.html` after appending a short javascript which opens a websocket and closes it upon receiving a message. 

What is problematic:

_sequence of actions_

* open a browser at localhost, works fine, will write "upgrade" and "pinged" in the console
* in Julia:

```julia
close(server)
@test server.status == 6
[...]

PR #406, opened 2019-04-02
Suggested fix for #405 (propagate information to all async threads upon close(server))

tlienart:
This is the suggested fix in #405. It makes sure that if you have

```julia
server = Sockets.listen(port)
@async HTTP.listen(Sockets.localhost, port, server=server) ... 

# open browser, do stuff
close(server)
```

all the threads from the `listenloop` are properly closed and that attempting to connect to `localhost:port` via a browser will show an "unable to connect" error message as it should.

________________________

**Edit**: after a short chat on Slack with Jacob, it seems that this solution is not ideal and that a better one would be to keep track of all open connections and subsequently close all of those when the server is closed. I still think this PR could be a temporary fix which could be reverted when the other solution is implemented but happy to close the PR if this is not desirable.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -246,7 +246,7 @@ function listenloop(f, server, tcpisvalid, connection_count,
             let io=io, count=count
                 @async try
                     verbose && @info "Accept ($count):  $conn"
-                    handle_connection(f, conn, reuse_limit, readtimeout)
+                    handle_connection(f, conn, reuse_limit, readtimeout, server)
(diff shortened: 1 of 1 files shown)


tlienart:
Bump. (sorry it's a rather annoying bug for our package 😕; thanks!)

quinnj:
Sorry for the delay @tlienart; I'm proposing #419 as a simpler fix. If you could try it out, I'd appreciate it.

PR #409, opened 2019-04-19
HTTP.serve should default to localhost

fredrikekre:
According to the docs, the host in HTTP.serve should default to localhost.


quinnj:
Thanks @fredrikekre !

PR #419, opened 2019-05-30
An alternative fix for #405; basically when handling a connection, le…

quinnj:
…t's keep track of the original server socket; if it closes, then we stop taking new requests on our connection socket.

cc: @tlienart, @asprionj 

This is a simpler fix than proposed in #406, while achieving the same result (hopefully!). Sorry this has taken so long to address.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -246,7 +246,7 @@ function listenloop(f, server, tcpisvalid, connection_count,
             let io=io, count=count
                 @async try
                     verbose && @info "Accept ($count):  $conn"
-                    handle_connection(f, conn, reuse_limit, readtimeout)
+                    handle_connection(f, conn, server, reuse_limit, readtimeout)
                     verbose && @info "Closed ($count):  $conn"
                 catch e
                     if e isa Base.IOError && e.code == -54
@@ -261,9 +261,9 @@ function listenloop(f, server, tcpisvalid, connection_count,
                 end
             end
         catch e
+            close(server)
             if e isa InterruptException
                 @warn "Interrupted: listen($server)"
-                close(server)
                 break
             else
                 rethrow(e)
@@ -280,14 +280,15 @@ Create a `Transaction` object for each HTTP Request received.
 After `reuse_limit + 1` transactions, signal `final_transaction` to the
 transaction handler.
 """
-function handle_connection(f, c::Connection, reuse_limit, readtimeout)
+function handle_connection(f, c::Connection, server, reuse_limit, readtimeout)
     wait_for_timeout = Ref{Bool}(true)
     if readtimeout > 0
         @async check_readtimeout(c, readtimeout, wait_for_timeout)
     end
     try
         count = 0
-        while isopen(c)
+        # if the connection socket or original server close, we stop taking requests
+        while isopen(c) && isopen(server)
             handle_transaction(f, Transaction(c);
                                final_transaction=(count == reuse_limit))
             count += 1


src/Servers.jl
```diff
@@ -280,14 +280,15 @@ Create a `Transaction` object for each HTTP Request received.
...
         @async check_readtimeout(c, readtimeout, wait_for_timeout)
     end
     try
         count = 0
-        while isopen(c)
+        # if the connection socket or original server close, we stop taking requests
+        while isopen(c) && isopen(server)
             handle_transaction(f, Transaction(c);
                                final_transaction=(count == reuse_limit))
             count += 1
```
tlienart: Isn't there a problem here if `isopen(c)` is true but `isopen(server)` isn't? i.e. the connection should be closed shouldn't it?

quinnj: It pretty much gets closed immediately by the listenloop once `handle_connection` returns

tlienart:
Ok so I tried the example described here: https://github.com/JuliaWeb/HTTP.jl/issues/405#issuecomment-479227654

`close(server)` effectively closes but throws an ugly error:

```
julia> ┌ Error: error handling request
│   exception =
│    IOError: stream is closed or unusable
│    Stacktrace:
│     [1] check_open at ./stream.jl:323 [inlined]
│     [2] uv_write_async(::TCPSocket, ::Ptr{UInt8}, ::UInt64, ::Task) at ./stream.jl:901
│     [3] uv_write(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:877
│     [4] unsafe_write(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:931
│     [5] unsafe_write at /Users/tlienart/.julia/packages/HTTP/yRq0H/src/ConnectionPool.jl:134 [inlined]
│     [6] macro expansion at ./gcutils.jl:87 [inlined]
│     [7] write at ./strings/io.jl:177 [inlined]
│     [8] writestartline(::HTTP.ConnectionPool.Transaction{TCPSocket}, ::HTTP.Messages.Response) at /Users/tlienart/.julia/packages/HTTP/yRq0H/src/Messages.jl:426
│     [9] writeheaders(::HTTP.ConnectionPool.Transaction{TCPSocket}, ::HTTP.Messages.Response) at /Users/tlienart/.julia/packages/HTTP/yRq0H/src/Messages.jl:437
│     [10] startwrite(::HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{TCPSocket}}) at /Users/tlienart/.julia/packages/HTTP/yRq0H/src/Streams.jl:85
│     [11] handle(::HTTP.Handlers.RequestHandlerFunction{getfield(Main, Symbol("##5#8"))}, ::HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{TCPSocket}}) at /Users/tlienart/.julia/packages/HTTP/yRq0H/src/Handlers.jl:277
│     [12] #4 at ./REPL[6]:2 [inlined]
│     [13] macro expansion at /Users/tlienart/.julia/packages/HTTP/yRq0H/src/Servers.jl:353 [inlined]
│     [14] (::getfield(HTTP.Servers, Symbol("##13#14")){getfield(Main, Symbol("##4#7")),HTTP.ConnectionPool.Transaction{TCPSocket},HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{TCPSocket}}})() at ./task.jl:268
└ @ HTTP.Servers ~/.julia/packages/HTTP/yRq0H/src/Servers.jl:357


```

quinnj:
Please do try it out, I'd appreciate it. As I dug thru the code, I think this should be sufficient. The reason is that once we get a new connection, the code will fall down to the `startread` call until the actual request comes over the wire. Once the request is received, it calls the handler and then pretty much immediately a new transaction is started and another `startread` is called (which causes it to wait). So basically, we'll always finish a request that got started, but never allow a new one to start.

Oh, I should also mention that an easy way to avoid this behavior w/ the current release is to do `HTTP.listen(; reuse_limit=0)`, which means that a connection will never allow more than one request.

tlienart:
Yeah with `reuse_limit=0` I'm not able to reproduce the error: 


```julia
using HTTP
using Sockets

function foo(server)
   listener = @async HTTP.listen(ip"127.0.0.1", server=server, reuse_limit=0) do http::HTTP.Stream
       HTTP.handle(HTTP.RequestHandlerFunction(req -> HTTP.Response(200, "hi")), http)
   end
end

# doing the next two lines a bunch of time seems to work fine, no error.
server = Sockets.listen(8000); foo(server)
close(server)
```

PR #421, opened 2019-05-31
removing @error in catch block

tlienart:
Following #419 I did a fair bit more testing and it still occasionally throws a IOStream, stream unusable error (see 1st comment in #419; this is also with `reuse_limit=0`). Removing this error line seems to fix the problem.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -354,7 +354,6 @@ function handle_transaction(f, t::Transaction; final_transaction::Bool=false)
         closeread(http)
         closewrite(http)
     catch e
-        @error "error handling request" exception=(e, stacktrace(catch_backtrace()))
         if isopen(http) && !iswritable(http)
             http.message.response.status = 500
             startwrite(http)


quinnj:
We can't remove this. There are a multitude of reasons handling a connection might fail and we need to report this. I actually think the stream is closed error log is very valid here and shouldn't be suppressed; the server was closed while some connections were still open and then operations were attempted on those connections as they were closed down (previously, those operations were allowed to continue). I think this could be useful in cases where the server is ***unexpectedly*** shut down for some reason.

tlienart:
I have two questions then:

1. how would you elegantly catch this error so that it does never show up? in LiveServer (and downstream usecases) we just want to start a server, listen, and close upon request, if at time of closing anything is open, all of that should be forcibly shut. As the code currently stands, this will happen but an error will be thrown which does not seem necessary. 
2. the line precedes a block of actions, are these actions ever applied?

quinnj:
1. I would suggest using `reuse_limit=0`; in the case you're describing, you want to avoid these kinds of errors when a connection gets reused after a server gets closed, so just don't let connections get reused
2. The line you're proposing to remove is an error ***log***, not a control-flow statement ***throwing*** an error; the error will be propagated elsewhere

tlienart:
1. the error still happens **with** `reuse_limit=0`. 
2. ok then I guess removing it changes nothing, I'll close this PR

quinnj:
img

quinnj:
> the error still happens with reuse_limit=0. (asprionj/LiveServer.jl#73)
Ah! That definitely shouldn't happen. Ok, I think I've got a fix for this. Thanks for being patient with me.

tlienart:
No worries, and sorry, it's difficult because my understanding of the problem is limited and lib_uv errors are very confusing (and hard to reproduce... :'( )

PR #423, opened 2019-05-31
Add support for using proxy env variables if set. Fixes #403

### LINKED ISSUE #403: Proxy support - Automatically picking from environment
>This issue is a follow up after #218 and https://github.com/JuliaWeb/HTTP.jl/issues/218#issuecomment-449367875

Currently, for using proxy support, we have to pass it as a parameter. eg - `HTTP.request("GET", url, proxy=proxy)`.
However, a lot of packages like Github.jl, DataDeps.jl ..... have HTTP.jl as a dependency. For these 200+ packages to operate behind a proxy, the packages would have to be edited and have an optional parameter for the proxy to be passed. 

Instead of this, I think it will be better to have HTTP.jl use the _system/environment proxy_ (if present in the system), besides also having the option to override it. (the same way it currently is)

quinnj:
I'd appreciate if someone could help look over and test this.

### DIFF
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -9,7 +9,7 @@ using MbedTLS: SSLContext
 import ..@debug, ..DEBUG_LEVEL
 
 # hasdotsuffix reports whether s ends in "."+suffix.
-hasdotsuffix(s, suffix) = length(s) > length(suffix) && s[length(s)-length(suffix)] == '.' && s[length(s)-length(suffix)+1:end] == suffix
+hasdotsuffix(s, suffix) = endswith(s, "." * suffix)
 
 function isnoproxy(host)
     for x in NO_PROXY
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -18,6 +18,8 @@ function isnoproxy(host)
     return false
 end
 
+const NO_PROXY = String[]
+
 function __init__()
     # check for no_proxy environment variable
     if haskey(ENV, "no_proxy")
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -29,14 +29,11 @@ function __init__()
 end
 
 function getproxy(scheme, host)
+    isnoproxy(host) && return nothing
     if scheme == "http" && haskey(ENV, "http_proxy")
-        if !isnoproxy(host)
-            return ENV["http_proxy"]
-        end
+        return ENV["http_proxy"]
     elseif scheme == "https" && haskey(ENV, "https_proxy")
-        if !isnoproxy(host)
-            return ENV["https_proxy"]
-        end
+        return ENV["https_proxy"]
     end
     return nothing
 end
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -8,6 +8,43 @@ using ..ConnectionPool
 using MbedTLS: SSLContext
 import ..@debug, ..DEBUG_LEVEL
 
+const HTTP_PROXY = Ref{String}()
+const HTTPS_PROXY = Ref{String}()
+const NO_PROXY = Set{String}()
+
+gethttp_proxy() = isdefined(HTTP_PROXY, :x) ? HTTP_PROXY[] : nothing
+gethttps_proxy() = isdefined(HTTPS_PROXY, :x) ? HTTPS_PROXY[] : nothing
+
+getproxy(scheme) = scheme == "http" ? gethttp_proxy() : scheme == "https" ? gethttps_proxy() : nothing
+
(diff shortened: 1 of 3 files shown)


src/ConnectionRequest.jl
```diff
@@ -8,6 +8,43 @@ using ..ConnectionPool
...
 import ..@debug, ..DEBUG_LEVEL
 
+const HTTP_PROXY = Ref{String}()
+const HTTPS_PROXY = Ref{String}()
+const NO_PROXY = Set{String}()
+
+gethttp_proxy() = isdefined(HTTP_PROXY, :x) ? HTTP_PROXY[] : nothing
+gethttps_proxy() = isdefined(HTTPS_PROXY, :x) ? HTTPS_PROXY[] : nothing
+
+getproxy(scheme) = scheme == "http" ? gethttp_proxy() : scheme == "https" ? gethttps_proxy() : nothing
```
oxinabox: Why not read the `ENV` var in these functions?
For a long running REPL I can imagine people wanting to change proxy settings  without restarting.

And the code would be simpler

quinnj: This will get called for ***every*** request, so I'm a little worried about performance, but maybe that wouldn't cost too much.

oxinabox:
```
julia> @btime (()->ENV["PATH"])()
  39.791 ns (1 allocation: 144 bytes)
```

```
julia> const Path = ENV["PATH"];

julia> @btime (()->Path)()
  1.487 ns (0 allocations: 0 bytes)
```


I am leaning towards it being negliabale compared to the IO time.

quinnj: Ok, I've refactored it a bit.

EricForgy:
I like the idea of having a `getproxy` method, but is it possible to design it so we can define our own `getproxy`? For example, I might prefer to get the proxy from a config file.

EricForgy:
Maybe have it be a `proxyHandlerFunction` or something? 🤔

oxinabox:
> I like the idea of having a getproxy method, but is it possible to design it so we can define our own getproxy? For example, I might prefer to get the proxy from a config file.

Raise an issue for it when and if it occurs.

To be honest it might be most reasonable, becuase it is a user spefic setting to do it via money-patching.
maybe even adding to your startup.jl:

```
using Requires
@require HTTP="cd3eb016-35fb-5094-929b-558a96fad6f3" begin
HTTP.getproxy(scheme, host) = "MyProxy"
end
```

src/ConnectionRequest.jl
```diff
@@ -8,6 +8,39 @@ using ..ConnectionPool
 using MbedTLS: SSLContext
 import ..@debug, ..DEBUG_LEVEL
 
+# hasdotsuffix reports whether s ends in "."+suffix.
+hasdotsuffix(s, suffix) = length(s) > length(suffix) && s[length(s)-length(suffix)] == '.' && s[length(s)-length(suffix)+1:end] == suffix
```
oxinabox:
```suggestion
hasdotsuffix(s, suffix) = endswith(s, "." * suffix)
```

src/ConnectionRequest.jl
```diff
@@ -8,6 +8,39 @@ using ..ConnectionPool
...
+    if haskey(ENV, "no_proxy")
+        for x in split(ENV["no_proxy"], ","; keepempty=false)
+            push!(NO_PROXY, startswith(x, ".") ? x[2:end] : x)
+        end
+    end
+    return
+end
+
+function getproxy(scheme, host)
+    if scheme == "http" && haskey(ENV, "http_proxy")
```
oxinabox:
```suggestion
    isnoproxy(host) && return nothing
    if scheme == "http" && haskey(ENV, "http_proxy")
```

And remove the `if`s below.

EricForgy:
> Raise an issue for it when and if it occurs.

I don't mind raising a separate issue, but I bring it up here because this HAS occurred for me 😅 

I was directed here by Jacob after asking about this on Slack 😊 

Maybe we can sort this out in this PR before merging? Not a big deal. We can go ahead and merge this and I create a separate issue if that is preferred. Either way, I think this is a good step in the right direction 👍

oxinabox:
Shouldn't the enviroment variables be in upper case?

Also might want to support  `ALL_PROXY` too
idk how common that is, but apparently curl supports it.

oxinabox:
> > 
> 
> I don't mind raising a separate issue, but I bring it up here because this HAS occurred for me

fair enough. What do you think of my monkey-patching idea?

EricForgy:
> What do you think of my monkey-patching idea?

Not sure 🤔 

Would that cause a warning because you are overriding a method with the same signature?

oxinabox:
Not in Julia 1.0 IIRC

PR #427, opened 2019-06-17
WIP/RFC: added parser for multipart form

acdupont:
* Rebase @piever parseform branch from PR #264 to HTTP.jl master
* Add few tweaks

I'll work on any issues to get this PR accepted

### DIFF
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -1,40 +1,68 @@
-const RETURN_BYTES = [0x0d, 0x0a]
-const DASH_BYTE = 0x2d
+const CR_BYTE = 0x0d # \r
+const LF_BYTE = 0x0a # \n
+const DASH_BYTE = 0x2d # -
+const HTAB_BYTE = 0x09 # \t
+const SPACE_BYTE = 0x20
+const RETURN_BYTES = [CR_BYTE, LF_BYTE]
 
 const FORMDATA_REGEX = r"Content-Disposition: form-data"
 const NAME_REGEX = r" name=\"(.*?)\""
 const FILENAME_REGEX = r" filename=\"(.*?)\""
 const CONTENTTYPE_REGEX = r"Content-Type: (\S*)"
 
-function find_boundary(bytes::AbstractVector{UInt8}, str::AbstractVector{UInt8}, dashes; start::Int = 1)
-    l = length(bytes)
+"""
+Returns the first and last index of the boundary delimiting a part, and if the discovered
+boundary is the terminating boundary.
+"""
+function find_boundary(bytes::AbstractVector{UInt8}, boundaryDelimiter::AbstractVector{UInt8}; start::Int = 1)
+    # The boundary delimiter line is prepended with two '-' characters
+    # The boundary delimiter line starts on a new line, so must be preceded by a \r\n.
+    # The boundary delimiter line ends with \r\n, and can have "optional linear whitespace" between
+    # the end of the boundary delimiter, and the \r\n.
+    # The last boundary delimiter line has an additional '--' at the end of the boundary delimiter
+    # [RFC2046 5.1.1](https://tools.ietf.org/html/rfc2046#section-5.1.1)
+
     i = start
-    cons_dash = 0
-    while i+length(str) <= l
-        @inbounds cons_dash = (bytes[i] == DASH_BYTE) ? cons_dash + 1 : 0
-        if cons_dash >= dashes && bytes[(i+1):(i+length(str))] == str
-            return [i-dashes, i+length(str)]
+    endIndex = i+length(boundaryDelimiter)-1
+    while endIndex <= length(bytes)
+        if bytes[i:endIndex] == boundaryDelimiter
+            # boundary delimiter line start on a new line ...
+            if i > 1
+                (i == 2 || bytes[i-2] != CR_BYTE || bytes[i-1] != LF_BYTE) && error("boundary delimiter found, but it was not the start of a line")
+                i-=2
+            end
+
+            # need to check if there are enough characters for the CRLF or for two dashes
+            endIndex < length(bytes)-1 || error("boundary delimiter found, but did not end with new line")
+
+            isTerminatingDelimiter = bytes[endIndex+1] == DASH_BYTE && bytes[endIndex+2] == DASH_BYTE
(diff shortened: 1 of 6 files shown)


quinnj:
Thanks for picking this up @acdupont! Just let me know when you think this is ready to review/merge and I can take a look.

rssdev10:
Hi, any progress in merging of this feature?

acdupont:
We are using the feature in our own code base, but I haven't had the time to write tests using RFCs as a reference. I am still planning on doing this.

src/parsemultipart.jl
```diff
@@ -0,0 +1,100 @@
...
+        j = find_boundary(bytes, str, dashes; start = j[2]+1)
+    end
+    return idxs
+end
+
+function find_boundaries(bytes::AbstractVector{UInt8}, boundary; start::Int = 1)
+    m =  match(r"^(-*)(.*)$", boundary)
+    m === nothing && return nothing
+    d, str = m[1], m[2]
+    find_boundaries(bytes, unsafe_wrap(Array{UInt8, 1}, String(str)), length(d); start = start)
```
c42f: This `unsafe_wrap` combined with the `String` construction looks a bit odd. Is it necessary? (I wondered if it required a `GC.@preserve` but it seems the resulting array knows about the string as the data owner so you should be safe there.)

Also, can you replace the regex match with a simple search `findfirst(c->c != '-', boundary)`? It looks `match(r"^(-*)(.*)$", boundary)` will always succeed except when there's a newline in `boundary`. Is that the intended behavior?

acdupont: I added my first test this evening, I'll take a look at these two items next.

acdupont: I pushed a refactor that addressed the use of the regular expression and attempts to follow RFC 2046 and RFC 7578, the code is completely different. I am not finished refactoring though:
* I am looking at the code that parses the "chunk" now.
* I'd like to change the find_boundary naive string parsing to something like Knuth–Morris–Pratt algorithm (or is there a regex that can search byte arrays in Julia?)

Also, I changed the code that writes to an IO stored in the Multipart to creating an IOBuffer around a view of the body sent from the Request. 
https://github.com/JuliaWeb/HTTP.jl/pull/427/commits/d7d54b2767e8d4c6156b8e00dc58190cfb21263b#diff-1be8563c7ec166272d8a3e9201b93db6R105

I point this out because I don't know if that will cause issues in real code once the Request goes away, although my test ran fine. Does that look okay?

acdupont:
@quinnj  and @c42f, this PR is at a state where it can be reviewed.

test/parsemultipart.jl
```diff
@@ -0,0 +1,93 @@
+using Test
+using HTTP
+
+
+function generateTestBody()
+    IOBuffer("----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"namevalue\"; filename=\"multipart.txt\"\r\nContent-Type: text/plain\r\n\r\nnot much to say\n\r\n----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"key1\"\r\n\r\n1\r\n----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"key2\"\r\n\r\nkey the second\r\n----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"namevalue2\"; filename=\"multipart-leading-newline.txt\"\r\nContent-Type: text/plain\r\n\r\n\nfile with leading newline\n\r\n----------------------------918073721150061572809433--\r\n").data
```
c42f: Function name should be `generate_test_body`? Ditto for `generateTestRequest`, etc.

Also, can/should you use the `@b_str` macro here rather than the `IOBuffer()` trick?

amellnik:
@acdupont Are you still working on this?  If you would like, I can take a pass at the issues that @c42f mentioned.

acdupont:
@amellnik I haven't had time to get back to this, it would be awesome if you picked it up.

c42f:
That would be fantastic @amellnik. I don't have time to work on this myself (other than providing some review), but we're using a hacked-up version of this in our own application-specific repo and it has proved useful so far.

nlw0:
I'm very interested in this feature, how can anyone help? Is the next step now to go trough the review?

acdupont:
The questions on August 16th from @c42f are the next iteration that need to be addressed.

pixel27:
I'm looking at implementing the changes mentioned by @c42f and have some questions.

First about the regex being defined at the top of the file. Should I leave that?  My own tests say that if you do an r"" at the call the regex compile is only called once, but I'm not an expert.  

Second my read of https://tools.ietf.org/html/rfc2183 suggests that he was correct we should handle the non quoted file name.  Which brings in a whole need to actually parse the line.  Is there a minimum version of Julia HTTP should work with?  I ask because there is a findnext() method that takes a character but that was only introduced in 1.3.

Lastly I'm not exactly sure how to submit my changes to acdupont's pull request....I could issue pull request to his repository or create a new pull request to this repository from my repository which was created of his repository...or is there something else?  I'm not real knowledgeable about github or git for that matter....I know enough to be dangerous but that's all I claim to know.

pixel27:
Oh one more thing, there was a comment that dropping the multipart while correct may make it difficult to debug.  Should I add a @warn message?  I see some use of @warn but they don't seem to be around malformed HTTP requests.  Is there another way I should indicate that the request does not appear valid?

c42f:
> First about the regex being defined at the top of the file. Should I leave that

I would change it; if the regexes are only used once it should be clearer and just as efficient to move them to the use site.

>  Is there a minimum version of Julia HTTP should work with

You can find a hint in Project.toml — HTTP still claims compatibility with 0.7 so I think it's safe to say you need to be conservative for now. I think you may use the predicate form `findnext(==(c), str, i)` form for `c::Char`.

> Lastly I'm not exactly sure how to submit my changes to acdupont's pull request....I could issue pull request to his repository or create a new pull request to this repository from my repository which was created of his repository...or is there something else?

I suggest it's simplest to create a new PR with your own repository and branch as the source by pulling this branch from @acdupont's repository and building on top of it in your own repo. Just put a link to this discussion on the new PR with a quick mention of the history of the code.

quinnj:
This was finished/merged in #514

PR #432, opened 2019-07-10
Update async error log message

nicoleepp:
I've noticed a lot of confusion about this warning and what it means. The immediate assumption made is that something is wrong and needs fixing but usually the case is that everything is fine (other than the connection closing before expected) and continues normally.

Preferably we could remove this msg altogether and rely on the exception thrown if the read fails, but this at least de-escalates the urgency of this msg and adds a bit of extra info.

Some more context here: https://github.com/JuliaWeb/HTTP.jl/pull/285/files/f4e2516abcb0df0a24f11460c2f9a782af2e5f5f#r213862000

### DIFF
--- a/src/StreamRequest.jl
+++ b/src/StreamRequest.jl
@@ -61,7 +61,10 @@ function request(::Type{StreamLayer}, io::IO, request::Request, body;
                           "Server likely closed the connection unexpectedly. " *
                           "Only an issue if unable to read the response and this error gets re-thrown. ",
                           exception=(write_error, catch_backtrace()))
-                    close(io)
+                    try
+                        close(io)
+                    catch e
+                    end
                 end
                 yield()
                 startread(http)
--- a/src/StreamRequest.jl
+++ b/src/StreamRequest.jl
@@ -57,7 +57,9 @@ function request(::Type{StreamLayer}, io::IO, request::Request, body;
                     writebody(http, request, body)
                 catch e
                     write_error = e
-                    @warn("Error in @async writebody task",
+                    @info("Error in @async writebody task. " *
+                          "Server likely closed the connection unexpectedly. " *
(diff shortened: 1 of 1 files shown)


src/StreamRequest.jl
```diff
@@ -57,7 +57,9 @@ function request(::Type{StreamLayer}, io::IO, request::Request, body;
                     writebody(http, request, body)
                 catch e
                     write_error = e
-                    @warn("Error in @async writebody task",
+                    @info("Error in @async writebody task. " *
+                          "Server likely closed the connection unexpectedly. " *
+                          "Only an issue if unable to read the response and this error gets re-thrown. ",
                           exception=(write_error, catch_backtrace()))
                     close(io)
```
quinnj: This change looks good to me; thanks for the contribution! While we're here in this code, one thing I've noticed in a few applications is that this `close(io)` will often fail because it tries to send a closing ACK or whatever to the server. I've considered adding a quick `try; close(io); catch e; end` to just swallow those errors since we don't really care whether the server received the close ACK or not. But then again, just swallowing errors always feel a little sketchy. Have you seen this at all or run into issues here?

nicoleepp: I haven't run into that issue but may only be because it's getting caught and retried in the code that calls HTTP.

nicoleepp: I can still make that change if that's something you want to add though.

nicoleepp: @quinnj Added

omus: Would `isopen(io) && close(io)` not be sufficient?

nicoleepp:
Bump

quinnj:
Sorry for the delay here @nicoleepp! I applied @omus's suggestion in #436 and I'll merge it once we have a CI run there.

PR #433, opened 2019-07-22
add support for duplicate query param keys

acdupont:
* Store values for duplicate keys in array
* Add two tests for parsing duplicate query param keys
   * Tests pass with same behavior as prior to adding the tests (server.jl tests sometimes hang)

### DIFF
--- a/src/URIs.jl
+++ b/src/URIs.jl
@@ -273,9 +273,21 @@ uristring(u::URI) = uristring(u.scheme, u.userinfo, u.host, u.port,
 queryparams(uri::URI) = queryparams(uri.query)
 
 function queryparams(q::AbstractString)
-    Dict(unescapeuri(k) => unescapeuri(v)
-        for (k,v) in ([split(e, "=")..., ""][1:2]
-            for e in split(q, "&", keepempty=false)))
+    result = Dict()
+    for (k,v) in ([split(e, "=")..., ""][1:2] for e in split(q, "&", keepempty=false))
+        key = unescapeuri(k)
+        value = unescapeuri(v)
+        # RFC 3986 does not specify how duplicate keys are to be handled, store 
+        # duplicate keys in an array
+        if haskey(result, key)
+            isa(result[key], Vector) || (result[key] = [result[key]])
+            push!(result[key], value)
+        else
+            result[key] = value
+        end
+    end
+    
+    return result
 end
 
 # Validate known URI formats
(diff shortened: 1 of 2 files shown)


quinnj:
@acdupont, sorry for the slow review on this. I do think we should handle the case of parsing duplicate keys. I just want to make sure we have a quick discussion on the best solution here, options include:

* return a `Vector{Tuple{String, String}}`, which we support in several other cases in HTTP.jl via the `Pairs` module; this would probably be a bit more awkward for common use-cases, because you have to use the `Pairs.getkv` and `Pairs.setkv` methods instead of the normal `getindex`/`setindex!` for `Dict`. This would also be breaking.
* we could make the return type of `HTTP.queryparams` `Dict{String, Vector{String}}`, which would mean _always_ returning a `Vector{String}`, even if there was only a single value for each key (the most common case). The advantage of this case is consistency when using the results of `queryparams` (i.e. always iterating the results of each key-value pair), but would also be breaking.
* This approach, i.e. only returning a `Vector{String}` if there are duplicate keys. This is non-breaking in the non-duplicate case, which is good. But also introduces the inconsistency of not knowing whether I'll get just a `String` or a `Vector{String}` when accessing a key.

@oxinabox @omus @ararslan @randyzwitch , do you all have any opinions on the best way forward here? We could also manage the breaking situations with various plans: introduce a new function that handles the duplicate keys while leaving `queryparams` as-is w/ a deprecation. etc.

oxinabox:
We should not fear breaking.
Everyone has upper-bounded their compat now.
And this is why HTTP.jl has not yet tagged version 1.
I would remove that from the consideration.

I lean towards option two.
Consistency is important.
And one can just write:
```
x, = params["key"]
```

On the other hand, some frameworks deem this various kinds of illegal,
and everything handles it differently.
https://stackoverflow.com/questions/1746507/authoritative-position-of-duplicate-http-get-query-keys

On that basis, I don't hate 3.
And can always use dispatch there.
Take advantage of julia being a dynamic language

acdupont:
I lean towards option 2 as well. I wrote the code for this PR to have as little impact as possible, but returning the same type no matter how many parameters there are is less surprising. I would expect Option 3 to cause frustrating problems in calling code because places where one `queryparams` is expected, but multiple are given, would most likely result in the code breaking. I like the simpler API in accessing the results of option 2 vs option 1. Lastly, I like the synatx @oxinabox showed above in that dealing with a `Vector` when you expect one result is simple.

fredrikekre:
FWIW I kinda like 3 since that is how I specify duplicate keys when passing them to HTTP.jl. But maybe the other options also roundtrips?

clarkevans:
This is a hard one.  Perhaps option #4... you have two functions. One that always returns a vector (or should it be a tuple?), the other that always returns a string. The one that always returns a string raise an exception if there there is not exactly one value; if the parameter is missing or the parameter is specified twice, it becomes an error.  Alternatively, if you always return a tuple, someone could write the one that asserts a single value in user-land. I guess option #2 as `x, = params["key"]` supports that singularity assertion via user pattern.

fredrikekre:
This code has now been moved to [URIs.jl](https://github.com/JuliaWeb/URIs.jl). See also [URIs.jl#30](https://github.com/JuliaWeb/URIs.jl/issues/30).

PR #443, opened 2019-08-30
Interface for adding custom layers to the HTTP stack

### LINKED ISSUE #389: The Layers mechanism is not easily extensible
## Context

I am currently working on excising the AWS Signature v4 signing functionality into a separate package, [AWSAuth](https://github.com/JuliaCloud/AWSAuth.jl) (see #355). Part of the functionality here is a `Layer` subtype for AWS Sig v4 authentication, `AWS4AuthLayer`, which gets added to the stack when calling `HTTP.request` with `aws_authorization=true`. I encountered the issue described below, which unfortunately prompted https://github.com/JuliaCloud/AWSAuth.jl/pull/10.

## The issue

A package can define a `Layer` subtype with a `request` method that does what you want, but there are some serious complications. One example is that not every `request(::Layer, args...)` method takes the same arguments, so to know how to formulate the next call to `request` in your method, you need to know where in the stack your layer is, so that you know the expected signature for the next call.

Internally, HTTP gets around this by having `HTTP.stack` define the layers in a specific order that works. In that case, the AWS authentication layer is not the outermost layer, and to ensure that the AWSAuth package puts the layer in the same spot, it would need to completely reconstruct the stack by hand, and ensure that any preprocessing that's done in the user-facing `request` methods that get called before those which take `Layer`s is taken care of. That essentially amounts to a reimplementation of `HTTP.request`.

## Resolution

I think a good goal here would be to provide something more easily extensible by packages, but I'm not sure exactly what that should look like. To make the current design really work for everyone, packages should be able to insert layers into the stack without too much trouble.

Thoughts? Ideas?

mattBrzezinski:
Problem
---
`HTTP.stack` defines the layers in its stack in a specific order. If you currently want to insert custom layers at a specific point you must completely reconstruct the stack and ensure that all pre-processing is handled as well.

Solution
---
Create a custom layer (see `README.md` for an example), `insert()` and create your `request()`.

Notes
---
- Resolves #389 
- Allows for the removal of `AWS4AuthRequest.jl` from this package and move it into `AWSCore.jl`

### DIFF
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -1,6 +1,9 @@
 module HTTP
 
-export startwrite, startread, closewrite, closeread
+export startwrite, startread, closewrite, closeread, stack, insert, AWS4AuthLayer, 
+    BasicAuthLayer, CanonicalizeLayer, ConnectionPoolLayer, ContentTypeDetectionLayer,
+    DebugLayer, ExceptionLayer, MessageLayer, RedirectLayer, RetryLayer, StreamLayer, 
+    TimeoutLayer
 
 const DEBUG_LEVEL = Ref(0)
 
@@ -23,6 +26,62 @@ include("ConnectionPool.jl")
 include("Messages.jl")                 ;using .Messages
 include("cookies.jl")                  ;using .Cookies
 include("Streams.jl")                  ;using .Streams
+include("layers.jl")                   ;using .Layers
+
+"""
+## Request Execution Stack
+
+The Request Execution Stack is separated into composable layers.
+
+Each layer is defined by a nested type `Layer{Next}` where the `Next`
+parameter defines the next layer in the stack.
+The `request` method for each layer takes a `Layer{Next}` type as
+its first argument and dispatches the request to the next layer
+using `request(Next, ...)`.
+
+The example below defines three layers and three stacks each with
+a different combination of layers.
+
+
+```julia
+abstract type Layer end
+abstract type Layer1{Next <: Layer} <: Layer end
+abstract type Layer2{Next <: Layer} <: Layer end
+abstract type Layer3 <: Layer end
+
+request(::Type{Layer1{Next}}, data) where Next = "L1", request(Next, data)
+request(::Type{Layer2{Next}}, data) where Next = "L2", request(Next, data)
+request(::Type{Layer3}, data) = "L3", data
+
+const stack1 = Layer1{Layer2{Layer3}}
+const stack2 = Layer2{Layer1{Layer3}}
+const stack3 = Layer1{Layer3}
+```
+
+```julia
+julia> request(stack1, "foo")
+("L1", ("L2", ("L3", "foo")))
+
+julia> request(stack2, "bar")
+("L2", ("L1", ("L3", "bar")))
+
+julia> request(stack3, "boo")
+("L1", ("L3", "boo"))
+```
+
+This stack definition pattern gives the user flexibility in how layers are
+combined but still allows Julia to do whole-stack compile time optimisations.
+
+e.g. the `request(stack1, "foo")` call above is optimised down to a single
+function:
+```julia
+julia> code_typed(request, (Type{stack1}, String))[1].first
+CodeInfo(:(begin
+    return (Core.tuple)("L1", (Core.tuple)("L2", (Core.tuple)("L3", data)))
+end))
+```
+"""
+const nobody = UInt8[]
 
 """
 
@@ -297,22 +356,17 @@ HTTP.open("POST", "http://music.com/play") do io
 end
 ```
 """
-request(method::String, url::URI, headers::Headers, body; kw...)::Response =
-    request(HTTP.stack(;kw...), method, url, headers, body; kw...)
-#FIXME consider @nospecialize for `body` ? (other places? in ConnectionPool?)
-
-
-const nobody = UInt8[]
-
-function request(method, url, h=Header[], b=nobody;
-                 headers=h, body=b, query=nothing, kw...)::Response
-
+function request(stack::Type{<:Layer}, method, url, h=Header[], b=nobody;
+                headers=h, body=b, query=nothing, kw...)::Response
     uri = URI(url)
     if query !== nothing
         uri = merge(uri, query=query)
     end
-    return request(string(method), uri, mkheaders(headers), body; kw...)
+
+    return request(stack, string(method), uri, mkheaders(headers), body; kw...)
 end
+
+request(a...; kw...)::Response = return request(HTTP.stack(;kw...), a...; kw...)
 
 """
     HTTP.open(method, url, [,headers]) do io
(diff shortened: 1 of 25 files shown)


src/HTTP.jl
```diff
@@ -1,6 +1,9 @@
 module HTTP
 
-export startwrite, startread, closewrite, closeread
+export startwrite, startread, closewrite, closeread, stack, insert, AWS4AuthLayer, 
+    BasicAuthLayer, CanonicalizeLayer, ConnectionPoolLayer, ContentTypeDetectionLayer,
+    DebugLayer, ExceptionLayer, MessageLayer, RedirectLayer, RetryLayer, StreamLayer, 
+    TimeoutLayer
```
oxinabox: This is a lot of exports to add.
I think I am down with it though.
With this PR it is going to be much more comfortable building custom stacks of layers,
so makes sense to make the layer types part of the public API

src/HTTP.jl
```diff
@@ -297,20 +300,25 @@ HTTP.open("POST", "http://music.com/play") do io
...
-
+function request(method::String, url::URI, headers::Headers, body; kw...)::Response
+    request(HTTP.stack(;kw...), method, url, headers, body; kw...)
+end
 
 const nobody = UInt8[]
 
 function request(method, url, h=Header[], b=nobody;
-                 headers=h, body=b, query=nothing, kw...)::Response
+                 headers=h, body=b, query=nothing, custom_stack=nothing, kw...)::Response
```
oxinabox: Do we really need this as a kwarg?

Isn't using the `request(custom_stack, method, url, ...)`  method enough?

src/HTTP.jl
```diff
@@ -470,7 +478,8 @@ CodeInfo(:(begin
 end))
 ```
 """
-abstract type Layer end
+abstract type AbstractLayer end
+abstract type Layer{Next <: AbstractLayer} <: AbstractLayer end
```
oxinabox: Do we need the `AbstractLayer` type?
While it lets us add a type constraint on `Next`
it also makes it easier for someone to mistakenly inherit from `AbstractLayer`
rather than `Layer`,

src/HTTP.jl
```diff
@@ -626,12 +636,54 @@ function stack(;redirect=true,
...
+"""
+function head(::Type{T}) where T <: Layer
+    return T.name.wrapper
+end
+head(::Type{Union{}}) = Union{}
+
+"""
+    insert(stack::Type{<:Layer}, layer_before::Type{<:Layer}, custom_layer::Type{<:Layer})
+
+Insert your custom_layer infront of the layer_before
```
oxinabox:
```suggestion
Returns a new stack, with your `custom_layer` inserted infront of `layer_before`.

If there are multiple layers in the `stack` that match `layer_before`, then your new layer
is inserted below the top-most one.
```

src/exceptions.jl
```diff
@@ -0,0 +1,7 @@
+struct LayerNotFound <: Exception
```
oxinabox: What do you think of:

```suggestion
struct LayerNotFoundException <: Exception
```

test/client.jl
```diff
@@ -1,8 +1,24 @@
-using HTTP, Test, JSON
```
oxinabox: these should not have been deleted.
I believe they are here so that this test can be run via
```
julia --project=. test/client
```

test/client.jl
```diff
@@ -1,8 +1,24 @@
...
+   @testset "High-level Request" begin
+        custom_stack = insert(stack(), StreamLayer, TestLayer)
+        result = request("GET", "https://httpbin.org/ip", custom_stack=custom_stack)
+
+        @test status(result) == 200
+    end
+
+    @testset "Shorthand Request" begin
+        custom_stack = insert(stack(), StreamLayer, TestLayer)
+        result = HTTP.get("https://httpbin.org/ip", custom_stack=custom_stack)
```
oxinabox: On seeing this in use, I still do not like it.

much rather stack was a first argument,
even though it means defining more overloads.

I could be convinced though

test/insert_layers.jl
```diff
@@ -0,0 +1,40 @@
...
+        result = insert(stack(), Union{}, TestLayer)
+
+        @test expected == result
+    end
+
+    @testset "Insert - Non-existant layer" begin
+        @test_throws HTTP.LayerNotFound insert(stack(), AWS4AuthLayer, TestLayer)
+    end
+
+    @testset "Insert - Multiple Same layer" begin
```
oxinabox: Should this just be an error?

src/HTTP.jl
```diff
@@ -626,12 +636,54 @@ function stack(;redirect=true,
...
+function next(::Type{S}) where {T, S<:Layer{T}}
+    return T
+end
+
+"""
+    head(::Type{T}) where T <: Layer
+
+Get the name of the type
+"""
+function head(::Type{T}) where T <: Layer
```
oxinabox: Why is this called `head` ?
why not say `wrapper_type`

src/HTTP.jl
```diff
@@ -626,12 +636,54 @@ function stack(;redirect=true,
...
+Return the next layer in the stack
+"""
+function next(::Type{S}) where {T, S<:Layer{T}}
+    return T
+end
+
+"""
+    head(::Type{T}) where T <: Layer
+
+Get the name of the type
```
oxinabox: >  Get the name of the type

That is **not** what this does.

```
julia> struct Foo{X} end

julia> nameof(Foo{Int})
:Foo

julia> Foo{Int}.name
Foo

julia> Foo{Int}.name.wrapper
Foo

julia> Foo{Int}.name |> typeof
Core.TypeName

julia> Foo{Int}.name.wrapper |> typeof
UnionAll

julia> Foo{Int}.name.wrapper == Foo{T} where T
true
```

I think this only applies to parametric types?
That should be made clear in docstring also

src/HTTP.jl
```diff
@@ -485,6 +494,7 @@ include("ConnectionRequest.jl");        using .ConnectionRequest
 include("DebugRequest.jl");             using .DebugRequest
 include("StreamRequest.jl");            using .StreamRequest
 include("ContentTypeRequest.jl");       using .ContentTypeDetection
+include("exceptions.jl")
 
 """
```
oxinabox: After review is done, 
can we move this whole sectuion to a new file
like 
`strack.jl`  or something?

src/HTTP.jl
iamed2: That won't get the pass-through from `get`, `post`, etc. We could add methods for those though?

src/HTTP.jl
iamed2: We could export them from an artificial `HTTP.Layers` submodule?

src/HTTP.jl
iamed2: `top_layer` maybe, given that this is specialized on Layers. It doesn't need to be generic to all types.

src/HTTP.jl
iamed2: Do you find it

di`strack`ting?

😎

src/HTTP.jl
mattBrzezinski: What Eric mentioned is the alternative, I thought this was a more clean solution to give access to the "low" and "high" level calls.

I can go either way on this change.

test/client.jl
mattBrzezinski: I wasn't able to find concrete answers to this, and how people test in Julia.

Personally I've just been going into the the REPL and running all tests from there. The `runtests.jl` file also contains all the deps and usages there, which makes things nice and clean.

But as you point out, if you want to run individual test files, they won't work. Maybe containing all deps in their respective test files is the best way of doing this?

Current form:
```julia
test/runtests.jl
using HTTP, Test, JSON
include("test/my_test.jl")
```

Proposed form:
```julia
```

src/HTTP.jl
oxinabox: I think adding methods for all of them is better.
Potentialy via an `@eval` loop

test/client.jl
oxinabox: The Current form, for *this file* at-least is the proposed form.
Changing that is beyond the scope of this PR.

mattBrzezinski: I have a WIP branch currently for re-factoring testing in this package. I'll revert this change (as I think I think I accidentally included it during a rebase), and when that CR goes up we can address this there.

src/HTTP.jl
oxinabox: I think they are fine as is

oxinabox:
Cool, add the example to top_layer.
And move stuff to own file and I will merge tomorrow

src/HTTP.jl
```diff
@@ -23,6 +26,62 @@ include("ConnectionPool.jl")
...
+e.g. the `request(stack1, "foo")` call above is optimised down to a single
+function:
+```julia
+julia> code_typed(request, (Type{stack1}, String))[1].first
+CodeInfo(:(begin
+    return (Core.tuple)("L1", (Core.tuple)("L2", (Core.tuple)("L3", data)))
+end))
+```
+"""
+const nobody = UInt8[]
```
omus: Docstring should be attached to the `Layer` type in layers.jl

src/exceptions.jl
```diff
@@ -0,0 +1,7 @@
+struct LayerNotFoundException <: Exception
+	var::String
+end
+
+function Base.showerror(io::IO, e::LayerNotFoundException)
+	println(io, typeof(e), ": ", e.var)
+end
```
omus: Should use spaces instead of tabs. Seems to be just with new files

oxinabox: Arg, I forgot to check that.
Good catch

Issue #454, opened 2019-10-10
ConnectionPool isn't threadsafe


kcajf:
following on from https://github.com/JuliaLang/julia/issues/33429

It appears ConnectionPool isn't threadsafe. I'm trying to work out the minimal correct locking to make it work. In https://github.com/JuliaWeb/HTTP.jl/blob/master/src/ConnectionPool.jl#L371, do you think `poollock` and `poolcondition` can be combined into one `Threads.Condition`? I'm not entirely sure of their individual purposes, but from first glance it seems like they should be able to.

tk3369:
Following Jeff B's comment from the other issue, we could possibly change the `Condition` object to `Threads.Condition`.  

Looks like the `closeall` function needs some work... I would guess the `close` function call may throw exception so we could add a `tryclose` function with try/catch wrapped around it.  

https://github.com/JuliaWeb/HTTP.jl/blob/5089010c05f33ad680e2ed28ea18b4eef774051a/src/ConnectionPool.jl#L379-L388

Also, shall we consider using this locking style according to the recent [multi-threading blog](https://julialang.org/blog/2019/07/multithreading)?  So then we don't need a separate `ReentrantLock`...

```julia
lock(cond::Threads.Condition)
try
    while !ready
        wait(cond)
    end
finally
    unlock(cond)
end
```

rana:
Encountered a bug related to HTTP ConnectionPool too; likely the same. Glad you're looking into it. Looking forward to a fix. Thank you.

```
 Info: Error in @async writebody task. Server likely closed the connection unexpectedly. Only an issue if unable to read the response and this error gets re-thrown. 
│   exception =
│    concurrency violation detected
│    Stacktrace:
│     [1] error(::String) at ./error.jl:33
│     [2] concurrency_violation() at ./condition.jl:8
│     [3] assert_havelock at ./condition.jl:25 [inlined]
│     [4] assert_havelock at ./condition.jl:48 [inlined]
│     [5] assert_havelock at ./condition.jl:72 [inlined]
│     [6] notify(::Base.GenericCondition{Base.AlwaysLockedST}, ::Any, ::Bool, ::Bool) at ./condition.jl:126
│     [7] #notify#465 at ./condition.jl:124 [inlined]
│     [8] notify at ./condition.jl:124 [inlined] (repeats 2 times)
[...]

PR #455, opened 2019-10-10
Changed poolcondition to be threadsafe

kcajf:
My (possibly naive) attempt to fix https://github.com/JuliaWeb/HTTP.jl/issues/454

It fixes my usecase at least, but might be incomplete, or overkill


kcajf:
Ok, need to have some julia-version-dependent logic. Any thoughts on how to do this cleanly in one place?

quinnj:
Implemented in https://github.com/JuliaWeb/HTTP.jl/pull/479

PR #457, opened 2019-10-12
Allow insertion of a layer after StreamLayer

christopher-dG:
Currently, `request(::Type{StreamLayer}, ...)` returns a response, so even if there are layers after it, they won't be used. 

This PR makes `StreamLayer` behave in the same way as the other layers (calling `request(Next, ...)`, and `Union{}` becomes the last layer which just returns the `Response` it's given.

Now you can add a layer after `StreamLayer`, which can be used for stuff like caching.
Other than that new capability, no behaviour is changed.


christopher-dG:
Not sure what is up with the one Mac build, the error is:
```
could not open file /Users/travis/build/JuliaWeb/HTTP.jl/test/aws4.jl
```

.travis.yml
```diff
@@ -13,7 +13,7 @@ arch:
 
 julia:
   - 1.0
-  - 1.3
```
quinnj, line 16: unintended change?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "HTTP"
 uuid = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 authors = ["Jacob Quinn", "Sam O'Connor", "contributors: https://github.com/JuliaWeb/HTTP.jl/graphs/contributors"]
-version = "0.8.14"
```
quinnj, line 4: I'd rather bump this to 0.8.15 since there's nothing breaking here (right?)

.travis.yml
christopher-dG: I replaced it with `1`, which expands to the latest 1.x. So when new Julia versions are released, we won't need to update this file. Unless we want to test 1.3 specifically, but I don't see much reason to.

Project.toml
christopher-dG: Ah, yeah I forgot that for 0.x, the minor version is basically seen as major.

christopher-dG:
Should be good to go unless you want the .travis.yml change reverted.

.travis.yml
quinnj: Oh, I didn't know that! Useful!

PR #464, opened 2019-11-05
Don't sidestep main entry method for HTTP.request, fixes #463.

### LINKED ISSUE #463: HTTP.request sidesteps a bit too much ?
This new method: https://github.com/JuliaWeb/HTTP.jl/blob/7bcb6f554f116eb075a3286d641be8fbb2b400bf/src/HTTP.jl#L316 sidesteps a bit too much. I assume it is meant to call the method just above: https://github.com/JuliaWeb/HTTP.jl/blob/7bcb6f554f116eb075a3286d641be8fbb2b400bf/src/HTTP.jl#L306-L314 in order to correctly deal with `headers` kwargs etc. But that does not happen, because we dispatch to https://github.com/JuliaWeb/HTTP.jl/blob/7bcb6f554f116eb075a3286d641be8fbb2b400bf/src/RedirectRequest.jl#L18 which completely ignores `headers` (it is included in `kwargs...`).

Here is an example, where `headers` is ignored:
```
julia> using HTTP

julia> uri = HTTP.URI("https://api.github.com");

julia> headers=Dict("User-Agent" => "HTTP.jl");

julia> HTTP.request("GET", uri; headers=headers)
ERROR: HTTP.ExceptionRequest.StatusError(403, "GET", "/", HTTP.Messages.Response:
"""
HTTP/1.0 403 Forbidden
Cache-Control: no-cache
Connection: close
Content-Type: text/html

Request forbidden by administrative rules. Please make sure your request has a User-Agent header (http://developer.github.com/v3/#user-agent-required). Check https://developer.github.com for other possible causes.
""")
```

fredrikekre:


fredrikekre:
cc @mattBrzezinski

fredrikekre:
Ok, of course it was not that simple :)

fredrikekre:
I can't get dispatch to work correctly here. It feels impossible to support both the `HTTP.request(::Layer, ....)` and `HTTP.request(method::String, ...)` entry points while not accidentally dispatching to some internal `request` function. It would be trivial if we could separate these into two functions, and let `HTTP.request` be the public one, and all the internal ones be `HTTP._request` or similar.

fredrikekre:
FWIW, here is the test-suite I built up:
```
@testset "Public entry point of HTTP.request and friends (e.g. issue #463)" begin
    headers = Dict("User-Agent" => "HTTP.jl")
    query = Dict("hello" => "world")
    body = UInt8[1, 2, 3]
    uri = HTTP.URI("https://httpbin.org/anything")
    stack = HTTP.stack()
    function test(r, m)
        @test r.status == 200
        d = JSON.parse(IOBuffer(HTTP.payload(r)))
        @test d["headers"]["User-Agent"] == "HTTP.jl"
        @test d["data"] == "\x01\x02\x03"
        @test endswith(d["url"], "?hello=world")
        @test d["method"] == m
    end
    # HTTP.request
    test(HTTP.request("GET", uri; headers=headers, body=body, query=query), "GET")
    test(HTTP.request("GET", uri, headers; body=body, query=query), "GET")
    test(HTTP.request("GET", uri, headers, body; query=query), "GET")
    test(HTTP.request(stack, "GET", uri; headers=headers, body=body, query=query), "GET")
    test(HTTP.request(stack, "GET", uri, headers; body=body, query=query), "GET")
    test(HTTP.request(stack, "GET", uri, headers, body; query=query), "GET")
    # HTTP.get
    test(HTTP.get(uri; headers=headers, body=body, query=query), "GET")
    test(HTTP.get(uri, headers; body=body, query=query), "GET")
    test(HTTP.get(uri, headers, body; query=query), "GET")
    # HTTP.put
    test(HTTP.put(uri; headers=headers, body=body, query=query), "PUT")
    test(HTTP.put(uri, headers; body=body, query=query), "PUT")
    test(HTTP.put(uri, headers, body; query=query), "PUT")
    # HTTP.post
    test(HTTP.post(uri; headers=headers, body=body, query=query), "POST")
    test(HTTP.post(uri, headers; body=body, query=query), "POST")
    test(HTTP.post(uri, headers, body; query=query), "POST")
    # HTTP.patch
    test(HTTP.patch(uri; headers=headers, body=body, query=query), "PATCH")
    test(HTTP.patch(uri, headers; body=body, query=query), "PATCH")
    test(HTTP.patch(uri, headers, body; query=query), "PATCH")
    # HTTP.delete
    test(HTTP.delete(uri; headers=headers, body=body, query=query), "DELETE")
    test(HTTP.delete(uri, headers; body=body, query=query), "DELETE")
    test(HTTP.delete(uri, headers, body; query=query), "DELETE")
end
```

fredrikekre:
Another option would be to be more strict with types in internal functions.

mattBrzezinski:
> I can't get dispatch to work correctly here. It feels impossible to support both the `HTTP.request(::Layer, ....)` and `HTTP.request(method::String, ...)` entry points while not accidentally dispatching to some internal `request` function. It would be trivial if we could separate these into two functions, and let `HTTP.request` be the public one, and all the internal ones be `HTTP._request` or similar.

I think this the best option, I remember struggling with getting dispatch to take the correct pathing when introducing the inserting layer changes. I did not realize that the headers were being missed in the requests when making my changes.

I haven't looked at this package for a while, but to resolve #464 we would just need to do the following (of the proposed solution):

1. Change all `Layer` functions from `request(::Type{Layer{Next}}...)` to `_request(::Type{Layer{Next}}...)`
2. Change the `src/HTTP.jl` `request(..)` to return `_request(...)`
3. Add the unit tests above for better coverage

fredrikekre:
Another option would be to introduce something like
```
struct Stack
   stack
end
```
and then, in order to call `request` with a user-defined stack you do something like:
```
stack = HTTP.stack()::Stack
HTTP.request(stack::Stack, args...) = HTTP.request(stack.stack, args...)
```
to get around that issue

mattBrzezinski:
> Another option would be to introduce something like
> 
> ```
> struct Stack
>    stack
> end
> ```
> 
> and then, in order to call `request` with a user-defined stack you do something like:
> 
> ```
> stack = HTTP.stack()::Stack
> HTTP.request(stack::Stack, args...) = HTTP.request(stack.stack, args...)
> ```
> 
> to get around that issue

I'm not sure if introducing functionality is the best approach for solving this, I think it'll just add more complexity down the road.

I should have some time to spare this week to play around and try again to get the dispatch to work properly. I might have some old branches locally sitting around with some other variants of this code (fingers crossed).

PR #465, opened 2019-11-05
Refactor Unit Testing

mattBrzezinski:
## Overview

This CR refactors the current state of testing for `HTTP.jl`, pushing towards using `@testsets` for like-wise tests. It's quite a lengthy CR so it has been broken apart into multiple commits each one related to the refactor of a given component to make reviewing easier.

There were a few test files never run; `body.jl`, `mwe.jl`, and `uri_benchmark.jl`. I have included them in the `run_tests.jl` so they are not forgotten, however in their current state they are broken (see below for failures). I'll log issues for each of these, but their `@testsets` are marked with `@test_skip` to keep the CI happy.

## References
- `test/body.jl` failure -- #466 
- `test/mwe.jl` failure -- #467
- `test/uri_benchmark.jl` failure -- #468


test/body.jl
```diff
@@ -1,53 +1,58 @@
...
-
-    io = Base.BufferStream()
-    @async begin
-        write(io, "Hello")
-        sleep(0.1)
-        write(io, "!")
-        sleep(0.1)
+@test_skip @testset "HTTP.Bodies" begin
+    @testset "Body" begin
+        @test String(take!(Body("Hello!"))) == "Hello!"
```
quinnj: I think `Body` can be changed to `HTTP.bodybytes`; though these tests are pretty old and crufty, so I'm not sure if they're still useful (i.e. we might just remove them). But if you change to `HTTP.bodybytes`, and they work, we can keep them.

test/cookies.jl
```diff
@@ -8,123 +8,129 @@
...
-        (Dict("Cookie" => "Cookie-1=v\$1; c2=v2"), "", [HTTP.Cookie("Cookie-1", "v\$1"), HTTP.Cookie("c2", "v2")]),
-        (Dict("Cookie" => "Cookie-1=v\$1; c2=v2"), "c2", [HTTP.Cookie("c2", "v2")]),
-        (Dict("Cookie" => "Cookie-1=\"v\$1\"; c2=\"v2\""), "", [HTTP.Cookie("Cookie-1", "v\$1"), HTTP.Cookie("c2", "v2")]),
-    ]
-    for (h, filter, cookies) in testcookies
-        @test HTTP.Cookies.readcookies(h, filter) == cookies
-    end
-end
-end; # @testset "Cookies"
+		(HTTP.Cookie("cookie-1", "v\$1"), "cookie-1=v\$1"),
```
quinnj: why the indentation change here?

test/messages.jl
```diff
@@ -17,167 +15,147 @@ using HTTP.MessageRequest: unknown_length
...
+        @test bodylength([SubString("Hello", 1,5), " ", SubString("World!",1,5)]) == 11
+        @test bodylength([UInt8[1,2,3], UInt8[4,5,6]]) == 6
+        @test bodylength([UInt8[1,2,3], view(UInt8[4,5,6],1:2)]) == 5
+        @test bodylength([view(UInt8[1,2,3],1:2), view(UInt8[4,5,6],1:2)]) == 4
+        @test bodylength(IOBuffer("foo")) == 3
+        @test bodylength([IOBuffer("foo"), IOBuffer("bar")]) == 6
+    end
 
+    @testset "Body Bytes" begin
+        @test bodybytes(7) == UInt8[]
```
quinnj: Looks like we already have test for `bodybytes`, so I say we just remove the `Body.jl` file.

test/uri_benchmark.jl
```diff
@@ -1,45 +1,45 @@
+using HTTP
 using Unitful
 
-using HTTP
+include("resources/urlparser.jl")
```
quinnj: this file wasn't really ever meant to be tests, so maybe just remove it, or move it to a top level `benchmark/` directory

quinnj:
Another change that would be good is to switch from using AppVeyor to doing windows testing on travis, see the CSV.jl travis file for reference: https://github.com/JuliaData/CSV.jl/blob/master/.travis.yml.

test/cookies.jl
mattBrzezinski: Looks like this happened accidentally, also this file is using tabs. Fixing this now.

test/body.jl
mattBrzezinski: Looking more into it, `bodybytes` are being tested in `test/messages.jl`. This file can be removed.

mattBrzezinski:
@quinnj TravisCI is passing now, can you remove the AppVeyor configuration for this project? I assume I don't have the perms to do that. I removed the `appveyor.yml` config file in this CR.

quinnj:
Deleted. Thanks!

PR #469, opened 2019-11-06
Separate HTTP.request from Layers.request

fredrikekre:

PR #470, opened 2019-11-06
Rework HTTP.request entrypoints to fix dispatch.

fredrikekre:


fredrikekre:
This is sufficient to fix almost all cases, except this signature (commented out in the tests):
```
HTTP.request(::Type{<:Layer}, ::String, ::URI; kwargs...)
```
In order to fix that I think something like  #469 would make life so much easier, but I could not come up with a way to make that non-breaking so that will have to wait for the next breaking release.

I also took the liberty to bump the version and would be nice with a new release :)

mattBrzezinski:
Looks like [this job](https://travis-ci.org/JuliaWeb/HTTP.jl/jobs/608157208?utm_medium=notification&utm_source=github_status) timed out and needs a kick, I don't perms to do it.

fredrikekre:
Retriggerd by rebase.

mattBrzezinski:
Just needs a project patch version bump and it's good to go!

oxinabox:
Thanks @mattBrzezinski 

Looks good to me also.
(i'll bump version for you so I can tag it tonight)
I think we do really want something like #469 to tidy it up and reduce code duplication.
But til then this will do

oxinabox:
actuually it doesn't need bumping,
as it looks like 0.8.7  forgot to be registered

mattBrzezinski:
That works too then!

PR #473, opened 2019-11-15
Set Content-Type correctly with body::Form

fredrikekre:


test/client.jl
```diff
@@ -242,4 +242,22 @@ end
         test(HTTP.delete(uri, headers; body=body, query=query), "DELETE")
         test(HTTP.delete(uri, headers, body; query=query), "DELETE")
     end
+    # Setting of Content-Type with body::Form
```
mattBrzezinski: Can this be in its own `@testset`?

test/client.jl
```diff
@@ -242,4 +242,22 @@ end
...
+    body = HTTP.Form(Dict())
+    uri = "https://httpbin.org/post"
+    for r in (HTTP.request("POST", uri, headers, body), HTTP.post(uri, headers, body))
+        @test r.status == 200
+        json = JSON.parse(IOBuffer(HTTP.payload(r)))
+        @test startswith(json["headers"]["Content-Type"], "multipart/form-data; boundary=")
+    end
+    # Deprecation of HTTP.post without header for body::Form
```
mattBrzezinski: This should probably be moved into its own `@testset` as well.

test/client.jl
```diff
@@ -242,4 +242,22 @@ end
...
+    proj = normpath(joinpath(pathof(HTTP), "..", "..", "Project.toml"))
+    vers = VersionNumber(match(r"^version\s*=\s*\"(.*?)\"$"m, read(proj, String)).captures[1])
```
mattBrzezinski: Can there be a comment explaining the `regex`? Makes it easier to know what you're matching at a quick glance.

src/multipart.jl
```diff
@@ -125,9 +125,6 @@ end
 content_type(f::Form) = "Content-Type" =>
                         "multipart/form-data; boundary=$(f.boundary)"
 
+# To be deprecated in HTTP@0.9
```
mattBrzezinski: Can we log an issue for this so it's not forgotten?

test/client.jl
fredrikekre: I guess.

src/multipart.jl
fredrikekre: Thats why I conditioned the tests on the version; when the version is bumped CI will fail and then someone has to enable this deprecation.

mattBrzezinski:
I don't have approve permissions. But looks good to me, ship it!

ararslan:
Travis is consistently timing out running the server tests, but the tests here all run before that and passed, so I think we're good.

PR #477, opened 2019-12-03
Change Condition to Threads.Condition in Julia 1.3 to make connection…

### LINKED ISSUE #454: ConnectionPool isn't threadsafe

quinnj:
… pool operations thread-safe. Fixes #454

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,14 @@ os:
   - windows
 julia:
   - 1.0
+  - 1.3
   - nightly
+matrix:
+  exclude:
+    - os: osx
+      arch: x86
+  allow_failures:
+  - julia: nightly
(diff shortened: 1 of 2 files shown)


quinnj:
Implemented in https://github.com/JuliaWeb/HTTP.jl/pull/479

Issue #453, opened 2019-10-09
listenloop() too aggressive about closing server


kwdye:
Julia 1.4
HTTP.jl 0.8.6
MbedTLS.jl 0.7.0

It looks like the HTTP server is very aggressive about shutting itself down when anything goes wrong with a new client connection.  I have been trying to use the HTTP server with a self-signed certificate and when a client connects and fails to validate the certificate, it causes the server to throw an error and exit completely.  This obviously isn't a great behavior for a server.

To reproduce run:
```
using MbedTLS
using HTTP

myRouter = HTTP.Router()
HTTP.@register(myRouter, "GET", "/foo", r->HTTP.Response(200, "{OK}"))
cfg=MbedTLS.SSLConfig("/Users/kwd/mycrt.crt","/Users/kwd/mykey.key")
MbedTLS.ssl_conf_renegotiation!(cfg, MbedTLS.MBEDTLS_SSL_RENEGOTIATION_ENABLED)
HTTP.serve(myRouter, "127.0.0.1", 8086, sslconfig=cfg)
```

and then from another terminal run

```curl https://127.0.0.1:8086/foo```

it will run fine if you tell curl not to verify the certificate

```curl https://127.0.0.1:8086/foo -k```

the error I get in julia is

```
ERROR: LoadError: MbedTLS error code -30592: SSL - A fatal alert message was received from our peer (You may need to enable ssl_conf_renegotiation!. See https://github.com/JuliaWeb/HTTP.jl/issues/342#issuecomment-432921180)
```
```
Stacktrace:
 [1] handshake(::MbedTLS.SSLContext) at /Users/kwd/.julia/packages/MbedTLS/a1JFn/src/ssl.jl:83
 [2] handshake! at /Users/kwd/.julia/packages/MbedTLS/a1JFn/src/MbedTLS.jl:141 [inlined]
 [3] getsslcontext(::Sockets.TCPSocket, ::SSLConfig) at /Users/kwd/.julia/packages/HTTP/hJSsm/src/Servers.jl:87
 [4] listenloop(::Function, ::HTTP.Servers.Server{SSLConfig,Sockets.TCPServer}, ::HTTP.Servers.var"#2#5"{Nothing,HTTP.Servers.var"#4#7"}, ::Base.RefValue{Int64}, ::Int64, ::Int64, ::Bool) at /Users/kwd/.julia/packages/HTTP/hJSsm/src/Servers.jl:81
[...]

PR #479, opened 2019-12-05
Make HTTP.jl threadsafe by making all global states per thread. Fixes…

quinnj:
… #453

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -258,7 +258,7 @@ function IOExtras.closewrite(t::Transaction)
     t.c.writebusy = false
     t.c.writecount += 1                           ;@debug 2 "🗣  Write done: $t"
     notify(t.c.writedone)
-    notify(poolcondition)
+    notify(poolcondition[Threads.threadid()])
 
     @ensure !iswritable(t)
     return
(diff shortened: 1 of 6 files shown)


vtjnash:
Wrong issue numbers?

src/URIs.jl
```diff
@@ -419,8 +419,9 @@ function absuri(uri::URI, context::URI)
 end
 
 function __init__()
-    Base.compile(uri_reference_regex.re)
-    initialize!(uri_reference_regex)
+    Threads.resize_nthreads!(uri_reference_regex)
+    foreach(x -> Base.compile(x.re), uri_reference_regex)
+    foreach(initialize!, uri_reference_regex)
```
vtjnash: Regex should generally not be copied per-thread

src/ConnectionPool.jl
```diff
@@ -686,27 +684,34 @@ function tcpstatus(c::Connection)
...
+        println(io, "</table>")
     end
-    println(io, "</table>")
-    unlock(poollock)
+end
+
+function __init__()
+    Threads.resize_nthreads!(pool)
+    Threads.resize_nthreads!(poollock)
+    Threads.resize_nthreads!(poolcondition)
```
vtjnash: These arrays are invalid to pass to `resize_nthreads!`

quinnj: It would be infinitely more helpful to explain why or to provide an alternative solution. Nothing from the `resize_nthreads!` docs seems to indicate that anything would be wrong with this approach.

PR #482, opened 2019-12-07
Set a user agent by default. Fixes #450

### LINKED ISSUE #450: HTTP.jl should send a User-Agent header by default
HTTP RFC says that clients "SHOULD" send a user agent by default. I would propose
```
User-Agent: HTTP.jl/$version
```
where `version` is the version number of the HTTP library. GitHub, for example, refuses to serve requests for URLs when the request doesn't have a user agent. Doesn't matter what it is, but it has to be set.

quinnj:


src/MessageRequest.jl
```diff
@@ -51,15 +51,16 @@ function request(::Type{MessageLayer{Next}},
     return request(Next, url, req, body; iofunction=iofunction, kw...)
 end
 
-const USER_AGENT = Ref{String}()
+const USER_AGENT = Ref{Union{String, Nothing}}("HTTP.jl/$VERSION")
```
fredrikekre: Should this not be the version of `HTTP.jl` instead of the Julia version?

quinnj: Why?

mattBrzezinski: I think that setting this to the `HTTP.jl` version rather than the Julia version makes more sense. If a user ever has an issue down the line, it makes more sense to tie the default user-agent to the package rather than the version of Julia.

Tying the default user-agent to the package also seems to be more in-line with the spirit of what the identification represents. Although I can understand the argument 

> https://en.wikipedia.org/wiki/User_agent#User_agent_identification

fredrikekre: `curl`, `wget` and `python-requests` uses its own version rather than the compiler too.

Issue #472, opened 2019-11-12
Using HTTP.get to https over proxy results in HTTP/1.1 400 Bad Request


oheil:
Windows 10
Julia 1.2
HTTP.jl v0.8.7

using following HTTP.get 
```
HTTP.get("https://github.com/JuliaWeb"; proxy="http://our-proxy-host:3128")
```
results in error:
```
ERROR: HTTP.ExceptionRequest.StatusError(400, "CONNECT", "github.com:", HTTP.Messages.Response:
"""
HTTP/1.1 400 Bad Request
Cache-Control: no-cache
Pragma: no-cache
Content-Type: text/html; charset=utf-8
Proxy-Connection: close
Connection: close
Content-Length: 691

""")
Stacktrace:
 [1] #request#1(::Base.Iterators.Pairs{Symbol,Union{Nothing, String},Tuple{Symbol,Symbol},NamedTuple{(:iofunction, :proxy),Tuple{Nothing,String}}}, ::typeof(HTTP.request), ::Type{ExceptionLayer{ConnectionPoolLayer{StreamLayer{Union{}}}}}, ::HTTP.URIs.URI, ::Vararg{Any,N} where N) at C:\Users\oheil\.julia\packages\HTTP\6Smhf\src\ExceptionRequest.jl:22
 [2] (::getfield(HTTP, Symbol("#kw##request")))(::NamedTuple{(:iofunction, :proxy),Tuple{Nothing,String}}, ::typeof(HTTP.request), ::Type{ExceptionLayer{ConnectionPoolLayer{StreamLayer{Union{}}}}}, ::HTTP.URIs.URI, ::HTTP.Messages.Request, ::Array{UInt8,1}) at .\none:0
[...]

PR #483, opened 2019-12-07
proxy: add case when target port is not specified

oheil:
PR for issue 
https://github.com/JuliaWeb/HTTP.jl/issues/472

### DIFF
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -114,7 +114,7 @@ function tunnel_request(Next, io, target_url, req, body; kw...)
 end
 
 function connect_tunnel(io, target_url, req)
-    if isempty(target_url.port)
+    if isempty( target_url.port )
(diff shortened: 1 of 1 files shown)


quinnj:
I think this was fixed in https://github.com/JuliaWeb/HTTP.jl/pull/507

oheil:
target_url.port can still be empty for non https proxy requests, which results in URLs like:
http://proxy.com:/
See
src/ConnectionRequest.jl :
```
function connect_tunnel(io, target_url, req)
    target = "$(URIs.hoststring(target_url.host)):$(target_url.port)"
```
So, my guess is, https://github.com/JuliaWeb/HTTP.jl/pull/507 doesn't fix all problems.
Furthermore, in some setups, there may be a proxy with protocol "https" on a different port than 443, which is overwritten with https://github.com/JuliaWeb/HTTP.jl/pull/507

PR #486, opened 2019-12-12
Removed AWS4AuthRequest Layer | Deprecate HTTP.Post without headers

mattBrzezinski:
The `AWS4AuthLayer` was introduced into this package because it was required for signing AWS signatures in HTTP. With #443 we are now able to insert custom layers into the HTTP stack, this means we can now move the `AWS4AuthLayer` out of this package.

I have already moved the `AWS4AuthLayer` into `AWSCore.jl` in [this commit](https://github.com/JuliaCloud/AWSCore.jl/commit/19e20b6#diff-3622fe17ed3d67519bc2425dc231ade3R1).  Going forward if you want to use the `AWS4Auth` Layer in your stack would be to do something like:

```
stack = stack()
insert(stack(), MessageLayer, AWS4AuthLayer)

request(...)
```

---

As per #473 this CR will also deprecate the usage of `HTTP.Post` without headers for body::Form.


test/insert_layers.jl
```diff
@@ -26,7 +26,7 @@ using ..TestRequest
     end
 
     @testset "Insert - Non-existant layer" begin
```
omus:
```suggestion
    @testset "Insert - Non-existent layer" begin
```

omus:
Appears to be additional changes to make to appease the tests

mattBrzezinski:
Forgot that we needed to deprecate `HTTP.post()` without headers in the next minor version. Added the changes to it in  
> https://github.com/JuliaWeb/HTTP.jl/pull/486/commits/1250ad995d55cb5211c1bb739807dc21daf44160

fredrikekre:
Before releasing a new breaking version it would be good to consider  #469.

mattBrzezinski:
That's a good idea, might as well get in what we can with this release.

quinnj:
This PR [has been noted](https://github.com/JuliaWeb/HTTP.jl/issues/786) as desirable before we do a 1.0 release; @mattBrzezinski, any chance you could update it? Even if just the src code and I can help with any test-related changes?

mattBrzezinski:
> This PR [has been noted](https://github.com/JuliaWeb/HTTP.jl/issues/786) as desirable before we do a 1.0 release; @mattBrzezinski, any chance you could update it? Even if just the src code and I can help with any test-related changes?

Yes. I can have this updated probably by the end of the year some time? Just wrapping up other stuff currently.

quinnj:
The removal of AWSLayer is now included in #789; I'm not familiar with the `post` deprecation that was included in this PR, so we might want to do that in a new commit if it's still relevant.

Issue #487, opened 2019-12-16
concurrency violation detected on HTTP.get() on Julia 1.3 with JULIA_NUM_THREADS > 1


quatrix:
Hi,

Running the following code with latest version of HTTP from git throws a `concurrency violation detected` exception when running with JULIA_NUM_THREADS>1

```julia
using HTTP

import Base.Threads.@spawn

res = Threads.@spawn begin
    HTTP.get("https://google.com")
end

fetch(res)
```

```bash
JULIA_NUM_THREADS=2 julia simple_get.jl
```

```julia
┌ Info: Error in @async writebody task. Server likely closed the connection unexpectedly. Only an issue if unable to read the response and this error gets re-thrown.
│   exception =
│    concurrency violation detected
│    Stacktrace:
│     [1] error(::String) at ./error.jl:33
│     [2] concurrency_violation() at ./condition.jl:8
│     [3] assert_havelock at ./condition.jl:25 [inlined]
│     [4] assert_havelock at ./condition.jl:48 [inlined]
│     [5] assert_havelock at ./condition.jl:72 [inlined]
│     [6] notify(::Base.GenericCondition{Base.AlwaysLockedST}, ::Any, ::Bool, ::Bool) at ./condition.jl:126
│     [7] #notify#465 at ./condition.jl:124 [inlined]
│     [8] notify at ./condition.jl:124 [inlined] (repeats 2 times)
│     [9] closewrite(::HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}) at /Users/quatrix/.julia/packages/HTTP/jHoHG/src/ConnectionPool.jl:261
│     [10] writebody(::HTTP.Streams.Stream{HTTP.Messages.Response,HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}}, ::HTTP.Messages.Request, ::Array{UInt8,1}) at /Users/quatrix/.julia/packages/HTTP/jHoHG/src/Streams.jl:120
└     [11] (::HTTP.StreamRequest.var"#2#3"{HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext},HTTP.Messages.Request,Array{UInt8,1},HTTP.Streams.Stream{HTTP.Messages.Response,HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}}})() at ./task.jl:333
```
I found the following resolved tickets #454 #477 #479 but I'm still getting the exception when using latest HTTP and MbedTLS from git.

[...]

PR #488, opened 2019-12-17
In Julia 1.3 lock condition before notify() or wait() on it

quatrix:


quatrix:
#487

DilumAluthge:
Instead of doing
```
lock(cond)
... stuff ...
unlock(cond)
```

Perhaps it would be better to do:
```
lock(cond) do
    ... stuff...
end
```

quatrix:
@DilumAluthge Thanks! changed to lock() do

quinnj:
@quatrix, can you show some tests or how this solves #487? I know it's tricky to test threaded scenarios on travis, so I'm not necessarily saying we need more tests, but at least something a few of us can try locally before/after these changes to get a feel for what's getting solved.

quatrix:
@quinnj, actually you can try the example in #487 it works with this PR and throws an exception with the master when running with ```JULIA_NUM_THREADS=2```

Maybe it's an overkill but I've added that as part of the ```client.jl``` tests, calling GET/POST/PUT/etc inside a ```Threads.@spawn``` and added to ```.travis.yml``` to run the tests with ```JULIA_NUM_THREADS=1``` and ```JULIA_NUM_THREADS=2```

src/ConnectionPool.jl
```diff
@@ -240,7 +262,7 @@ function IOExtras.startwrite(t::Transaction)
     @require !iswritable(t)                     ;t.c.writecount != t.sequence &&
                                                    @debug 1 "⏳  Wait write: $t"
     while t.c.writecount != t.sequence
-        wait(t.c.writedone)
+        _wait(t.c.writedone)
     end                                           ;@debug 2 "👁  Start write:$t"
     t.c.writebusy = true
```
NHDaly: You almost certainly want to be holding the lock around this entire section, not just within the `wait` function itself. Note this example on wikipedia about how to use a Monitor (a combo Condition Variable + Mutex, which is what Julia's `Condition()` type provides):
https://en.wikipedia.org/wiki/Monitor_(synchronization)#Monitor_usage
```
acquire(m); // Acquire this monitor's lock.
while (!p) { // While the condition/predicate/assertion that we are waiting for is not true...
	wait(m, cv); // Wait on this monitor's lock and condition variable.
}
// ... Critical section of code goes here ...
signal(cv2); -- OR -- notifyAll(cv2); // cv2 might be the same as cv or different.
release(m); // Release this monitor's lock.
```

You'll need to lock during accesses to `writebusy`, `writecount`, `sequence`, etc, here, lest they be updated simultaneously from multiple threads.

It might be that this entire function, actually, should be within the lock.

src/ConnectionPool.jl
```diff
@@ -257,8 +279,8 @@ function IOExtras.closewrite(t::Transaction)
 
     t.c.writebusy = false
     t.c.writecount += 1                           ;@debug 2 "🗣  Write done: $t"
-    notify(t.c.writedone)
-    notify(poolcondition[Threads.threadid()])
+    _notify(t.c.writedone)
+    _notify(poolcondition[Threads.threadid()])
 
     @ensure !iswritable(t)
     return
```
NHDaly: Same here; it looks to me like this whole section should be `lock`ed, not just within the `_notify` function itself.

src/ConnectionPool.jl
```diff
@@ -274,7 +296,7 @@ function IOExtras.startread(t::Transaction)
                                                    @debug 1 "⏳  Wait read:  $t"
     t.c.timestamp = time()
     while t.c.readcount != t.sequence
-        wait(t.c.readdone)
+        _wait(t.c.readdone)
```
NHDaly: and here, etc

quinnj:
superseded by on-going work in https://github.com/JuliaWeb/HTTP.jl/pull/509

PR #494, opened 2020-01-08
Closing connection still serves one connection

SimonDanisch:
After closing e.g. as in:
```Julia
    server = Sockets.listen(Sockets.InetAddr(parse(IPAddr, "127.0.0.1"), port))
    tsk = @async HTTP.listen(handler, "127.0.0.1", port; server=server)
    sleep(3.0)
    r = testget("http://127.0.0.1:$port")
    @test occursin(r"HTTP/1.1 200 OK", r)
    close(server)
    connection_refused = try
        HTTP.get("http://127.0.0.1:$port", readtimeout=3, retries=1)
        false
    catch e
        e isa HTTP.IOExtras.IOError && e.e == Base.IOError("connect: connection refused (ECONNREFUSED)", -4078)
    end
    @test connection_refused
```
This `HTTP.get("http://127.0.0.1:$port", readtimeout=3, retries=1)` get will still go through fine.
This happens, because startread is still blocking on the open connection to the client:
https://github.com/JuliaWeb/HTTP.jl/blob/master/src/Servers.jl#L340
And will happily continue serving another request, even though the server was closed.

This is bad, since serving a request in closed state is undefined behavior.

I'm not sure if my fix is any good - it looks a bit fishy, so I'd be very happy if someone has a better solution!
I'll add some tests, but right now the test suite hangs for me locally


SimonDanisch:
bump... @quinnj, any ideas about this? I'm not 100% sure, but it seems like this behavior, keeping connections, waiting on them and responding to them after closing, makes my CI unreliable...

quinnj:
Fix in https://github.com/JuliaWeb/HTTP.jl/pull/513

PR #495, opened 2020-01-10
WIP/RFC: added parser for multipart form with updates

pixel27:
This is basically the https://github.com/JuliaWeb/HTTP.jl/pull/427 pull request with changes asked for by @c42f here https://github.com/JuliaWeb/HTTP.jl/pull/427#pullrequestreview-275747560.

The biggest change where the processing of the arguments on the "Content-Disposition" command according to https://tools.ietf.org/html/rfc2183.  Basically to parse that line correctly I created a content_disposition_tokenize() method that would take the line:

`
Content-Disposition: form-data; name=abc ; filename=xyz.txt ; inline
`

And convert it to the list of SubString objects:

`
[ "name", "=","abc", ";", "filename", "=", "xyz.txt", ";", "inline" ]
`

Then a method content_disposition_extract() which parses that list into a tuple with a boolean indicating of the element is a key value pair, followed by the "key", then the "value" if present so the above line would be turned into:

`
[
    (true, "name", "abc"),
    (true, "filename", "xyz.txt"),
    (false, "inline", nothing)
]
`

Then the name and filename can be extracted.

I tried to merge the latest changes from JuliaWeb/HTTP.jl master into the branch, not sure if that complicated things, since I'm still getting the "Can't automatically merge" warning.  If you would like the pull request without merging in the latest changes I could try to create that instead.


c42f:
It seems something went wrong in 62de6f5 — looks like you may have cherry picked + squashed all the changes from master, rather than merging them?  This is probably why there's still conflicts and so much seems to have changed which is unrelated to multipart parsing. It might be necessary to reset the branch to 431f26ab49ac8f533fe343902e4ed59d9f5f3f94 to clean that up. Then use `git merge origin/master` to merge in the latest code.

pixel27:
Nuts, okay, I'll try it again.

c42f:
The new version in #496 looks a lot better thanks.

Side note - you don't need to close a PR and reopen a new one when this kind of thing happens.  You can always clean up your branch locally and force push it (something like `git push -f parseform`) to keep the discussion in one place. To clean up history, it's a good idea to get comfortable with interactive rebase (`git rebase -i <some_base_commit_on_your_branch>`) which will let you clean up most mistakes in a fairly flexible way.

PR #496, opened 2020-01-10
WIP/RFC: added parser for multipart form with updates

pixel27:
This is basically the https://github.com/JuliaWeb/HTTP.jl/pull/427 pull request with changes asked for by @c42f here https://github.com/JuliaWeb/HTTP.jl/pull/427#pullrequestreview-275747560.

The biggest change where the processing of the arguments on the "Content-Disposition" command according to https://tools.ietf.org/html/rfc2183.  Basically to parse that line correctly I created a content_disposition_tokenize() method that would take the line:

`
Content-Disposition: form-data; name=abc ; filename=xyz.txt ; inline
`

And convert it to the list of SubString objects:

`
[ "name", "=","abc", ";", "filename", "=", "xyz.txt", ";", "inline" ]
`

Then a method content_disposition_extract() which parses that list into a tuple with a boolean indicating of the element is a key value pair, followed by the "key", then the "value" if present so the above line would be turned into:

`
[
    (true, "name", "abc"),
    (true, "filename", "xyz.txt"),
    (false, "inline", nothing)
]
`

Then the name and filename can be extracted.

This is the second attempt for this pull.  The first had a bad merge.  Now it seems to have worked correctly.

### DIFF
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -38,9 +38,17 @@ function find_multipart_boundary(bytes::AbstractVector{UInt8}, boundaryDelimiter
             is_terminating_delimiter && (end_index+=2)
 
             # ... there can be arbitrary SP and HTAB space between the boundary delimiter ...
-            while (end_index < length(bytes) && (bytes[end_index+1] in [HTAB_BYTE, SPACE_BYTE])) end_index+=1 end
+            while (end_index < length(bytes) && (bytes[end_index+1] in [HTAB_BYTE, SPACE_BYTE]))
+                end_index+=1
+            end
             # ... and ends with a new line
-            (end_index < length(bytes)-1) && bytes[end_index+1] == CR_BYTE && bytes[end_index+2] == LF_BYTE || error("boundary delimiter found, but did not end with new line")
+            newlineEnd = end_index < length(bytes)-1 &&
+                         bytes[end_index+1] == CR_BYTE &&
+                         bytes[end_index+2] == LF_BYTE
+            if !newlineEnd
+                error("boundary delimiter found, but did not end with new line")
+            end
+
             end_index += 2
 
             return (is_terminating_delimiter, i, end_index)
@@ -61,7 +69,7 @@ looking for the data between the boundary delimiters in the byte array.  A vecto
 the start/end pairs is returned.
 """
 function find_multipart_boundaries(bytes::AbstractVector{UInt8}, boundary::AbstractVector{UInt8}; start=1)
-    idxs = []
+    idxs = Tuple{Int, Int}[]
     while true
         (is_terminating_delimiter, i, end_index) = find_multipart_boundary(bytes, boundary; start = start)
         push!(idxs, (i, end_index))
@@ -98,7 +106,7 @@ end
 """
     content_disposition_tokenize(str)
 
-Tokenize the "arguments" for the Content-Disposition delcaration.  A vector of
+Tokenize the "arguments" for the Content-Disposition declaration.  A vector of
 strings is returned that contains each token and separator found in the source
 string. Tokens are separated by either an equal sign(=) or a semi-colon(;) and
 may be quoted or escaped with a backslash(\\). All tokens returned are stripped
@@ -199,7 +207,7 @@ function parse_multipart_chunk(chunk)
     disposition = match(r"(?i)Content-Disposition: form-data(.*)\r\n", headers)
 
     if disposition === nothing
-        @warn "Content disposition is not specified dropping the chunk."
+        @warn "Content disposition is not specified dropping the chunk." chunk
         return # Specifying content disposition is mandatory
     end
 
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -1,0 +1,269 @@
+const CR_BYTE = 0x0d # \r
+const LF_BYTE = 0x0a # \n
+const DASH_BYTE = 0x2d # -
+const HTAB_BYTE = 0x09 # \t
+const SPACE_BYTE = 0x20
+const RETURN_BYTES = [CR_BYTE, LF_BYTE]
+
+"""
+    find_multipart_boundary(bytes, boundaryDelimiter; start::Int=1)
+
+Find the first and last index of the next boundary delimiting a part, and if
+the discovered boundary is the terminating boundary.
+"""
+function find_multipart_boundary(bytes::AbstractVector{UInt8}, boundaryDelimiter::AbstractVector{UInt8}; start::Int=1)
+    # The boundary delimiter line is prepended with two '-' characters
+    # The boundary delimiter line starts on a new line, so must be preceded by a \r\n.
+    # The boundary delimiter line ends with \r\n, and can have "optional linear whitespace" between
+    # the end of the boundary delimiter, and the \r\n.
+    # The last boundary delimiter line has an additional '--' at the end of the boundary delimiter
+    # [RFC2046 5.1.1](https://tools.ietf.org/html/rfc2046#section-5.1.1)
+
+    i = start
+    end_index = i+length(boundaryDelimiter)-1
+    while end_index <= length(bytes)
+        if bytes[i:end_index] == boundaryDelimiter
+            # boundary delimiter line start on a new line ...
+            if i > 1
+                (i == 2 || bytes[i-2] != CR_BYTE || bytes[i-1] != LF_BYTE) && error("boundary delimiter found, but it was not the start of a line")
+                # the CRLF preceding the boundary delimiter is "conceptually attached
+                # to the boundary", so account for this with the index
+                i-=2
+            end
+
+            # need to check if there are enough characters for the CRLF or for two dashes
+            end_index < length(bytes)-1 || error("boundary delimiter found, but did not end with new line")
+
+            is_terminating_delimiter = bytes[end_index+1] == DASH_BYTE && bytes[end_index+2] == DASH_BYTE
+            is_terminating_delimiter && (end_index+=2)
+
+            # ... there can be arbitrary SP and HTAB space between the boundary delimiter ...
+            while (end_index < length(bytes) && (bytes[end_index+1] in [HTAB_BYTE, SPACE_BYTE])) end_index+=1 end
+            # ... and ends with a new line
+            (end_index < length(bytes)-1) && bytes[end_index+1] == CR_BYTE && bytes[end_index+2] == LF_BYTE || error("boundary delimiter found, but did not end with new line")
+            end_index += 2
+
+            return (is_terminating_delimiter, i, end_index)
+        end
+
+        i += 1
+        end_index += 1
+    end
+
+    error("boundary delimiter not found")
+end
+
+"""
+    find_multipart_boundaries(bytes, boundary; start=1)
+
+Find the start and end indexes of all the parts of the multipart object.  Ultimately this method is
+looking for the data between the boundary delimiters in the byte array.  A vector containing all
+the start/end pairs is returned.
+"""
+function find_multipart_boundaries(bytes::AbstractVector{UInt8}, boundary::AbstractVector{UInt8}; start=1)
+    idxs = []
+    while true
+        (is_terminating_delimiter, i, end_index) = find_multipart_boundary(bytes, boundary; start = start)
+        push!(idxs, (i, end_index))
+        is_terminating_delimiter && break
+        start = end_index + 1
+    end
+    return idxs
+end
+
+"""
+    find_header_boundary(bytes)
+
+Find the end of the multipart header in the byte array.  Returns a Tuple with the
+start index(1) and the end index. Headers are separated from the body by CRLFCRLF.
+
+[RFC2046 5.1](https://tools.ietf.org/html/rfc2046#section-5.1)
+[RFC822 3.1](https://tools.ietf.org/html/rfc822#section-3.1)
+"""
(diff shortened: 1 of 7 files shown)


mattBrzezinski:
Also needs a version bump, and probably a few commits squashed.

pixel27:
I've performed all the changes suggested by @mattBrzezinski.  I squashed everything but left the version bump as it's own commit.

test/multipart.jl
```diff
@@ -21,4 +21,13 @@
             @test_throws MethodError HTTP.post(uri, body)
         end
     end
+    @testset "HTTP.Multipart ensure show() works correctly" begin
+		# testing that there is no error in printing when nothing is set for filename
+		str = sprint(show, (HTTP.Multipart(nothing, IOBuffer("some data"), "plain/text", "", "testname")))
+		@test findfirst("contenttype=\"plain/text\"", str) != nothing
+	end
+	@testset "HTTP.Multipart test constructor" begin
```
mattBrzezinski: Just needs to be de-intended (is that a word?)

src/parsemultipart.jl
```diff
@@ -0,0 +1,220 @@
...
+        end
+
+        i += 1
+        end_index += 1
+    end
+
+    error("boundary delimiter not found")
+end
+
+function find_multipart_boundaries(bytes::AbstractVector{UInt8}, boundary::AbstractVector{UInt8}; start=1)
```
mattBrzezinski: Would be nice to have docstrs for all the functions in this file.

test/multipart.jl
pixel27: Looks like my editor inserted tabs rather than spaces, fixed the issue, it looks fine now, squashed it into the the main rev.

src/parsemultipart.jl
pixel27: Went through all the functions and added docstrs.  Squashed it down to the single checkin.

pixel27:
Not sure if I should be doing these fixes as single commits and saving the squash for the end.  Currently just squashing as I go......

c42f:
> Not sure if I should be doing these fixes as single commits and saving the squash for the end. Currently just squashing as I go.

During review it's easiest for reviewers if you leave the intermediate "fixup" commits in the history until reviewers have had a chance to look at them. Then when the code seems to have converged you can do a final squash. Of course this is not always clear cut and for complex reviews it may be helpful to do this several times. Generally it's helpful to consider the question "which commit history will be the easiest for other people to understand"? It's all about clear communication with reviewers but also with people consuming the code in the future. (Often that's your future self :-) )

Your commit history here looks a lot simpler now which is good. A couple more things to fix:
* You can remove the merge commit — it seems to be empty now and serves no purpose?
* The squash commit should mention all authors as this code seems to have a long history which has been lost in the `git rebase`. The way to work around this is to add several `Co-authored-by` trailer lines into the commit message by hand as documented at https://github.blog/2018-01-29-commit-together-with-co-authors/#how-it-works. This will make sure that both @piever, @acdupont and yourself are credited with the work they did on this :-) You can fix this up with `git rebase -i` with a "reword" to reword the squash commit message.

pixel27:
Well looks like I screwed up....somehow...removed the merge record and the updated the commit message...and now github thinks there are 24 commits. for this pull request....So the pull request is probably bad till I figure this out.  Sorry.

pixel27:
It *looks* better, but I don't think it's correct yet.

c42f:
The git history looks basically fine now (the attribution looks good :-) ).  You just have a spurious empty merge in there which isn't doing anything.  You should be able to get rid of that easily though, using `git rebase origin/master`.

pixel27:
I think I got everything straightened out now. Please check it over and let me know.

pixel27:
I'm just checking in.  Is there anything else you need from me for this?

c42f:
Thanks for bumping this. It appears tests fail on Julia 1.0 - that will need fixing before this could be merged.

quinnj:
Note that I fixed some tests in #508 that are unrelated to any changes we've made (httpbin.org changed some headers they pass). So you could maybe pull the test file changes into this PR if we merge this one first.

c42f:
Errors on Julia 1.0 seemed related to using new APIs, or at least I noticed `isnothing(x)` which should be written `x === nothing` for 1.0 compatibility.

quinnj:
Note that master has a lot of test cleanups so that CI is more reliable, so you probably want to rebase this, like `git pull --rebase origin master` on your local branch.

pixel27:
Okay all the checks are working now.  If you are happy I can merge the fix into the main check-in.

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+            end
+
+            # need to check if there are enough characters for the CRLF or for two dashes
+            end_index < length(bytes)-1 || error("boundary delimiter found, but did not end with new line")
+
+            is_terminating_delimiter = bytes[end_index+1] == DASH_BYTE && bytes[end_index+2] == DASH_BYTE
+            is_terminating_delimiter && (end_index+=2)
+
+            # ... there can be arbitrary SP and HTAB space between the boundary delimiter ...
+            while (end_index < length(bytes) && (bytes[end_index+1] in [HTAB_BYTE, SPACE_BYTE])) end_index+=1 end
```
quinnj: maybe split this into 3 lines so it's more readable?

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+
+The boundary delimiter MUST NOT appear inside any of the encapsulated parts. Note
+that the boundary delimiter does not need to have '-' characters, but a line using
+the boundary delimiter will start with '--' and end in \r\n.
+[RFC2046 5.1](https://tools.ietf.org/html/rfc2046#section-5.1.1)
+"""
+function parse_multipart_form(req::Request)::Vector{Multipart}
+    # parse boundary from Content-Type
+    m = match(r"multipart/form-data; boundary=(.*)$", req["Content-Type"])
+    m === nothing && return nothing
```
quinnj: It seems like this will cause problems because you've set the function return type as `Vector{Multipart}`; you'll either need to change the return type to `::Union{Nothing, Vector{Multipart}}` or return `Multipart[]`.

quinnj: Also would probably be good to add a test-case for this specifically

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+"""
+    find_header_boundary(bytes)
+
+Find the end of the multipart header in the byte array.  Returns a Tuple with the
+start index(1) and the end index. Headers are separated from the body by CRLFCRLF.
+
+[RFC2046 5.1](https://tools.ietf.org/html/rfc2046#section-5.1)
+[RFC822 3.1](https://tools.ietf.org/html/rfc822#section-3.1)
+"""
+function find_header_boundary(bytes::AbstractVector{UInt8})
```
c42f: I feels like this "find subsequence" operation should be in `Base`, but I can't find it.

The `Base._search` function seems to do what you need but it's internal.

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+
+"""
+    parse_multipart_body(body, boundary)::Vector{Multipart}
+
+Parse the multipart body received from the client breaking it into the various
+chunks which are returned as an array of Multipart objects.
+"""
+function parse_multipart_body(body::AbstractVector{UInt8}, boundary::AbstractString)::Vector{Multipart}
+    multiparts = Multipart[]
+    idxs = find_multipart_boundaries(body, IOBuffer("--$(boundary)").data)
```
c42f: I think `.data` is internal and should be accessed with `take!(io)` instead. But that still seems like a backhanded way to get at the data as a byte array. Perhaps nicer just to use

```suggestion
    idxs = find_multipart_boundaries(body, Vector{UInt8}("--"*boundary))
```

src/parsemultipart.jl
pixel27: I agree and there is a "findnext" method which sounds promising, but it operates on an array of Bools and finds the next true index.

I hadn't looked at this code closely before, I noticed that it's using `bytes[i:end_index]` which appears to create a new vector.  Being in a loop that might cause a surge in garbage collection?  I'm not sure if I should change it to something like `view(bytes, i:end_index)`.

src/parsemultipart.jl
pixel27: Fixed this in a502a23d.

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+            # need to check if there are enough characters for the CRLF or for two dashes
+            end_index < length(bytes)-1 || error("boundary delimiter found, but did not end with new line")
+
+            is_terminating_delimiter = bytes[end_index+1] == DASH_BYTE && bytes[end_index+2] == DASH_BYTE
+            is_terminating_delimiter && (end_index+=2)
+
+            # ... there can be arbitrary SP and HTAB space between the boundary delimiter ...
+            while (end_index < length(bytes) && (bytes[end_index+1] in [HTAB_BYTE, SPACE_BYTE])) end_index+=1 end
+            # ... and ends with a new line
+            (end_index < length(bytes)-1) && bytes[end_index+1] == CR_BYTE && bytes[end_index+2] == LF_BYTE || error("boundary delimiter found, but did not end with new line")
```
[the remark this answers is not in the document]
pixel27: Fixed this in a502a23.

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+        i += 1
+        end_index += 1
+    end
+    error("no delimiter found separating header from multipart body")
+end
+
+"""
+    content_disposition_tokenize(str)
+
+Tokenize the "arguments" for the Content-Disposition delcaration.  A vector of
```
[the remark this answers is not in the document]
pixel27: Fixed this in a502a23.

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+
+"""
+    find_multipart_boundaries(bytes, boundary; start=1)
+
+Find the start and end indexes of all the parts of the multipart object.  Ultimately this method is
+looking for the data between the boundary delimiters in the byte array.  A vector containing all
+the start/end pairs is returned.
+"""
+function find_multipart_boundaries(bytes::AbstractVector{UInt8}, boundary::AbstractVector{UInt8}; start=1)
+    idxs = []
```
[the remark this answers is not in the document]
pixel27: Fixed this in a502a23.

src/parsemultipart.jl
```diff
@@ -0,0 +1,269 @@
...
+function parse_multipart_chunk(chunk)
+    (startIndex, end_index) = find_header_boundary(chunk)
+
+    headers = String(view(chunk, startIndex:end_index))
+    content = view(chunk, end_index+1:lastindex(chunk))
+
+    disposition = match(r"(?i)Content-Disposition: form-data(.*)\r\n", headers)
+
+    if disposition === nothing
+        @warn "Content disposition is not specified dropping the chunk."
```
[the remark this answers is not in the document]
pixel27: Fixed this in a502a23.

src/parsemultipart.jl
pixel27: Fixed this in 67356191.

src/parsemultipart.jl
pixel27: Updated the return type and added a test in 24edb33d.

pixel27:
I think I've addressed the new issues.  I do have a question about using `bytes[i:end_index]` inside a loop as opposed to `view(bytes, i:end_index)` which might be easier on the garbage collector.  Or I could just do four comparisons.  This might all be premature optimizations on my part.....

src/parsemultipart.jl
c42f: I noticed this too but I thought I would avoid nitpicking at that :-)

If you do want to make this allocation free, the most straightforward way will just be to hard code the four values which are searched for. (I believe even `@view`/`view` do allocate a small amount for compiler-technical reasons, though this may be fixed in julia 1.4)

quinnj:
I'm cleaning up some more CI/testing issues in https://github.com/JuliaWeb/HTTP.jl/pull/513; I think I've got it right now and will merge that in the next little while. Unfortuantely the HTTP.jl tests were not in a great state for the last little while which led to a lot of CI unreliability.

src/parsemultipart.jl
```diff
@@ -0,0 +1,277 @@
...
+
+"""
+    parse_multipart_body(body, boundary)::Vector{Multipart}
+
+Parse the multipart body received from the client breaking it into the various
+chunks which are returned as an array of Multipart objects.
+"""
+function parse_multipart_body(body::AbstractVector{UInt8}, boundary::AbstractString)::Vector{Multipart}
+    multiparts = Multipart[]
+    idxs = find_multipart_boundaries(body, Vector{UInt8}("--$(boundary)"))
```
quinnj: I'm not sure if this is that performance critical, but you should be able to do `codeunits("--$boundary")` and it should work the same, but avoiding a copy of the boundary string. `codeunits` returns an `AbstractVector{UInt8}` that can index individual bytes.

c42f: True, I only just discovered what `codeunits` was for at the end of reviewing this.

quinnj:
Ok, I got all clean CI in https://github.com/JuliaWeb/HTTP.jl/pull/514, so I merged it there. Thanks @pixel27 for all your work/help here!

PR #503, opened 2020-01-29
Remove unsupported kwarg basic_authorization for tests of basic auth

fredrikekre:
Remove unsupported kwarg `basic_authorization` for tests of basic authorization, which was missed in #481.


fredrikekre:
bum

fredrikekre:
p

PR #507, opened 2020-02-10
Fix for HTTPS proxy

jademackay:
Instantiate a new `target_url` specifying the port (443) for https proxy.

### DIFF
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -82,6 +82,9 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
 
     try
         if proxy !== nothing && target_url.scheme == "https"
+            target_url = URI(scheme=target_url.scheme,
+                             host=target_url.host,
+                             port=443)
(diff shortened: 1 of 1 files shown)


src/ConnectionRequest.jl
```diff
@@ -82,6 +82,9 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
 
     try
         if proxy !== nothing && target_url.scheme == "https"
+            target_url = URI(scheme=target_url.scheme,
```
quinnj: I think we could just do `target_url = merge(target_url, port=443)`, right? That way we keep the rest of the uri components?

quinnj:
Sorry for the slowness in responding here; I'm trying to follow up on things now. I left a comment above.

src/ConnectionRequest.jl
jademackay: Yes, that is better.
I have made the change and successfully executed `HTTP.request("GET", "https://httpbin.org/ip")`.

jademackay:
On the contrary, I really appreciate all the work you're doing. 👍

PR #509, opened 2020-02-20
Refactor concurrency support

quinnj:


src/ConnectionPool.jl
```diff
@@ -237,12 +250,19 @@ end
...
-    while t.c.writecount != t.sequence
-        wait(t.c.writedone)
-    end                                           ;@debug 2 "👁  Start write:$t"
-    t.c.writebusy = true
+    @require !iswritable(t)
+
+    writebusy(t.c) && @debug 1 "⏳  Wait write: $t"
+    while true
+        lock = take!(t.c.writelock)
+        if t.c.writecount == t.sequence
```
quinnj: I do wonder whether `writecount` and `readcount` should be `Threads.Atomic{Int}` on the `Connection` type; they're only ever updated by a `Transaction` who holds the corresponding `writelock`/`readlock`, but they could be accessed by multiple transactions who are all waiting to pipeline a request through. In my current mental model, this should be fine since the field "updating" is locked, while reading is a free-for-all, but I don't know if there's some kind of potential issue/danger of, say, multiple threads all accessing a single struct's Int field value at the same time.

src/ConnectionPool.jl
```diff
@@ -70,17 +69,19 @@ mutable struct Connection{T <: IO}
...
+    writelock::Channel{Bool}
     readcount::Int
-    readbusy::Bool
-    readdone::Condition
+    readlock::Channel{Bool}
     timestamp::Float64
 end
 
+hashconn(x::Connection{T}) where {T} = hashconn(T, x.host, x.port, x.pipeline_limit, x.require_ssl_verification, x.clientconnection)
+hashconn(T, host, port, pipeline_limit, require_ssl_verification, client) = hash(T, hash(host, hash(port, hash(pipeline_limit, hash(require_ssl_verification, hash(client, UInt(0)))))))
```
quinnj: Used for connection pooling Dict key; these properties "make up" a type of connection that can be re-used by another request that also has the same properties.

src/ConnectionPool.jl
```diff
@@ -125,7 +133,7 @@ getrawstream(t::Transaction) = t.c.io
 inactiveseconds(t::Transaction) = inactiveseconds(t.c)
 
 function inactiveseconds(c::Connection)::Float64
-    if !c.readbusy && !c.writebusy
+    if isbusy(c)
```
quinnj: I think this line was wrong before

src/ConnectionPool.jl
c42f: For some really nice commentary on this, check out yuyichao's post here:
https://github.com/JuliaGPU/CUDAapi.jl/issues/106#issuecomment-585568882
Also look at the article linked at the top.

As a very rough summary: the problem with non-atomic access on the reader side is the reader may see something valid-looking for the variable, but even so, they can't use that to make assumptions about the rest of the data structures in memory which have been touched by the writer thread: the order in which those writes are done can look completely random to the reader even though they are *apparently sequential* in the writer thread.

This is a fact about compilers, but also a fundamental fact about computer hardware which can buffer writes in a cache before flushing them out for other cores to see.

If you have one writer and multiple readers and you are in doubt at all, just use `Atomic` or a lock. If you don't do this there are many very subtle ways things can go wrong. (Perhaps if you don't rely on the value that the readers get — ie, you never take a branch or do a load based on this value you might be ok. But then, why are you reading the value :-) )

src/ConnectionPool.jl
```diff
@@ -70,17 +69,19 @@ mutable struct Connection{T <: IO}
     peerport::UInt16 # debug only
     localport::UInt16 # debug only
     io::T
+    clientconnection::Bool
     buffer::IOBuffer
     sequence::Int
     writecount::Int
-    writebusy::Bool
-    writedone::Condition
+    writelock::Channel{Bool}
```
vtjnash: This probably either has the wrong name or wrong type. Seems like you meant to use a `RecursiveLock` here?

src/ConnectionPool.jl
```diff
@@ -270,13 +290,20 @@ end
...
-                                                   @debug 1 "⏳  Wait read:  $t"
+    @require !isreadable(t)
+
+    readbusy(t.c) && @debug 1 "⏳  Wait read:  $t"
     t.c.timestamp = time()
-    while t.c.readcount != t.sequence
-        wait(t.c.readdone)
-    end                                           ;@debug 2 "👁  Start read: $t"
-    t.c.readbusy = true
+    while true
```
vtjnash: busy waiting on a condition variable is generally a pretty strong anti-pattern

src/ConnectionPool.jl
```diff
@@ -358,115 +383,45 @@ function purge(c::Connection)
...
     closeall()
 
 Close all connections in `pool`.
 """
 function closeall()
-    lock(poollock[Threads.threadid()]) do
-        for c in pool[Threads.threadid()]
-            close(c)
+    for pod in values(POOL.conns)
+        while isready(pod.conns)
```
vtjnash: this is race-y (you should be holding the `pod.conns` lock)

src/ConnectionPool.jl
```diff
@@ -358,115 +383,45 @@ function purge(c::Connection)
...
+            close(take!(pod.conns))
         end
-        empty!(pool[Threads.threadid()])
+        Threads.atomic_xchg!(pod.numactive, 0)
```
vtjnash: weird to use xchg (and atomic at that), but ignore the result. seems like you just mean `pod.numactive[] = 0`?

src/ConnectionPool.jl
```diff
@@ -483,59 +438,65 @@ function getconnection(::Type{Transaction{T}},
...
-                c = Connection(host, port,
-                               pipeline_limit, idle_timeout,
-                               require_ssl_verification,
-                               io)
-                push!(pool[Threads.threadid()], c)                ;@debug 1 "🔗  New:            $c"
-                return client_transaction(c)
+    pod = getpod(POOL, hashconn(T, host, port, pipeline_limit, require_ssl_verification, true))
+    while isready(pod.conns)
+        conn = take!(pod.conns)
+@label check_connection
```
vtjnash:
```suggestion
        @label check_connection
```

src/ConnectionPool.jl
```diff
@@ -483,59 +438,65 @@ function getconnection(::Type{Transaction{T}},
...
-                                                  require_ssl_verification,
-                                                  kw...)
-                c = Connection(host, port,
-                               pipeline_limit, idle_timeout,
-                               require_ssl_verification,
-                               io)
-                push!(pool[Threads.threadid()], c)                ;@debug 1 "🔗  New:            $c"
-                return client_transaction(c)
+    pod = getpod(POOL, hashconn(T, host, port, pipeline_limit, require_ssl_verification, true))
+    while isready(pod.conns)
```
vtjnash: this is race-y—you're missing a lock

src/ConnectionPool.jl
```diff
@@ -358,115 +383,45 @@ function purge(c::Connection)
...
-               c.host == host &&
-               c.port == port &&
-               c.pipeline_limit == pipeline_limit &&
-               c.require_ssl_verification == require_ssl_verification &&
-               c.writecount < reuse_limit &&
-               c.writecount - c.readcount < pipeline_limit + 1 &&
-               isopen(c.io)), pool[Threads.threadid()])
+struct Pod
+    conns::Channel{Connection}
+    numactive::Threads.Atomic{Int}
```
vtjnash: `Base.n_avail(conns)`?

src/ConnectionPool.jl
```diff
@@ -358,115 +383,45 @@ function purge(c::Connection)
...
+incr!(pod::Pod) = Threads.atomic_add!(pod.numactive, 1)
 
-    filter(c->(typeof(c.io) == T &&
-               c.host == host &&
-               c.port == port &&
-               c.readcount >= reuse_limit &&
-               !c.readbusy &&
-               isopen(c.io)), pool[Threads.threadid()])
+function release(c::Connection)
+    pod = getpod(POOL, hashconn(c))
```
vtjnash: You should never use `hash` as a dict key

src/ConnectionPool.jl
quinnj: That makes a lot of sense. Thanks for the explanation!

src/ConnectionPool.jl
quinnj: I'm not familiar w/ `RecursiveLock`; it doesn't seem to be defined in Base? I agree that the "writelock" here is a bit awkward; the idea is that the `Connection` object only wants to allow one Transaction at a time to have a read lock or write lock. So Transactions want to acquire the lock in `startwrite`/`startread`, and wait if it's already locked by another transaction. A `Channel` does the `take!`-ing part of what I want, but we don't really care about what we're getting from the Channel.

Could you point me to wherever RecursiveLock is defined?

src/ConnectionPool.jl
quinnj: Yes, this was one of the last changes I made because I saw some test cases where pipelined requests got out of order; e.g. you have multiple pipelined requests and when one finishes writing, it needs to be the next transaction to read from that same connection; otherwise you get requests that end up with responses to other requests on the same connection.

This code seems to solve it, though it definitely felt "icky".

src/ConnectionPool.jl
quinnj: is it common to lock a channel when calling `isready` and then `take!`? it seems like having a `maybetake!` might be convenient for the use-case of wanting an unblocked `take!` that returns `nothing` if there isn't items available.

src/ConnectionPool.jl
quinnj: ah yes, I think I was just deep in the Atomic docs and saw this and used it w/o thinking about it more reasonably

src/ConnectionPool.jl
quinnj: Not quite. We don't keep all our connections in `conns`, only ones that are idle/available to re-use. So connections are "owned" by a transaction while there's a request going on, and then `put!` back in `conns` when the transaction finishes. So in order to track max pool size, we need to keep track of the total # of connections that are out there; currently transacting or idle in our `conns`.

src/ConnectionPool.jl
quinnj: why? my thought here was avoiding the Tuple allocation of doing `(T, host, port, pipeline_limit, require_ssl_verification)` and having that be the pool Dict key.

Is there some danger w/ the core hash value changing during the lifetime of a session/application or something?

src/ConnectionPool.jl
c42f: By the way, there are also more obvious problems with non-atomic loads for variables which are wider than the machine's native width or misaligned in a way that particular hardware doesn't like: in those cases it's possible to see a partial update (I believe this is what yuyichao is referring to as a "torn read" in the other thread).

src/ConnectionPool.jl
```diff
@@ -37,7 +37,6 @@ const default_connection_limit = 8
 const default_pipeline_limit = 16
 const nolimit = typemax(Int)
```
c42f: Are the docs at the top of this file for the `ConnectionPool` still an accurate description of how this stuff works after these changes?

src/ConnectionPool.jl
c42f: `Threads.ReentrantLock` perhaps?

src/ConnectionPool.jl
c42f: Because of hash collisions?

c42f: I mean, does it matter that you may have `hashconn(c1) == hashconn(c2)` for different `c1` and `c2`?

src/ConnectionPool.jl
c42f: That makes sense - could do with just putting your explanation into the code as a comment?

c42f: Though for the reasons discussed further down it may not be helpful to do the hash here, unless the lookup is insensitive to hash collisions.  Could be renamed `connection_reuse_key`?

src/ConnectionPool.jl
oxinabox: machine's native width isn't enough. Only atomic is atomic.
Modern CPUs are wild, and anything can be nonatomic if it feels like it.
There are no promises that you won't see torn reads of 32  bit values on a 32 bit system.

src/ConnectionPool.jl
quinnj: The issue with ReentrantLock is that the locking/unlocking may occur in different tasks (though wouldn't ever occur in different threads); this is because the lock tends to happen in the "main" thread doing the toplevel request, whereas the unlock (may) occur in the `@async writebody` task. I tried this before and ReentrantLock complained about the wrong task unlocking it.

I found https://www.justsoftwaresolutions.co.uk/threading/locks-mutexes-semaphores.html, which seems like a pretty good walk-through of basic threading primitives. From that, it feels like what we want here is a Semaphore, since we want the locking and signaling functionality, but are not as concerned with "who" does the acquire/release, since we're managing that logic ourselves through our `Transaction` structure.

quinnj: Ok, after some more mulling over, I switched our lock structure here back to a `Condition`. Correct me if I'm wrong, but after studying the `Channel` implementation in Base, a `Condition` is like a lock that can also signal to waiters, which is basically what we want here. So we have `writelock::Condition` and `writebusy::Threads.Atomic{Bool}`, so query-ers can check if a Connection is busy by checking `conn.writebusy[]`, and we manage the setting/unsetting of `writebusy` in `startwrite`/`closewrite` exclusively. We also currently have the `writelock` locked when updating `writebusy` and `writecount`, even though they're atomics anyway, but just to more document the data we're "locking" updates on.

src/ConnectionPool.jl
quinnj: Moving this back to a `lock` -> check -> `wait` pattern on a `Condition` feels much more standard.

src/ConnectionPool.jl
quinnj: Updated

src/ConnectionPool.jl
quinnj: Updated

src/ConnectionPool.jl
quinnj: Updated

src/ConnectionPool.jl
quinnj: I've added some docs to `hashconn`; yeah, I guess it's a bit of a misnomer using `hash` in the name, but what we're really checking here is whether Connections could be re-useable by multiple requests, so we're only interested in hashing a few key properties.

src/ConnectionPool.jl
c42f: > anything can be nonatomic if it feels like it

Well I don't think it's so bad as that; normal writes still need to go through the some of the same hardware so there will be some guarantees better than "all bets are off" (the following article suggests word size and alignment are the main things https://dave.cheney.net/2018/01/06/if-aligned-memory-writes-are-atomic-why-do-we-need-the-sync-atomic-package).

But regardless of that, "just use `Atomic`" is the right option for high level source code.

quinnj:
I'm going to merge this now, since it fixes the original issue, and fixes HTTP.jl tests overall currently blocking other PRs. I'd still appreciate any feedback or review that people want to give.

src/ConnectionPool.jl
```diff
@@ -291,10 +347,13 @@ Increment `readcount` and wake up tasks waiting in `startread`.
 function IOExtras.closeread(t::Transaction)
     @require isreadable(t)
 
-    t.c.readbusy = false
-    t.c.readcount += 1
-    notify(t.c.readdone)                          ;@debug 2 "✉️  Read done:  $t"
-    notify(poolcondition[Threads.threadid()])
+    @v1_3 lock(t.c.readlock)
```
vtjnash: `lock` should always be followed by `try/finally`

src/ConnectionPool.jl
```diff
@@ -291,10 +347,13 @@ Increment `readcount` and wake up tasks waiting in `startread`.
...
+    t.readbusy[] = false
+    Threads.atomic_add!(t.c.readcount, 1)         ;@debug 2 "✉️  Read done:  $t"
+    t.c.readbusy[] = false
+    notify(t.c.readlock)
+    @v1_3 unlock(t.c.readlock)
+    release(t.c)
```
vtjnash: any lines of code appearing after these statement are probably invalid

src/ConnectionPool.jl
```diff
@@ -483,59 +473,87 @@ function getconnection(::Type{Transaction{T}},
...
-    while true
-
-        lock(poollock[Threads.threadid()])
-        try
-            # Close connections that have reached the reuse limit...
-            if reuse_limit != nolimit
-                for c in findoverused(T, host, port, reuse_limit)
-                    close(c)
-                end
+    pod = getpod(POOL, hashconn(T, host, port, pipeline_limit, require_ssl_verification, true))
```
vtjnash: This does not look like a valid use of a `hash` function: you need to use the tuple of values, not their hash, as the lookup key.

src/ConnectionPool.jl
```diff
@@ -358,115 +416,47 @@ function purge(c::Connection)
...
-    for c in pool[Threads.threadid()]
-        if c.idle_timeout > 0 &&
-          !c.readbusy &&
-          !c.writebusy &&
-           time() - c.timestamp > c.idle_timeout
-
-            close(c.io)                       ;@debug 1 "⌛️  Timeout:        $c"
-        end
-    end
+const POOL = Pool(Dict{UInt, Pod}())
```
vtjnash: This object appears to be missing a lock

src/ConnectionPool.jl
```diff
@@ -358,115 +416,47 @@ function purge(c::Connection)
...
 """
     closeall()
 
 Close all connections in `pool`.
 """
 function closeall()
-    lock(poollock[Threads.threadid()]) do
-        for c in pool[Threads.threadid()]
-            close(c)
+    for pod in values(POOL.conns)
```
vtjnash: missing lock on `POOL` access?

src/ConnectionPool.jl
```diff
@@ -358,115 +416,47 @@ function purge(c::Connection)
...
+        @v1_3 lock(pod.conns)
+        while isready(pod.conns)
+            close(take!(pod.conns))
         end
-        empty!(pool[Threads.threadid()])
+        @v1_3 unlock(pod.conns)
+        pod.numactive[] = 0
```
vtjnash: this is a data race

src/ConnectionPool.jl
```diff
@@ -483,59 +473,87 @@ function getconnection(::Type{Transaction{T}},
...
-                                                   require_ssl_verification,
-                                                   reuse_limit)
-            idle = filter(c->!c.readbusy, writable)
-            if !isempty(idle)
-                c = rand(idle)                 ;@debug 2 "♻️  Idle:           $c"
-                return client_transaction(c)
-            end
+    # If there are not too many connections to this host:port,
+    # create a new connection...
+    if pod.numactive[] < connection_limit
```
vtjnash: missing a lock—and doesn't seem like a benign race

src/ConnectionPool.jl
```diff
@@ -483,59 +473,87 @@ function getconnection(::Type{Transaction{T}},
...
-                               io)
-                push!(pool[Threads.threadid()], c)                ;@debug 1 "🔗  New:            $c"
-                return client_transaction(c)
-            end
+    # Wait for `closewrite` or `close` to signal that a connection is ready.
+    # sym = gensym()
+    # @debug 0 "$sym: connection pool full, waiting..."
+    # tt = time()
+    while true
+        conn = take!(pod.conns)
```
vtjnash: missing a lock

src/ConnectionPool.jl
```diff
@@ -483,59 +473,87 @@ function getconnection(::Type{Transaction{T}},
...
+        if isvalid(pod, conn, reuse_limit, pipeline_limit)
+            # this is a reuseable connection, so use it
+            @debug 2 "2 reusing connection: $conn"
+            return client_transaction(conn)
+        elseif pod.numactive[] < connection_limit
```
vtjnash: this a data race, needs a lock

src/ConnectionPool.jl
quinnj: Can you explain more why this isn't a valid use of a `hash` function? The reason for using `hash` in my mind is to avoid allocating the tuple just to do the pod lookup, since we don't really need the tuple of values other than for looking up the right pod of connections.

src/ConnectionPool.jl
```diff
@@ -483,59 +473,87 @@ function getconnection(::Type{Transaction{T}},
...
     end
+
+    return !writebusy(conn)
+end
+
+function newconnection(pod, T, host, port, pipeline_limit, require_ssl_verification, idle_timeout; kw...)
+    io = getconnection(T, host, port;
+            require_ssl_verification=require_ssl_verification, kw...)
+    c = Connection(host, port, pipeline_limit, idle_timeout, require_ssl_verification, io)
+    incr!(pod)
```
vtjnash: This is not atomic without something also being synchronized—probably should have a lock and should not use `Atomic`

src/ConnectionPool.jl
vtjnash: It's just not what a hash function does

vtjnash:
Left some specific comments. One high level comment is that you should stop using `Atomic`, as all of your uses are either wrong (missing a lock), or redundant (already inside a locked region)

src/ConnectionPool.jl
```diff
@@ -237,12 +272,21 @@ end
 Wait for prior pending writes to complete.
 """
 function IOExtras.startwrite(t::Transaction)
-    @require !iswritable(t)                     ;t.c.writecount != t.sequence &&
-                                                   @debug 1 "⏳  Wait write: $t"
-    while t.c.writecount != t.sequence
-        wait(t.c.writedone)
-    end                                           ;@debug 2 "👁  Start write:$t"
-    t.c.writebusy = true
+    @require !iswritable(t)
```
vtjnash: Why do (and did) we appear to assert that there wasn't a previous write, then immediate wait for a previous write to finish?? This could use some comments.

src/ConnectionPool.jl
```diff
@@ -255,10 +299,13 @@ Signal that an entire Request Message has been written to the `Transaction`.
...
-    notify(poolcondition[Threads.threadid()])
+    @v1_3 lock(t.c.writelock)
+    t.writebusy[] = false
+    Threads.atomic_add!(t.c.writecount, 1)          ;@debug 2 "🗣  Write done: $t"
+    t.c.writebusy[] = false
+    notify(t.c.writelock)
+    @v1_3 unlock(t.c.writelock)
+    release(t.c)
 
     @ensure !iswritable(t)
```
vtjnash: This is false now, as you've already released the writelock on the object (the `unlock` call a few lines earlier), so you cannot safely look at any fields now

src/ConnectionPool.jl
```diff
@@ -339,7 +398,6 @@ function Base.close(c::Connection)
     if bytesavailable(c) > 0
         purge(c)
     end
-    notify(poolcondition[Threads.threadid()])
     return
 end
```
vtjnash: The below function (purge) seems like it should be holding the writecount-lock to be valid (otherwise access to `c.io` is a potential data-race)

src/ConnectionPool.jl
```diff
@@ -70,17 +85,34 @@ mutable struct Connection{T <: IO}
...
+    clientconnection::Bool
     buffer::IOBuffer
-    sequence::Int
-    writecount::Int
-    writebusy::Bool
-    writedone::Condition
-    readcount::Int
-    readbusy::Bool
-    readdone::Condition
+    sequence::Threads.Atomic{Int}
```
vtjnash: I noticed after saying that that this Atomic seems correct to keep (or you could make it non-atomic and protected by the writebusy-lock, but then the ordering would be imposed by `startwrite` instead of `Transaction`, which might not be what you wanted)

src/ConnectionPool.jl
quinnj: `@require !iswritable(t)` just asserts that _this specific_ transaction hasn't been written yet, so we require that first, and then try to acquire the write lock of the connection, if we get it, but the ordering doesnt' match yet (i.e. we need to wait for a previously pipelined request to finish first), that's when we wait for another `closewrite` to notify that it's done and we'll check if it's our turn.

src/ConnectionPool.jl
quinnj: Yeah, I'm refactoring right now and I left this as Atomic.

PR #515, opened 2020-02-26
In query key-value pairs, decode plus as space

quinnj:


garborg:
This seems off to me:
```jl
julia> HTTP.queryparams(HTTP.escapeuri(Dict("a+b" => "c d+e")))
Dict{String,String} with 1 entry:
  "a b" => "c d e"
```

quinnj:
Why?

garborg:
Is it wrong for values to have a literal "+", escaped as "%2B", and expect them to round trip to "+" after unescaping?

garborg:
I.e. I thought"+"-decoding had to happen before pct-unescaping.

quinnj:
Yeah, you're right. My bad. I rushed this a bit. I've got a fix incoming.

PR #521, opened 2020-03-14
Update status_messages.jl

o314:
Add some http status codes used by [symfony](https://symfony.com/projects) [v4.0.9](https://github.com/symfony/http-foundation/blob/v4.0.9/Response.php).
Rechecked against <https://developer.mozilla.org/en-US/docs/Web/HTTP/Status> too.

*gotten via*
`join -v 1 php_symfo_4-0-9.csv jl_http_v7253252.csv`


o314:
Glad to help !

PR #529, opened 2020-04-15
Re-arrange error handling a bit

Keno:
If we did throw an error and tried to close the connection,
there's no reason to try to do so again during the finally block.
Also, if we are in the error block, close(io) has quite a high chance
for failing, so wrap it in try to suppress such errors and instead
rethrow the original error.

### DIFF
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -86,17 +86,18 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
             return tunnel_request(Next, io, target_url, req, body; kw...)
         end
 
-        return request(Next, io, req, body; kw...)
+        r =  request(Next, io, req, body; kw...)
 
-    catch e
-        @debug 1 "❗️  ConnectionLayer $e. Closing: $io"
-        close(io)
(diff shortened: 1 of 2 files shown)


Keno:
This isn't quite right. I'll fix it in the morning.

Keno:
I've pushed another commit here that tries to propagate the correct error to the user. Before this, e.g. the AWS retry layer would get very mad at the random errors thrown by these close functions and just retry a request that failed for good reason.

src/ConnectionRequest.jl
```diff
@@ -86,17 +86,18 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
...
-    finally
         if (io.sequence >= reuse_limit
         || (proxy !== nothing && target_url.scheme == "https"))
             close(io)
         end
+
+        return r
+    catch e
+        @debug 1 "❗️  ConnectionLayer $e. Closing: $io"
+        try; close(io); catch; end
```
vtjnash, line 99: I think this is why we say `close` shouldn't error?

Keno: Yeah...

PR #536, opened 2020-05-09
Documentation: header() isn't exported so you need to use HTTP.header().

pixel27:
Just a small documentation change I ran across.


quinnj:
Thanks!

PR #538, opened 2020-05-20
don't print non-UTF8 binary data when showing Message

stevengj:
When `show`ing a `Message`, you should omit printing a snippet of the body if it is raw binary data and not valid text.

See also JuliaLang/IJulia.jl#927


stevengj:
I'll add a test shortly.  I just noticed that we can't use `isvalid(String, x)` for subarrays `x`, so I'll update it to use the lower-level `Base.byte_string_classify` function until support can be added to Base (or Compat) via JuliaLang/julia#36047.

stevengj:
Test added.

stevengj:
I'm not sure what the "Labeler" CI script is about?

PR #541, opened 2020-05-27
Add a method to `escapeuri` for `NamedTuple`s

### LINKED ISSUE #540: NamedTuple as query
NamedTuples don't really behave as sources of query string parameters, I think we just need one more method of `escapeuri`.

```jl
julia>  HTTP.escapeuri((foo="bar",))
"%62=%61"

julia>  HTTP.escapeuri(Dict(:foo => "bar"))
"foo=bar"
```

christopher-dG:
Closes #540 

Turns out it was a pretty easy fix...

### DIFF
--- a/test/uri.jl
+++ b/test/uri.jl
@@ -485,6 +485,9 @@ urltests = URLTest[
         @test HTTP.queryparams(HTTP.URI("http://www.example.com/path/foo+bar/path?query+name=query+value")) ==
             Dict("query name" => "query value")
         @test HTTP.queryparams(HTTP.escapeuri(Dict("a+b" => "c d+e"))) == Dict("a+b" => "c d+e")
+
+        # Issue 540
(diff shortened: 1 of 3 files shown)


test/uri.jl
```diff
@@ -485,6 +485,9 @@ urltests = URLTest[
         @test HTTP.queryparams(HTTP.URI("http://www.example.com/path/foo+bar/path?query+name=query+value")) ==
             Dict("query name" => "query value")
         @test HTTP.queryparams(HTTP.escapeuri(Dict("a+b" => "c d+e"))) == Dict("a+b" => "c d+e")
+
+        # Issue 540
+        @test HTTP.escapeuri((a=1, b=1)) == "a=1&b=2"
```
quinnj, line 490: Looks like you need `(a=1, b=2)` or change the answer to `"a=1&b=1`

christopher-dG: Oh woops, typo 🙃

PR #550, opened 2020-06-30
[BREAKING] Don't close response_stream after writing response body

### LINKED ISSUE #526: incorrect streamed output in case of redirect
I'm using the `response_stream` argument to save an HTTP response directly to disk:
```jl
open(path, write=true) do io
    HTTP.get(url, response_stream=io, status_exception=false)
end
```
However, in this case if there is a redirect (301 or 302), then the body of the initial redirect response is what gets saved to the output file at `path` rather than what I want, which is the body of the final HTTP response in the redirect sequence.

HTTP v0.8.13
Julia Version 1.5.0-DEV.529
Commit 110e4356dc (2020-03-30 13:55 UTC)

quinnj:
Fixes #526. The original issue was when a response_stream was passed to
a request, and that request directed, the _redirect_ body was written
to the response_stream, then the response_stream was closed, then we
followed the redirect, and tried to write the subsequent _redirected_
body to the response_stream again, but it was already closed. Curiously,
`IOStream` doesn't throw a non-writeable error when trying to write to
it closed.

The change here involves two things:
* Not closing response_stream after we've written the response body to
it; this is the technically breaking piece, though I think it's probably
not super common that people were relying on this behavior. The HTTP.jl
tests had a few cases of using a `Base.BufferStream` that expected to be
closed, but even the HTTP.jl docs for response_stream include doing the
open-io-with-do-block example, which will close the IOStream for you.
* The 2nd change is that in the RedirectLayer, we now call `mark` and
`reset` on the response_stream, to ensure that the _redirected_ body
gets written where the original _redirect_ body was written to.

I'd appreciate any feedback/concerns on these changes; they sound like
good things in theory, but I'm just not sure if we'll be really breaking
anything for people.

And to top things off, httpbin.org's redirect testing endpoints aren't working anymore; seriously?!

### DIFF
--- a/src/RedirectRequest.jl
+++ b/src/RedirectRequest.jl
@@ -17,11 +17,15 @@ export RedirectLayer
 
 function request(::Type{RedirectLayer{Next}},
                  method::String, url::URI, headers, body;
-                 redirect_limit=3, forwardheaders=true, kw...) where Next
+                 redirect_limit=3, forwardheaders=true,
+                 response_stream=nothing, kw...) where Next
     count = 0
     while true
-
-        res = request(Next, method, url, headers, body; kw...)
+        # save position of response_stream in case of redirect
+        if response_stream !== nothing
+            mark(response_stream)
+        end
+        res = request(Next, method, url, headers, body; response_stream=response_stream, kw...)
 
         if (count == redirect_limit
         ||  !isredirect(res)
@@ -53,6 +57,9 @@ function request(::Type{RedirectLayer{Next}},
         @debug 1 "➡️  Redirect: $url"
 
         count += 1
+        if response_stream !== nothing
+            reset(response_stream)
+        end
     end
 
     @assert false "Unreachable!"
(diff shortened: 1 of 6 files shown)


iamed2:
AWSCore provides the option to return the response_stream, and creates a BufferStream to do so. However, I can't see how people could use this API to actually perform streaming requests, since AWSCore still calls the request function synchronously. I can't find any examples of people using this API on GitHub, and Invenia never uses this option internally. 

> And to top things off, httpbin.org's redirect testing endpoints aren't working anymore; seriously?!

Can you run the httpbin docker image instead?

quinnj:
Thanks for the response @iamed2; due to some concerns from @staticfloat, I've put up an alternative approach [here](https://github.com/JuliaWeb/HTTP.jl/pull/552), which should be non-breaking and more robust to what we're trying to solve here.

> Can you run the httpbin docker image instead?

Hmmm.....now here's an idea. I'll look into it.

PR #557, opened 2020-07-21
Capture errors in WebSocket

AStupidBear:


christopher-dG:
What exactly is this doing? In the first case, the `throw` seems to be just adding a stack frame, whereas the second case with `rethrow` seems to be doing nothing at all. Could you elaborate on what this fixes?

PR #560, opened 2020-07-25
Couple of fixes for various things and making cookie tracking on by default

quinnj:
These are just a few small fixes. The biggest change is making `cookies=true` by default, which means we'll track cookies for users automatically. There was at least one case recently where a user was frustrated because it didn't seem like HTTP.jl could do "session management" like they were used to in python `requests`. We've made a couple of changes recently that make HTTP.jl act more and more like a friendlier client, so I think this makes sense.

cc-ing a couple people in case they have opinions on tracking cookies by default: @staticfloat , @StefanKarpinski , @oxinabox , @mattBrzezinski

### DIFF
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -555,7 +555,7 @@ relationship with [`HTTP.Response`](@ref), [`HTTP.Parsers`](@ref),
 """
 function stack(;redirect=true,
                 aws_authorization=false,
-                cookies=true,
+                cookies=false,
                 canonicalize_headers=false,
                 retry=true,
                 status_exception=true,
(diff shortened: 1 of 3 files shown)


staticfloat:
That's fine with me.  Does HTTP also remember HTTP 301 requests?

quinnj:
Ok, I reverted the `cookies=true` change for now, since I'd like to get more feedback and because I'm trying to get these other bugs fixed quickly.

oxinabox:
I am a big fan of keeping managing the cookies. Cookie management by hand is just annoying.
You got a cookie, you probably wanted it and needed it for functionality.

quinnj:
Cool. Just fyi, I took out the cookies=true part of this PR before merging, because there were a bunch of tests that need updating and I needed the other bugfixes for my workshop this morning, but I'll clean it up and do another PR to make cookies auto-handled.

PR #561, opened 2020-07-26
Never generate empty filename for download

christopher-dG:
Ran into a strange error when I was trying `HTTP.download`, it tried to save the file to `/tmp/`, which obviously is not a valid path for a file. Found out that the Content-Disposition header listed an empty filename which is... odd.

```jl
julia> using HTTP

julia> resp = HTTP.get("https://pretalx.com/media/juliacon2020/images/9A8DCP/logo_kafyqWH.png");

julia> println(HTTP.header(resp, "Content-Disposition"))
attachment; filename=""
```

I could add this to the test suite but I'm afraid of the URL not being very permanent. Thoughts on that?

### DIFF
--- a/src/download.jl
+++ b/src/download.jl
@@ -29,7 +29,7 @@ function try_get_filename_from_headers(resp)
                 # It was in quotes, so it will be double escaped
                 filename = unescape_string(quoted_filename[1])
             end
-            return filename
+            return filename == "" ? nothing : filename
         end
     end
     return nothing
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "HTTP"
 uuid = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 authors = ["Jacob Quinn", "Sam O'Connor", "contributors: https://github.com/JuliaWeb/HTTP.jl/graphs/contributors"]
-version = "0.8.17"
+version = "0.8.18"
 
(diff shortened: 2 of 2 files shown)


src/download.jl
```diff
@@ -29,7 +29,7 @@ function try_get_filename_from_headers(resp)
                 # It was in quotes, so it will be double escaped
                 filename = unescape_string(quoted_filename[1])
```
quinnj, line 30: Unrelated to this PR, but is `quoted_filename[1]` valid? I don't think that works when doing `match`, because it returns a `RegexMatch` object, right? So trying to index it is going to be an error? Am I missing something here?

pfitzseb: `RegexMatch`es are indexable:
```
julia> match(r"(x)(y)", "xyz")[1]
"x"

julia> match(r"(x)(y)", "xyz")[2]
"y"
```

quinnj: Ah, my confusion was missing the match groups in the regex. Thanks.

StefanKarpinski:
Is it possible that this is because you can open a directory these days and it acts like an empty file? That is a bug: https://github.com/JuliaLang/julia/issues/36605.

christopher-dG:
I'm guessing it was totally unintentional, having a filename set in Content-Disposition as explicitly empty seems like an edge case.

PR #569, opened 2020-08-14
Fix Websocket status codes

SimonDanisch:
* apply mask unconditional to fix status codes
* show more descriptive messages about Websocket status


quinnj:
Thanks! We're always looking for/love contributions to the websocket code! If you're able, we'd love some help on other issues w/ the websocket label!

PR #580, opened 2020-09-16
Ensure https proxy connections don't clog up the ConnectionPool

### LINKED ISSUE #562: HTTPS request gets blocked while using proxy
```
http_proxy=http://127.0.0.1:3128 https_proxy=http://127.0.0.1:3128 julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.1 (2020-04-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |
julia> using HTTP
julia> count = 0
0
julia> using HTTP

julia> while true
           global count
           count+=1
           @info "Request $count"
           HTTP.get("https://example.com")
       end
[ Info: Request 1
[ Info: Request 2
[ Info: Request 3
[ Info: Request 4
[ Info: Request 5
[ Info: Request 6
[ Info: Request 7
[ Info: Request 8 ## This gets blocked when default connection limit is 8
```

This happens only for HTTPS connections. 

Julia 1.4.1
HTTP.jl 0.8.16 
MbedTLS.jl 1.0.2
Proxy: https://hub.docker.com/r/datadog/squid/

quinnj:
Fixes #562. Despite lots of back and forth on this issue, it turns out
there's a simpler diagnosis & fix that worrying too much about when
sockets will throw when `close`ing (which we should certainly get all
cleaned up one day).

The _real_ issue here is in `ConnectionPool.sslupgrade` which was
effectively swapping our `Connection` that we formally "acquired" from
the connection pool with a new one that had a https-ssl-upgraded
`SSLContext` to do the request tunneling. The issue with just swapping
the `Connection` is that the original connection was never properly
"released" to the connection pool, so connections were acquired, but
never released, and eventually we hit the `connection_limit`. The
proposed solution here (which works locally to avoid getting hung up on
the connection pool clogging), is to introduce a new
`kill!(c::Connection)` method that "releases" a connection without
returning it to the connection pool. So we signal to the pool that a
connection has finished, but it's not put back in the queue to be
re-used, which is what we want. I should note that this does mean that
proxied https requests then don't really respect the `connection_limit`,
but also note that it never has due to this "swapping".

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -482,6 +482,22 @@ function release(c::Connection)
     return
 end
 
+# "release" a Connection, without returning to pod for re-use
+# used for https proxy tunnel upgrades which shouldn't be reused
+function kill!(c::Connection)
+    h = hashconn(c)
+    if haskey(POOL.conns, h)
+        pod = getpod(POOL, h)
+        @v1_3 lock(pod.conns)
+        try
+            decr!(pod)
+        finally
+            @v1_3 unlock(pod.conns)
+        end
+    end
+    return
+end
+
 struct Pool
     lock::ReentrantLock
     conns::Dict{UInt, Pod}
@@ -694,6 +710,7 @@ function sslupgrade(t::Transaction{TCPSocket},
                         require_ssl_verification=require_ssl_verification,
                         kw...)
     c = Connection(tls; require_ssl_verification=require_ssl_verification)
+    kill!(t.c)
     return client_transaction(c)
 end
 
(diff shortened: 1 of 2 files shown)


vdayanand:
@quinnj Thanks a lot for the fix!

Issue #585, opened 2020-09-21
Example application fails with `UndefVarError: @ip_str not defined`


clarkevans:
Julia 1.5.1
HTTP.jl 0.8.19 
MbedTLS.jl 1.0.2

The example application in the documentation fails to load at the very last line.

```
julia> HTTP.serve(ANIMAL_ROUTER, ip"127.0.0.1", 8080)
ERROR: LoadError: UndefVarError: @ip_str not defined
in expression starting at REPL[15]:1
```

fredrikekre:
Needs `using Sockets` (but you probably figured that out).

PR #586, opened 2020-09-22
add missing Sockets import to documentation example

clarkevans:
The webserver example doesn't work because it's missing `using Sockets`.  This fixes https://github.com/JuliaWeb/HTTP.jl/issues/585.


quinnj (approved):
Thanks!

PR #588, opened 2020-09-24
Update URI docstring

fonsp:
My changes fix the following issues:

#### Old: "A type representing a valid uri."

This is misleading, because the type also represents invalid URIs. 
```julia
julia> HTTP.URI("/////::::    ")
HTTP.URI("/////::::    ")
```

#### Old: "With a raw, already-encoded uri string, use `parse(HTTP.URI, str)` to parse the HTTP.URI directly. "
This sounds like something that I should do, but it describes the internals. 

#### Added note about unescaping
The current docstring suggests that `HTTP.URI(str)` will unescape that string, but it does not. I have added a note about how to unescape.

### DIFF
--- a/src/URIs.jl
+++ b/src/URIs.jl
@@ -21,11 +21,12 @@ end
     HTTP.URI(; scheme="", host="", port="", etc...)
     HTTP.URI(str) = parse(HTTP.URI, str::String)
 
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Thanks!

PR #593, opened 2020-10-01
Fix code example in the docstring for `serve`. (#592)

### LINKED ISSUE #592: Error in the docstring for HTTP.serve
The `HTTP.serve` docstring contains [a code example](https://github.com/JuliaWeb/HTTP.jl/blob/75ce5369eff54982f4a807400e4973108f61c215/src/Handlers.jl#L335) that uses `Sockets.serve`, which isn't a function in the standard library.

kernelmethod:
Replace incorrect call to Sockets.serve with a call to Sockets.listen in
the code example demonstrating how to start an HTTP server using a
server socket.

Closes #592

### DIFF
--- a/src/Handlers.jl
+++ b/src/Handlers.jl
@@ -332,7 +332,7 @@ end
 
 # pass in own server socket to control shutdown
 using Sockets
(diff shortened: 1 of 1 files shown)


quinnj:
Thanks!

PR #599, opened 2020-10-06
RFC: Enable user to run functions on server shutdown

### LINKED ISSUE #591: Feature request: on_shutdown for closing gracefully
Add similar functionality to aiohttp's [`on_shutdown`](https://docs.aiohttp.org/en/stable/web_advanced.html#aiohttp-web-graceful-shutdown) function.

This would enable users to specify custom shutdown functions (for example close database connections) that would trigger when an HTTP server stops. I'm not sure on the how its routed through, but it would be great if this could be triggered when, for example, `CMD+C`ing a Docker container which is running a Julia App hosting a HTTP server.

I'm happy to work on this and will give it a look later, but if anyone has any pointers that would be great.

mmiller-max:
This PR adds the functionality for the user to add one or more functions to be run when the server is shutting down. This is useful for things like ensuring database connections are closed.

Example use:
```Julia
# One function
myshutdownfn() = println("I'm shutting down")

function run()
    HTTP.serve(requestHandler, "0.0.0.0", 5082, on_shutdown=myshutdownfn)
end

# Multiple functions
myshutdownfn1() = println("I'm shutting down")
myshutdownfn2() = println("I'm still shutting down")

function run()
    HTTP.serve(requestHandler, "0.0.0.0", 5082, on_shutdown=[myshutdownfn1, myshutdownfn2])
end
```

This needs some documentation/examples/tests, but wanted to check the approach was okay first.

Closes #591.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -74,10 +74,29 @@ struct Server{S <: Union{SSLConfig, Nothing}, I <: Base.IOServer}
     server::I
     hostname::String
     hostport::String
+    on_shutdown::Any
 end
 
 Base.isopen(s::Server) = isopen(s.server)
-Base.close(s::Server) = close(s.server)
+Base.close(s::Server) = (shutdown(s.on_shutdown); close(s.server))
+
+"""
+    shutdown(fns::Vector{<:Function})
+    shutdown(fn::Function)
+    shutdown(::Nothing)
+
+Runs function(s) in `on_shutdown` field of `Server` when
+`Server` is closed.
+"""
+shutdown(fns::Vector{<:Function}) = foreach(shutdown, fns)
+shutdown(::Nothing) = nothing
+function shutdown(fn::Function)
+    try
+        fn()
+    catch e
+        @error "shutdown function $fn failed" exception=(e, catch_backtrace())
+    end
+end
 
 Sockets.accept(s::Server{Nothing}) = accept(s.server)::TCPSocket
 Sockets.accept(s::Server{SSLConfig}) = getsslcontext(accept(s.server), s.ssl)
@@ -118,6 +137,10 @@ Optional keyword arguments:
     allowed per client IP address; excess connections are immediately closed.
     e.g. 5//1.
  - `verbose::Bool=false`, log connection information to `stdout`.
+ - `on_shutdown::Union{Function, Vector{<:Function}, Nothing}=nothing`, one or
+    more functions to be run if the server is closed (for example by an
+    `InterruptException`). Note, shutdown function(s) will not run if a
+    `IOServer` object is supplied and closed by `close(server)`.
 
 e.g.
 ```julia
@@ -203,7 +226,8 @@ function listen(f,
                 rate_limit::Union{Rational{Int}, Nothing}=nothing,
                 reuse_limit::Int=nolimit,
                 readtimeout::Int=0,
-                verbose::Bool=false)
+                verbose::Bool=false,
+                on_shutdown::Union{Function, Vector{<:Function}, Nothing}=nothing)
 
     inet = getinet(host, port)
     if server !== nothing
@@ -230,7 +254,7 @@ function listen(f,
         x -> f(x) && check_rate_limit(x, rate_limit)
     end
 
-    s = Server(sslconfig, tcpserver, string(host), string(port))
+    s = Server(sslconfig, tcpserver, string(host), string(port), on_shutdown)
     return listenloop(f, s, tcpisvalid, connection_count,
                          reuse_limit, readtimeout, verbose)
 end
--- a/test/server.jl
+++ b/test/server.jl
@@ -168,4 +168,28 @@ end
     @test_skip HTTP.hasheader(HTTP.get("http://httpbin.org/redirect-to?url=https://httpbin.org/response-headers?Authorization=auth"), "Authorization")
 end # @testset
 
+@testset "on_shutdown" begin
+    @test HTTP.Servers.shutdown(nothing) === nothing
+
+    IOserver = Sockets.listen(Sockets.InetAddr(parse(IPAddr, "127.0.0.1"), 8052))
+    
+    # Shutdown adds 1
+    TEST_COUNT = Ref(0)
+    shutdown_add() = TEST_COUNT[] += 1
+    server = HTTP.Servers.Server(nothing, IOserver, "host", "port", shutdown_add)
+    close(server)
+
+    # Shutdown adds 1, performed twice
+    @test TEST_COUNT[] == 1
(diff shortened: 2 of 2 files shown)


src/Servers.jl
```diff
@@ -69,15 +69,20 @@ function check_rate_limit(tcp, rate_limit::Rational{Int})
 end
 
 "Convenience object for passing around server details"
-struct Server{S <: Union{SSLConfig, Nothing}, I <: Base.IOServer}
+struct Server{S <: Union{SSLConfig, Nothing}, I <: Base.IOServer, F <: Union{Function, Vector{Function}, Nothing}}
     ssl::S
     server::I
     hostname::String
     hostport::String
+    on_shutdown::F
```
fredrikekre, line 77: It doesn't seem critical to parametrize on this so perhaps just
```suggestion
    on_shutdown::Any
```
would be just as good?

src/Servers.jl
```diff
@@ -69,15 +69,20 @@ function check_rate_limit(tcp, rate_limit::Rational{Int})
...
 end
 
 Base.isopen(s::Server) = isopen(s.server)
-Base.close(s::Server) = close(s.server)
+Base.close(s::Server) = (shutdown(s.on_shutdown); close(s.server))
```
fredrikekre, line 81: Would it be useful to pass `s` to the shutdown functions?

src/Servers.jl
```diff
@@ -69,15 +69,20 @@ function check_rate_limit(tcp, rate_limit::Rational{Int})
...
+
+shutdown(fns::Vector{Function}) = shutdown.(fns)
```
fredrikekre, line 83:
```suggestion
shutdown(fns::Vector{Function}) = foreach(shutdown, fns)
```

src/Servers.jl
```diff
@@ -69,15 +69,20 @@ function check_rate_limit(tcp, rate_limit::Rational{Int})
...
+shutdown(fn::Function) = fn()
+shutdown(::Nothing) = ()
```
fredrikekre, line 85:
```suggestion
shutdown(::Nothing) = nothing
```

src/Servers.jl
mmiller-max: Yep, makes sense!

src/Servers.jl
mmiller-max: If there's a possibility we might want to do other (not specified by the user) shutdown tasks that needed the other fields this would make sense? If not I think it's cleaner as it is for now, and it would always be easy to add it in the future with another method for `shutdown`.

src/Servers.jl
mmiller-max: Nice, didn't know about this function, thanks!

src/Servers.jl
fredrikekre: It was just a thought that maybe you need the server object for shutting down your own stuff too, but yea can always be added later by checking `applicable` I think.

src/Servers.jl
fredrikekre: The reason is that `shutdown.(fns)` will allocate an array for the results, but you just want to call them and discard the result.

src/Servers.jl
mmiller-max: Oh yeah `applicable` would let you do that, good idea. It's probably worth having the extra functionality in? I'll add it along with some tests and documentation.

fredrikekre: I don't have a usecase for it from top of my head, so maybe it can wait. Let's see what others have to say :)

src/Servers.jl
mmiller-max: Yeah that makes sense, I've used broadcasting and been slightly frustrated by the allocation before, so good to know about `foreach`

mmiller-max:
Added `try`/`catch` to stop an erring shutdown function stopping the server from closing.

I've realised that the shutdown functions won't run if an `IOServer` object that has been supplied to `listen` is closed, as `close(s::Server)` will never be called. However if someone is manually closing their server then they can probably do anything they need to do before it.

test/server.jl
```diff
@@ -168,4 +168,18 @@ end
...
+    # shutdown unit tests
+    @test HTTP.Servers.shutdown(nothing) === nothing
+    a = 0
+    shutdown_add() = a += 1
+    HTTP.Servers.shutdown(shutdown_add)
+    @test a == 1
+    HTTP.Servers.shutdown([shutdown_add, shutdown_add])
+    @test a == 3
+    shutdown_throw() = throw(ErrorException("Broken"))
+    @test_logs (:error, r"hutdown function shutdown_throw failed") HTTP.Servers.shutdown([shutdown_throw, shutdown_add])
```
mmiller-max, line 181:
```suggestion
    @test_logs (:error, r"shutdown function shutdown_throw failed") HTTP.Servers.shutdown([shutdown_throw, shutdown_add])
```

mmiller-max: Also looks like this test fails on 1.0

fredrikekre: The exact printing of function can change so perhaps just
```suggestion
    @test_logs (:error, r"hutdown function .* failed") HTTP.Servers.shutdown([shutdown_throw, shutdown_add])
```

mmiller-max: Yep that's nice, thanks

mmiller-max:
I've modified the tests to actually test `close` rather than `shutdown` itself. I'm not sure why the Labeler is failing - do I need to squash?

Apart from that I think this is ready.

mmiller-max:
Squash didn't help with the labeler. Any help with this would be appreciated

mmiller-max:
Just bumping.

src/Servers.jl
bauglir: In the past, I have built similar functionality in a wrapper around `Server` in a custom framework. I'm not sure if this use case would still be applicable here, but I used the shutdown functionality there to correctly close any open WebSockets instead of just destroying them (which can leave browsers confused).

quinnj (approved):
Sorry for the slow review; this looks good to me.

mmiller-max:
@quinnj no worries, thanks for reviewing and merging!

src/Servers.jl
mmiller-max: Yeah that sounds like a good use case for this - are WebSockets destroyed when `close(s.server)` is performed?

bauglir: Yes, just prior to it. The `stop` function for the wrapper, which keeps track of open WebSockets, simply iterates over all the WebSockets that are still open, initiates closing them, waits for them to actually close and then calls `close(wrapper.server)`.

Quite happy to see these types of changes making it into HTTP.jl. The main reason for this wrapper to exist, is for purposes such as this (and #602 which has also been fixed). Thanks for putting in the work!

PR #608, opened 2020-10-29
Add functions for setting default custom layers

christopher-dG:
Some motivation for this: [BrokenRecord.jl](https://github.com/JuliaTesting/BrokenRecord.jl) wants to add some behaviour to all calls to `HTTP.request`, which sounds like a good case for a custom layer, but since it is not in control of where `HTTP.request` is called, it can't insert a modified stack into all those calls. Instead, it uses Cassette to achieve something similar, and it works, but it can be horribly slow. With the ability to add default layers, BrokenRecord would no longer need Cassette and would be 500000x faster and more reliable.

Naming for these functions is open for discussion, I didn't think too hard about it.
I tried to keep the API similar to the existing `insert` function.
If something like this is actually wanted, I'll write some documentation for it.

- Running a test suite with currently released BrokenRecord: 143s 😴 
- Same test suite with new HTTP + BrokenRecord: 28s 🔥

### DIFF
--- a/src/layers.jl
+++ b/src/layers.jl
@@ -1,5 +1,7 @@
 module Layers
-export Layer, next, top_layer, insert
+export Layer, next, top_layer, insert, insert_default!, remove_default!
+
+const EXTRA_LAYERS = Set{Tuple{Union{UnionAll, Type{Union{}}}, UnionAll}}()
 
 include("exceptions.jl")
 
@@ -107,5 +109,11 @@ function insert(stack::Type{<:Layer}, layer_before::Type{<:Layer}, custom_layer:
     end
     throw(LayerNotFoundException("$layer_before not found in $stack"))
 end
+
+insert_default!(before::Type{<:Layer}, custom_layer::Type{<:Layer}) =
+    push!(EXTRA_LAYERS, (before, custom_layer))
+
+remove_default!(before::Type{<:Layer}, custom_layer::Type{<:Layer}) =
+    delete!(EXTRA_LAYERS, (before, custom_layer))
 
 end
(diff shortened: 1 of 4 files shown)


src/layers.jl
```diff
@@ -1,5 +1,7 @@
 module Layers
-export Layer, next, top_layer, insert
+export Layer, next, top_layer, insert, insert_default!, remove_default!
+
+const EXTRA_LAYERS = Set{Tuple{Union{UnionAll, Type{Union{}}}, UnionAll}}()
```
quinnj, line 4: Hmmmm.....I don't quite understand the eltype of the `Set` here; why all the Tuple/Union/UnionAll dance?

christopher-dG: I was trying to keep all the type parameters as concrete as possible. The `Set` elements are `Tuple`s with two elements: the first is the before layer, which can be either a `Type{<:Layer}` or `Type{Union{}}` (in the case that we insert a layer as the last layer), and the second is the custom layer, which is always a `Type{<:Layer}`. Since layer types are always parametric, a non-parameterized layer type is a `UnionAll`. Does that make sense? Maybe it's totally unnecessary to parametrize the `Set` like this.

oxinabox:
When I saw BrokeRecord.jl I thought it was a good use case for a custom layer and was surprised that it instead used cassette.

christopher-dG:
Yeah, it's not possible (AFAIK). If some transitive dependency calls `HTTP.get`, there's no way to put a custom layer in there.

christopher-dG:
Any more thoughts on this @quinnj?

quinnj (approved):
Yeah, I don't love this approach, but I think it's fine w/ our current setup. I've started mulling over ideas for removing the strongly typed layer system for something a little simpler; like perhaps using an array of "layer functions" (which would work similarly to the approach in this PR), or something along those lines.

findmyway:
This feature is quite useful.

One thing I do not understand is, why is a `Set` used here, isn't the order guaranteed?

PR #609, opened 2020-10-29
URIs: Don't output ? for empty queries

fonsp:
Before:
```julia
julia> HTTP.URI(scheme="http", host="a.net", query=[]) |> string
"http://a.net?"
```
After:
```julia
julia> HTTP.URI(scheme="http", host="a.net", query=[]) |> string
"http://a.net"
```


quinnj:
Thanks @fonsp! If you wouldn't mind, could you push this same change/test to https://github.com/JuliaWeb/URIs.jl? We're at a slightly awkward in-between state of splitting the URIs module out from HTTP.jl.

fonsp:
Sure! https://github.com/JuliaWeb/URIs.jl/pull/5

PR #613, opened 2020-11-01
Allow passing in a custom boundary to HTTP.Form

christopher-dG:
This one is also motivated by BrokenRecord. If you want your HTTP request body to be deterministic, you need to be able to set the multipart boundary yourself.

### DIFF
--- a/src/multipart.jl
+++ b/src/multipart.jl
@@ -53,9 +53,8 @@ function Base.read(f::Form, n::Integer)
     return result
 end
 
-function Form(d)
+function Form(d; boundary=string(rand(UInt128), base=16))
     @require eltype(d) <: Pair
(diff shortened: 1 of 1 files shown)


aviks (approved):

src/multipart.jl
```diff
@@ -53,9 +53,8 @@ function Base.read(f::Form, n::Integer)
     return result
 end
 
-function Form(d)
+function Form(d; boundary=string(rand(UInt128), base=16))
```
quinnj, line 56: If we're allowing people to pass in a boundary, it might be helpful to assert they pass a string of the right length

christopher-dG: Good idea, I can open another PR in the morning.

PR #616, opened 2020-11-03
Remove internal URIs module and depend on URIs.jl instead

### LINKED ISSUE #605: Splitting out URI parsing into URIs.jl
Hi all,

As mentioned in https://github.com/JuliaWeb/URIParser.jl/issues/71, I've extracted the code related to URI parsing into its own library so that it can be used independently from HTTP.jl

This has resulted in https://github.com/JuliaWeb/URIs.jl, where I've tried to keep all git history relevant to the parser. Also I've copied the attribution and license etc from HTTP.jl.

URIs.jl seems to work, but I thought I'd open this issue to make people aware of the split and to discuss further action we should take as it relates to HTTP.jl. Presumably HTTP should be made to depend on URIs to prevent the codebases from diverging.

Thoughts?

c42f:
Everything from the internal URIs module has been moved to URIs.jl, with
the exception of the resource() function which is HTTP specific.
(request-target defined in HTTP/1.1
https://tools.ietf.org/html/rfc7230#section-5.3).

This was moved into MessageRequest.jl as it appears it's only used in
implementing `request()`. Currently it's also available as `HTTP.resource()`
in case somebody was relying on it.

See also https://github.com/JuliaWeb/URIs.jl/issues/2

Closes #605


c42f:
I thought we should get this done to avoid any more duplicate work as in
#609
https://github.com/JuliaWeb/URIs.jl/pull/5

quinnj (approved):
Thanks for doing this!

PR #621, opened 2020-11-12
Fix tests for HTTP v0.9.0.

fredrikekre:
HTTP.post(uri, body) was supposed to be deprecated for 0.9.0,
but now that deprecations are silent by default it can be
deprecated now in 0.9.1. Also rewrites the tests for this
such that they don't start to fail again with HTTP 0.10.0
(or 1.0.0).


quinnj (approved):
Thanks for doing this

PR #624, opened 2020-11-20
Improve error logging display

fonsp:
The `exception` argument to `@error` is currently treated as the error message, not as a keyword argument, because it is the first argument. Example:

image


quinnj (approved):
Thanks for this!

PR #625, opened 2020-11-20
Only print -4077 error when verbose

fonsp:
Pluto uses HTTP.jl, and it prints the following error message very often:

image

I have not investigated this error thoroughly (little HTTP knowledge), but my observations are:
- It happens when firefox navigates away from a HTTP.jl-served page that has an open WS connection. It does not happen on Chrome, it does not happen on pages with no WS connection.
- **The error can safely be ignored**.
- The error always has the `-4077` code.

I am submitting this PR because:
- Many Pluto users are confused by this error. It makes Pluto look unstable, and users often post this error in unrelated bug reports.
- There is no way for a package using HTTP.jl to prevent this error from printing.
- The existing code suggests that a similar situation occurs with the `-54` error. I am assuming that this `-4077` error is similar, but might not have been found before because the WS part of HTTP.jl is less commonly used.

What do you think?

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -290,7 +290,7 @@ function listenloop(f, server, tcpisvalid, connection_count,
                 handle_connection(f, conn, server, reuse_limit, readtimeout)
                 # verbose && @info "Closed ($count):  $conn"
             catch e
-                if e isa Base.IOError && e.code == -54
+                if e isa Base.IOError && (e.code == -54 || e.code == -4077)
                     verbose && @warn "connection reset by peer (ECONNRESET)"
                 else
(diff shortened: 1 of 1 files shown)


fredrikekre:
The best way to protect users from internals like this is probably to run your application with a custom logger that can filter however you like. This PR might make sense anyway though, I have no idea.

quinnj (commented):
Yes, this is fine and documented in libuv as the error code for connection reset: https://github.com/libuv/libuv/blob/fc2c1a2341c0bedfca3bded96ac7dce8b3575b53/include/uv/errno.h#L132. I actually can't quite remember/track down where the 54 code came from, but I wonder if that was an old version of this error code.

PR #634, opened 2020-12-01
Allow other string types in the cookie dictionary

cmcaine:
Otherwise users have to convert their SubStrings to String, for example.


cmcaine:
`struct Cookie` in `src/cookies.jl` doesn't need to be changed because julia will use the `convert` machinery.

quinnj (approved):
LGTM

PR #635, opened 2020-12-02
Flush TCP connections during `closewrite()`

staticfloat:
TCP connections are generally (depending on kernel whimsy) governed by
the Nagle algorithm, which may not immediately transmit data but instead
wait for the TCP buffer to fill before spitting a packet out onto the
network.  This is due to the fact that the kernel does not necessarily
know when a "message" is complete and simply does its best to maximize
throughput (by decrease packet counts) while minimizing latency (by
transmitting quickly).  We can help the OS out a bit here by flushing
the TCP buffer when we're done with a message (e.g. in `closewrite()`)
which (on some kernels) reduces the minimum latency of a message from
~40ms to < 1ms.  We flush by toggling the Nagle algorithm off and then
on again on that particular socket, forcing the OS to transmit any
pieces of data that were queued to be sent out of that socket
immediately.

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -186,6 +186,23 @@ writebusy(t::Transaction) = @v1_3 lock(() -> t.writebusy, t.c.writelock) t.write
 readbusy(t::Transaction) = @v1_3 lock(() -> t.readbusy, t.c.readlock) t.readbusy
 
 """
+    flush(c::Connection)
+
+Flush a TCP buffer by toggling the Nagle algorithm off and on again for a socket.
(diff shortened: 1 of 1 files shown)


staticfloat:
@quinnj I'm not entirely sure why uninitialized sockets can make their way into the `closewrite()` method, but I'm guarding `flush()` against trying to set `nagle()` on those sockets as well.

quinnj:
Do we know when `nagle` was added? Looks like the 1.0 builds aren't passing because they need a version check

quinnj (approved):

Issue #631, opened 2020-11-27
Improperly-handled HTTP responses cause connection freezes


staticfloat:
If an HTTP server does not follow the proper flow, clients may be stuck waiting for a response that never comes.  Example improperly-written server:

```julia
# http_slippage.jl
using HTTP, Sockets

function serve(host, port)
    num_requests = 1
    @info("Listening on $(host):$(port)")
    HTTP.listen(host, port) do http
		# Decrease server latency
        Sockets.nagle(http.stream.c.io, false)

        # Once every hundred requests, accidentally don't set any status or start writing or whatnot.
        if num_requests%100 != 0
            # By default, we return the URI back to the user
            HTTP.setstatus(http, 200)
            HTTP.startwrite(http)
            write(http, http.message.target)
        end

        if num_requests%10 == 0
            println("Served $(num_requests) requests")
        end
        num_requests += 1
    end
end

serve("0.0.0.0", 8000)
```

k6 tester:
```javascript
// slippage_test.js
// If you don't want to install `k6`, you can run this via:
// docker run -v $(pwd):/config:ro --net=host -ti loadimpact/k6 run /config/slippage_test.js
import http from 'k6/http';
import { check } from 'k6';

// Simulate 10 users for 5 minutes
export let options = {
    vus: 10,
    duration: '5m',
    // Un-comment this to eliminate connection freezing, but suffer a performance impact
    // noConnectionReuse: true,
};

// Get the root page, /registries, and follow the /registries redirect
export default function () {
    // Hit the index page (static file)
    let resource = '/data/' + Math.random();
    let res = http.get('http://localhost:8000' + resource);
    check(res, {
        '200 OK': (r) => r.status === 200,
    });
    check(res, {
        'URI reflected': (r) => r.body === resource,
    })
}
```

Running this shows good performance for the first few milliseconds, followed by _all ten_ virtual users getting stuck after the 1000th request (presumably beacause each virtual user hits a different 100th request and gets stuck there).  They freeze for 60s, after which they are freed and continue consuming, due to `k6` timing out the connection.

It is worth noting that this freezing behavior is fixable by configuring `k6` to not re-use connections, but that has a performance impact and isn't how many clients (such as web browsers and `nginx` when serving as a reverse-proxy) like to work.

I propose that `HTTP.jl` detect when the callback is poorly-behaved, perhaps by checking to see if no status has been set, and if so, it sets a default such as `HTTP 500`.

quinnj:
[...]

PR #636, opened 2020-12-03
Handle poorly-behaved callbacks more reliably

staticfloat:
If `startwrite()` is never called during a server callback, the client
may be left waiting for a response and never get one.  This can cause
lockups and general misery, especially when connections are proxied
through an intermediary (such as `nginx`) with a finite connection
limit.

This PR detects a server callback that never calls `startwrite(http)`
and throws an error, triggering the HTTP 500 mechanism.  It also adds a
`closewrite(http)` onto that branch as without it, it appears that the
final chunk in the `Transfer-Encoding: Chunked` response may not appear.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -397,6 +397,11 @@ function handle_transaction(f, t::Transaction, server; final_transaction::Bool=f
 
     @async try
         f(http)
+
+        # If `startwrite()` was never called, throw an error so we send a 500 and log this
+        if isopen(http) && !iswritable(http)
+            error("Server never wrote a response")
+        end
         @debug 2 "server closeread"
         closeread(http)
         @debug 2 "server closewrite"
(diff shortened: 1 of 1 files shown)


staticfloat:
@quinnj is this the best way to detect a lack of `startwrite(http)` in the callback?  I was halfway to adding an `http.write_started` flag that `startwrite()` would set, but then noticed that `write()` will conditionally call `startwrite()` automatically if `!iswritable(http)` is true, so I figured I'd follow that lead.

Also, I found that I needed to add a `closewrite(http)` to the error branch here to get the whole error message transmitted when chunked encoding is in effect.  I'm not sure if I need to `closeread()` as well.

staticfloat:
Closes https://github.com/JuliaWeb/HTTP.jl/issues/631

quinnj:
Nightly failures look unrelated; some Pkg error. This looks great to me; can I make one small request? Would you mind adding a big `*NOTE:` explanation of what's required in your own Stream handler to the docs [here](https://github.com/JuliaWeb/HTTP.jl/blob/6806422439ee99269569c32714f64d6e13cf8423/src/Servers.jl#L123)? I think that could possibly help the next poor traveler who stumbles along the way. Basically just that they should call `startwrite`/`closewrite` when handling their stream.

quinnj (approved):

PR #640, opened 2020-12-08
set userinfo from http(s)_proxy to proxy-auth

p1scescom:
add a process to set proxy-authorization basic like src/BasicAuthRequest.jl if proxy-url has userinfo
I confirmed that it works in my proxy environment.


quinnj (approved):
Seems fine to me

PR #641, opened 2020-12-08
Document connect_timeout and set it to the nonzero minimum of itself and readtimeout

sean-bennett112:
This is to make it the same format as `readtimeout`.

Also changes `connecttimeout` to be the nonzero minimum of `connecttimeout` and `readtimeout` if either is set.

This is to get around connections hanging even if `readtimeout` is set. For example, this occurs if you try to connect to the EC2 metadata IP from outside AWS.

I'm not sure of a great way to add a test for this. This works, for example:

```julia
@test_throws Exception HTTP.get("http://169.254.169.254/latest/meta-data/iam/info"; connectiontimeout=1, retry=false)
```

But I'd obviously prefer to avoid trying to hit the EC2 metadata service.

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -640,7 +640,7 @@ function getconnection(::Type{TCPSocket},
                        host::AbstractString,
                        port::AbstractString;
                        keepalive::Bool=false,
-                       connecttimeout::Int=0,
+                       connect_timeout::Int=0,
                        readtimeout::Int=0,
                        kw...)::TCPSocket
 
@@ -648,9 +648,9 @@ function getconnection(::Type{TCPSocket},
 
     @debug 2 "TCP connect: $host:$p..."
 
-    timeouts = filter(!iszero, [connecttimeout, readtimeout])
-    connecttimeout = isempty(timeouts) ? 0 : minimum(timeouts)
-    if connecttimeout == 0
+    timeouts = filter(!iszero, [connect_timeout, readtimeout])
+    connect_timeout = isempty(timeouts) ? 0 : minimum(timeouts)
+    if connect_timeout == 0
         tcp = Sockets.connect(host == "localhost" ? ip"127.0.0.1" : Sockets.getalladdrinfo(host)[1], p)
         keepalive && keepalive!(tcp)
         return tcp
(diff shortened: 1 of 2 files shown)


src/ConnectionPool.jl
```diff
@@ -640,14 +640,17 @@ function getconnection(::Type{TCPSocket},
                        host::AbstractString,
                        port::AbstractString;
                        keepalive::Bool=false,
-                       connect_timeout::Int=0,
+                       connecttimeout::Int=0,
```
mattBrzezinski, line 643: Even thought it's not documented, I think this needs to remain `connect_timeout` otherwise it's a breaking change.

src/ConnectionPool.jl
```diff
@@ -641,12 +641,15 @@ function getconnection(::Type{TCPSocket},
...
                        connect_timeout::Int=0,
+                       readtimeout::Int=0,
                        kw...)::TCPSocket
 
     p::UInt = isempty(port) ? UInt(80) : parse(UInt, port)
 
     @debug 2 "TCP connect: $host:$p..."
 
+    timeouts = filter(!iszero, [connect_timeout, readtimeout])
+    connect_timeout = isempty(timeouts) ? 0 : minimum(timeouts)
```
fredrikekre, line 652: Couldn't this be just
```suggestion
    connect_timeout = min(connect_timeout, readtimeout)
```

src/ConnectionPool.jl
sean-bennett112: Makes sense, fixed.

src/ConnectionPool.jl
mattBrzezinski: No, because we want the smallest non-zero between the two. It could be:

```julia
connect_timeout = minimum(filter(!iszero, [connect_timeout, readtimeout]))
```

sean-bennett112: I don't believe so. If either `connect_timeout` or `readtimeout` is 0, then that will return 0, meaning no connection timeout will be set.

quinnj: How about just:
```julia
connect_timeout = connect_timeout == 0 && readtimeout > 0 ? readtimeout : connect_timeout
```
That avoids allocating an unnecessary array and the need to call filter and minimum.

sean-bennett112: Nice, I like that better, fixed.

Issue #662, opened 2021-01-07
Obtain status and headers for in-flight request


fredrikekre:
I have some code similar to this:
```julia
function proxy_pass(http::HTTP.Stream, upstream::URIs.URI)
    buffer = Base.BufferStream()
    # Make the request
    req = HTTP.request("GET", upstream, http.message.headers;
                       response_stream=buffer, status_exception=false)
    # Write response back to http
    HTTP.setstatus(http, req.status)
    foreach(h -> HTTP.setheader(http, h), req.headers)
    HTTP.startwrite(http)
    HTTP.write(http, buffer)
    return nothing
end
```
i.e. I make a request and want to stream the response directly back to a `HTTP.Stream`. To do this efficiently I want to use a buffer for the response body, and then read from that directly back to `http`. The problem here is that I need the statuscode and headers from the request before `HTTP.startwrite` so I can't just `@async` the request since I would need to `fetch` that task directly for the status and headers.

Is there a good way to achieve this?

cc @staticfloat

fredrikekre:
Perhaps one can implement something similar to the `Channel` interface which lets you pass a `taskref` for obtaining the task.
```julia
function proxy_pass(http::HTTP.Stream, upstream::URIs.URI)
    buffer = Base.BufferStream()
    headerref = HTTP.headerref()
    @sync begin
        # Make the request
        @async HTTP.request("GET", upstream, http.message.headers;
                           response_stream=buffer, status_exception=false, headerref=headerref)
        # Write response back to http
        status, headers = fetch(headerref) # This will fetch as soon as headers are available and let us keep working
        HTTP.setstatus(http, status)
        foreach(h -> HTTP.setheader(http, h), headers)
        HTTP.startwrite(http)
        @async HTTP.write(http, buffer)
    end
    return nothing
end
```

quinnj:
Hmmmm, I'm not quite following the goal/intent here; I'm a pretty visual learner, is there a way you could sketch what is going where real quick?

fredrikekre:
Here is a proof-of-concept (`proxy3`). It is ona branch of my fork
```julia
# using Pkg
# Pkg.add(; url="https://github.com/fredrikekre/HTTP.jl.git", rev="fe/status_headers")
```
to install and then I ran this script (with `--threads=4`).
```julia
using SimpleBufferStream, HTTP, BenchmarkTools

# Cache a big file
[...]

PR #665, opened 2021-01-13
Implement HTTP.StatusHeaders as a "partial" Response

fredrikekre:
This allows for "previewing" the status code and
headers from an in-flight asynchronous HTTP request.

Example:
```
sh = HTTP.StatusHeaders()
req = @async HTTP.get("https://github.com", status_headers=sh)

fetch(sh) # Blocks until status and headers are ready
# Do something with status/headers while body is still transmitting

fetch(req) # Blocks until body is ready
```

Fixes https://github.com/JuliaWeb/HTTP.jl/issues/662.

### DIFF
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -62,7 +62,7 @@ export Message, Request, Response, HeaderSizeError,
        readchunksize,
        writeheaders, writestartline,
        bodylength, unknown_length,
-       payload
+       payload, StatusHeaders
 
 import ..HTTP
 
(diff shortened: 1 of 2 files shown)


fredrikekre:
cc @staticfloat

staticfloat:
Needs tests, but looks good to me!  I think it would be good to test getting the status/headers of both a redirect and the eventual target of the redirect.

quinnj:
Do we still need this for some reason, given the alternative solution proposed in #662?

fredrikekre:
Probably not; this is essentially the `startwrite` call from https://github.com/JuliaWeb/HTTP.jl/issues/662#issuecomment-759657721

PR #666, opened 2021-01-25
Add Accept: */* as default header

### LINKED ISSUE #661: Include Accept: */* in headers by default
This is the default behavior of `curl`, `libcurl`, and Python's [requests](https://github.com/psf/requests/blob/c2b307dbefe21177af03f9feb37181a89a799fcc/requests/utils.py#L824).  While [RFC7231 ](https://tools.ietf.org/html/rfc7231#section-5.3.2) suggests that sites should treat a missing accept header as `*/*` that's not always true in practice.

quinnj:
Fixes #661. This matches what curl/requests do in terms of default
headers.

### DIFF
--- a/src/MessageRequest.jl
+++ b/src/MessageRequest.jl
@@ -33,6 +33,7 @@ function request(::Type{MessageLayer{Next}},
                  parent=nothing, iofunction=nothing, kw...) where Next
 
     defaultheader!(headers, "Host" => url.host)
+    defaultheader!(headers, "Accept" => "*/*")
     if USER_AGENT[] !== nothing
         defaultheader!(headers, "User-Agent" => USER_AGENT[])
     end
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Curl and requests also add a Content-Type header by default, according to https://discourse.julialang.org/t/help-translating-curl-command-to-http-jl/53790/22

Should we do that as well?

quinnj:
We do this when the body is a `Form` (https://github.com/JuliaWeb/HTTP.jl/blob/cf438e9cedfed9730a85562a1f9551e7aaaa4c0a/src/MessageRequest.jl#L50). I think we also used to do this more, but it caused issues or something, so we took it out. I'll have to try and remember where the discussion was.

mbauman:
Ah, yeah, Requests has just a bit more logic there for url encoded parameters.  In any case — this is something that is different enough (it's not unconditional) that it should be done separately.

PR #674, opened 2021-02-16
Proxy environment variables are only used when they are non-empty

### LINKED ISSUE #671: empty HTTP_PROXY and HTTPS_PROXY are still used
Julia 1.5.3 
HTTP.jl 0.9.3 

this is a bit of question:
now if there are HTTP_PROXY="" and HTTPS_PROXY="", they are still used. 
Should it be like this or should it treat empty variables as not being set? 
I think it should, because lots of software does not differentiate between empty env var and env var not being set, and sometimes it's easier to set env var to empty string than unset it.

If this is considered a bug, I can make PR which will fix it.

racinmat:
Motivation: 
Once environment variable is set, it's quite hard to unset it, e.g. in docker it's impossible to unset environment variable once it's set, see: https://stackoverflow.com/questions/55789409/how-to-unset-env-in-dockerfile

Lots of other software treat empty environment variable in same way as unset environment variable. 
For example here https://docs.pivotal.io/application-service/2-8/operating/config-proxy.html in troubleshooting section.
Also curl states in https://curl.se/libcurl/c/CURLOPT_PROXY.html
"Setting the proxy string to "" (an empty string) will explicitly disable the use of a proxy, even if there is an environment variable set for it."

Git mentions them too in https://git-scm.com/docs/git-config/2.9.5#Documentation/git-config.txt-httpproxy , linking curl documentation, so I assume the behavior is same as in curl.

I'm not sure how exactly it's in Pkg, but for correct functionality of Pkg.jl behind proxy setting them is advised too, e.g. here: https://discourse.julialang.org/t/install-packages-behind-the-proxy/23298/18 but I'm not sure if Pkg.jl handles them on its own or delegates it to git, which delegates it to curl.

Fixes #671 . 

I'm not sure if this is desired behavior or not, so let's discuss this if we want to have such behavior or if it's better to keep the old one.

I wanted to also add tests for this, but I couldn't find a way to call some method with environment variable set without affecting it globally. 

Or should I simply add tests with
```julia
ENV["HTTP_PROXY"] = ""
@test getproxy.....
delete!(ENV, "HTTP_PROXY")
```
?

### DIFF
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -34,13 +34,13 @@ end
 
 function getproxy(scheme, host)
     isnoproxy(host) && return nothing
-    if scheme == "http" && haskey(ENV, "http_proxy")
+    if scheme == "http" && haskey(ENV, "http_proxy") && !isempty(ENV["http_proxy"])
         return ENV["http_proxy"]
-    elseif scheme == "http" && haskey(ENV, "HTTP_PROXY")
+    elseif scheme == "http" && haskey(ENV, "HTTP_PROXY") && !isempty(ENV["HTTP_PROXY"])
         return ENV["HTTP_PROXY"]
-    elseif scheme == "https" && haskey(ENV, "https_proxy")
+    elseif scheme == "https" && haskey(ENV, "https_proxy") && !isempty(ENV["https_proxy"])
         return ENV["https_proxy"]
-    elseif scheme == "https" && haskey(ENV, "HTTPS_PROXY")
+    elseif scheme == "https" && haskey(ENV, "HTTPS_PROXY") && !isempty(ENV["HTTPS_PROXY"])
         return ENV["HTTPS_PROXY"]
     end
     return nothing
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -36,12 +36,12 @@ function getproxy(scheme, host)
     isnoproxy(host) && return nothing
     if scheme == "http" && (p = get(ENV, "http_proxy", ""); !isempty(p))
         return p
-    elseif scheme == "http" && !isempty(get(ENV, "HTTP_PROXY", ""))
-        return ENV["HTTP_PROXY"]
-    elseif scheme == "https" && !isempty(get(ENV, "https_proxy", ""))
-        return ENV["https_proxy"]
-    elseif scheme == "https" && !isempty(get(ENV, "HTTPS_PROXY", ""))
-        return ENV["HTTPS_PROXY"]
+    elseif scheme == "http" && (p = get(ENV, "HTTP_PROXY", ""); !isempty(p))
+        return p
+    elseif scheme == "https" && (p = get(ENV, "https_proxy", ""); !isempty(p))
+        return p
+    elseif scheme == "https" && (p = get(ENV, "HTTPS_PROXY", ""); !isempty(p))
+        return p
     end
     return nothing
 end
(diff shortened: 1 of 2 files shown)


quinnj:
There's the `withenv` function:
```julia
help?> withenv
search: withenv

  withenv(f::Function, kv::Pair...)

  Execute f in an environment that is temporarily modified (not replaced as in setenv) by zero or more "var"=>val arguments kv. withenv is
  generally used via the withenv(kv...) do ... end syntax. A value of nothing can be used to temporarily unset an environment variable (if it
  is set). When withenv returns, the original environment has been restored.
```

src/ConnectionRequest.jl
```diff
@@ -34,13 +34,13 @@ end
 
 function getproxy(scheme, host)
     isnoproxy(host) && return nothing
-    if scheme == "http" && haskey(ENV, "http_proxy")
+    if scheme == "http" && haskey(ENV, "http_proxy") && !isempty(ENV["http_proxy"])
```
christopher-dG, line 37:
```suggestion
    if scheme == "http" && !isempty(get(ENV, "http_proxy", ""))
```

You could do this with all of them, if you want (same behaviour)

christopher-dG (commented):
This seems like a fine idea to me, although I'm not super familiar with how proxies work.

racinmat:
Added tests, I hope it's a right place, I can move them somewhere else if it's needed.

src/ConnectionRequest.jl
```diff
@@ -34,13 +34,13 @@ end
 
 function getproxy(scheme, host)
     isnoproxy(host) && return nothing
-    if scheme == "http" && haskey(ENV, "http_proxy")
+    if scheme == "http" && !isempty(get(ENV, "http_proxy", ""))
         return ENV["http_proxy"]
```
fredrikekre, line 38: If you want to avoid two dict-lookups:
```suggestion
    if scheme == "http" && (p = get(ENV, "http_proxy", ""); !isempty(p))
        return p
```

racinmat:
Something more, or is it ready for merge? Tests are failing on nightly, but they seem totally unrelated to my PR.

racinmat:
@christopher-dG  @fredrikekre

test/utils.jl
```diff
@@ -35,4 +35,33 @@ import HTTP.URIs
         )
         @test HTTP.Strings.iso8859_1_to_utf8(bytes) == utf8
     end
-end # testset
\ No newline at end of file
+
+    @test HTTP.ConnectionRequest.getproxy("https", "https://julialang.org/") == nothing
```
fredrikekre, line 39:
```suggestion
    withenv("HTTPS_PROXY"=>nothing, "https_proxy"=>nothing) do
        @test HTTP.ConnectionRequest.getproxy("https", "https://julialang.org/") == nothing
    end
```

test/utils.jl
```diff
@@ -35,4 +35,33 @@ import HTTP.URIs
...
+        @test HTTP.ConnectionRequest.getproxy("https", "https://julialang.org/") == nothing
+    end
+    withenv("HTTPS_PROXY"=>"https://user:pass@server:80") do
+        @test HTTP.ConnectionRequest.getproxy("https", "https://julialang.org/") == "https://user:pass@server:80"
+    end
+    withenv("https_proxy"=>"https://user:pass@server:80") do
+        @test HTTP.ConnectionRequest.getproxy("https", "https://julialang.org/") == "https://user:pass@server:80"
+    end
+
+    @test HTTP.ConnectionRequest.getproxy("http", "http://julialang.org/") == nothing
```
fredrikekre, line 54:
```suggestion
    withenv("HTTP_PROXY"=>nothing, "http_proxy"=>nothing) do
        @test HTTP.ConnectionRequest.getproxy("http", "http://julialang.org/") == nothing
    end
```

fredrikekre (commented):
LGTM (but I don't have the power to take action).

Might be good to reset the env since tests will fail otherwise for users that do have them set (see comments).

Issue #672, opened 2021-02-12
Unexpected behavior with localhost and readtimeout


fredrikekre:
Over in https://github.com/JuliaPackaging/PkgServer.jl/pull/115 there are problems connecting to a server with the combination of `localhost` and `readtimeout`.

Looking at the code I found two things:

1. `localhost` is hardcoded to ipv4, but _**only**_ with `readtimeout=0`: https://github.com/JuliaWeb/HTTP.jl/blob/a469ac66e341a1e8f4e8454b73c7417cff0727e8/src/ConnectionPool.jl#L653 while the other code path uses `getalladdrinfo("localhost")[1]`: https://github.com/JuliaWeb/HTTP.jl/blob/a469ac66e341a1e8f4e8454b73c7417cff0727e8/src/ConnectionPool.jl#L659 so if `getalladdrinfo("localhost")[1]` is the ipv6 address HTTP can't connect.
2. HTTP.jl seems to never care about `getaddrinfo(host)[2:end]`.

I suggest that HTTP.jl should loop over `getalladdrinfo(host)` in both code paths. That is what `curl` seems to do for example (even with `--retry=no` which is the default).

Related: https://github.com/JuliaWeb/HTTP.jl/issues/511, https://github.com/JuliaWeb/HTTP.jl/pull/512, https://github.com/JuliaWeb/HTTP.jl/pull/428

CC: @staticfloat

PR #428, opened 2019-06-24
Connect over IPv6 if the system prefers it

lbilli:

PR #675, opened 2021-02-19
Try all ip addresses when establishing a connection

fredrikekre:
Try all ip addresses when establishing a connection

Fixes https://github.com/JuliaWeb/HTTP.jl/issues/672
Reenables https://github.com/JuliaWeb/HTTP.jl/pull/428

### DIFF
--- a/test/client.jl
+++ b/test/client.jl
@@ -1,5 +1,6 @@
 using ..TestRequest
 using HTTP
+using Sockets
 using JSON
 using Test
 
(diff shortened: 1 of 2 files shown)


test/client.jl
```diff
@@ -215,6 +215,28 @@ end
     @test_throws HTTP.IOError HTTP.get("http://httpbin.org/delay/5"; readtimeout=1, retry=false)
 end
 
+@testset "Retry all resolved IP addresses" begin
+    # See issue https://github.com/JuliaWeb/HTTP.jl/issues/672
+    # Bit tricky to test, but can at least be tested if localhost
+    # resolves to both IPv4 and IPv6 by listening to the respective
+    # interface
+    alladdrs = Sockets.getalladdrinfo("localhost")
+    if Sockets.ip"127.0.0.1" in alladdrs && Sockets.ip"::1" in alladdrs
```
fredrikekre, line 224: This is true on GitHub Actions. Is there a better way of forcing IP adresses that is not the first? Is there an easy way to insert a bad IP for some host?

test/client.jl
```diff
@@ -215,6 +215,28 @@ end
     @test_throws HTTP.IOError HTTP.get("http://httpbin.org/delay/5"; readtimeout=1, retry=false)
 end
 
+@testset "Retry all resolved IP addresses" begin
+    # See issue https://github.com/JuliaWeb/HTTP.jl/issues/672
+    # Bit tricky to test, but can at least be tested if localhost
+    # resolves to both IPv4 and IPv6 by listening to the respective
+    # interface
+    alladdrs = Sockets.getalladdrinfo("localhost")
```
quinnj, line 223: Looks like we need a `using Sockets` at the top of this file

quinnj (approved):

PR #676, opened 2021-02-19
Improved documentation for HTTP.Form and HTTP.Multipart.

fredrikekre:
xref e.g. https://stackoverflow.com/questions/66268254/julia-http-post-request-with-image-attachment

### DIFF
--- a/src/multipart.jl
+++ b/src/multipart.jl
@@ -1,13 +1,4 @@
 # Form request body
-"""
-    Form(dict::Dict)
-
-A type representing a request body using the multipart/form-data encoding.
-The key-value pairs in the Dict argument will constitute the name and value of each multipart boundary chunk.
-Files and other large data arguments can be provided as values as IO arguments: either an `IOStream` such as returned via `open(file)`,
-an `IOBuffer` for in-memory data. For complete control over a multipart chunk's details, an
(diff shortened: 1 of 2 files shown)


src/multipart.jl
```diff
@@ -126,12 +157,11 @@ mutable struct Multipart{T <: IO} <: IO
     data::T
     contenttype::String
     contenttransferencoding::String
-    name::String
 end
 
-function Multipart(f::Union{AbstractString, Nothing}, data::T, ct::AbstractString="", cte::AbstractString="", name::AbstractString="") where {T<:IO}
```
fredrikekre, line 132: `name` seems unused (?) and possibly a relic of old code where `Multipart` was not the value in `Form`?

quinnj: Yeah, that sounds right

quinnj: Oh, but we do have some tests that rely on it, which are failing here

fredrikekre: Yea it seems that parsing a multipart body uses the name field. Perhaps parsing should give back `Vector{Pair{name::String,Multipart}}` instead to mirror how you construct a `HTTP.Form`, but I will leave that for another day and just leave the `name` field here.

PR #682, opened 2021-02-25
undeprecate the accessor methods of a HTTP message and updates the docs

### LINKED ISSUE #681: Documentation about `request` and `response` struct and `getfield` function.
Thank you for great OSS!!.
I found a small enhancement for HTTP.jl documentation.

I think [Current document](https://juliaweb.github.io/HTTP.jl/stable/) does not explain how to get HTTP information from request and response object clearly.

Because the docs of `request` and `response` struct are in internal interfaces section, and the docs does not explain how to get it: [Internal Interfaces · HTTP\.jl](https://juliaweb.github.io/HTTP.jl/stable/internal_interface/#HTTP.Messages.Request)

When I checked the source code, I found `getfield` can be used for it.
https://github.com/JuliaWeb/HTTP.jl/blob/edc5e28e53d33c470930f2284ff2cf0e3f877afe/src/Messages.jl#L227-L230
https://github.com/JuliaWeb/HTTP.jl/blob/edc5e28e53d33c470930f2284ff2cf0e3f877afe/src/Messages.jl#L157-L159

I think it would be better that the docs of request and response are moved to public interface sections and it should explain how to get these data for beginners.

Thanks again.

Julia 1.5.3 
HTTP.jl 0.8.19
MbedTLS.jl 1.0.3

AtsushiSakai:
I undeprecated the accessor methods of a HTTP message and updates the docs to fix #681.

### DIFF
--- a/docs/src/internal_interface.md
+++ b/docs/src/internal_interface.md
@@ -15,8 +15,6 @@ HTTP.Parsers.parse_chunk_size
 ## Messages Interface
 
 ```@docs
-HTTP.Messages.Request
-HTTP.Messages.Response
 HTTP.Messages.iserror
 HTTP.Messages.isredirect
 HTTP.Messages.ischunked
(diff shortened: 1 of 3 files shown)


quinnj (approved):
Thanks @AtsushiSakai!

PR #683, opened 2021-02-26
Removed unnecessary export for HeaderSizeError

AtsushiSakai:
This is a very small PR to remove unnecessary export for HeaderSizeError.
I think `HeaderSizeError` function is not defined in the file, so it is not needed.


quinnj (approved):
Thanks @AtsushiSakai!

PR #686, opened 2021-03-04
add example for a persistent session and logging into a webform

Crghilardi:
I've been working on porting some code over from a co-worker and found myself searching for the equivalent of the various "session"  functions (pythons `requests.Session()` or `rvest::html_session()` in R). I asked about this on slack and it sounds like `HTTP.get(...; cookies = true)` is functionally equivalent.

This PR adds a new example file starting a session and logging into a webform along with a little blurb explaining. Any additions/modifications welcome as I am not a webdev person.


quinnj (approved):
Thanks!

PR #687, opened 2021-03-04
fix minor typos in request docstring

ueliwechsler:
I saw these typos during browsing the documentation.

Not sure if this is the right way to address it though.


quinnj (approved):
Thanks!

PR #692, opened 2021-03-06
Save stream to body if StreamRequest errors

### LINKED ISSUE #691: If StreamRequest fails, received data is lost
Julia 1.5.3
HTTP.jl 0.9.5

When doing a get request with a `response_stream`, if it fails any response is lost. This means that some of the error information can be lost.

For example, when using `AWSS3.jl` to stream a file to a local file, if the file name is wrong the request returns a 404 error but no information (it was in the BufferStream). The stream data actually contains a useful error message saying the file doesn't exist, but this is lost in the exception handling.

A fix would be to read the `response_stream` if the `StatusRequest` has errored and add it to the `response.body`. I'll put up a PR for discussion.

Ref: https://github.com/JuliaCloud/AWS.jl/pull/304

mmiller-max:
I don't know if this behaviour is always desirable, but if so it is useful to add the streamed data to the response body if a `StreamRequest` errors.

If this seems good to people here, I can add some documentation/tests for this.

Closes #691

### DIFF
--- a/src/StreamRequest.jl
+++ b/src/StreamRequest.jl
@@ -66,6 +66,7 @@ function request(::Type{StreamLayer{Next}}, io::IO, req::Request, body;
                 @debug 2 "client startread"
                 startread(http)
                 readbody(http, response, response_stream, reached_redirect_limit)
+                iserror(response) && save_response_stream!(response, response_stream)
             else
                 iofunction(http)
             end
@@ -149,6 +150,15 @@ function readbody(http::Stream, res::Response, response_stream, reached_redirect
             close(response_stream)
         end
     end
+end
+
+function save_response_stream!(response::Response, response_stream)
+    resp = UInt8[]
+    response_stream_copy = deepcopy(response_stream)
+    while !eof(response_stream_copy)
+        append!(resp, readavailable(response_stream_copy))
+    end
+    response.body = resp
 end
 
 end # module StreamRequest


src/StreamRequest.jl
```diff
@@ -151,4 +152,13 @@ function readbody(http::Stream, res::Response, response_stream, reached_redirect
     end
 end
 
+function save_response_stream!(response::Response, response_stream)
+    resp = UInt8[]
+    response_stream_copy = deepcopy(response_stream)
```
quinnj, line 157: why the `deepcopy` here?

quinnj:
It looks like the tests are failing on a method error `eof(::Nothing)`, so we might need an additional guard in there like we have for `readbody`. I'll admit I don't entirely follow the reasoning for why this is needed; specifically, it's not clear why the response body wouldn't be set from `readbody` even if the request resulted in an error? Could you perhaps provide a more extended example of where/why this fails?

mmiller-max:
Thanks for taking a look, yep of course, sorry it wasn't clear.

My use case for this was using `s3_get_file` from the `AWSS3` package. This function uses a `StreamRequest` by passing a `Base.BufferStream` to the `response_stream` kwarg of `HTTP.request`.

When one attempts to use `s3_get_file` to get a file that doesn't exist, AWS returns a 404 and `XML` formatted text saying that the specified key does not exist, and `HTTP.jl` throws the exception due to the 404. However this `XML` content is in the `BufferStream` and not in the message body, which contains `[Message Body was streamed]`. As `HTTP.jl` throws an exception, the `BufferStream` is never returned and the text from AWS is effectively lost. `AWS.jl` doesn't then know the source of the error when it tries to interpret and can only show the user a less useful 404 exception like this:

```Julia
julia> s3_get_file(config, bucket, "//a//","a")
ERROR: AWS.AWSExceptions.AWSException("404", "AWSException", Dict{String,Dict}(), HTTP.ExceptionRequest.StatusError(404, "GET", "/bucket///a//?return_stream=true", HTTP.Messages.Response:
"""
HTTP/1.1 404 Not Found
x-amz-request-id: 
x-amz-id-2: 
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Wed, 10 Mar 2021 03:31:11 GMT
Server: AmazonS3

[Message Body was streamed]"""))
```

With this PR, when the stream request fails, the stream is read to the `body` so that it is not lost and can be used by downstream packages as part of the error reporting. For example, with this PR, the error message for an incorrect file looks like this:

```Julia
julia> s3_get_file(config, bucket, "//a//","a")
ERROR: AWS.AWSExceptions.AWSException("NoSuchKey", "The specified key does not exist.", EzXML.Document(EzXML.Node(<DOCUMENT_NODE@0x00007fd392859650>)), HTTP.ExceptionRequest.StatusError(404, "GET", "/bucket///a//?return_stream=true", HTTP.Messages.Response:
"""
HTTP/1.1 404 Not Found
x-amz-request-id:
x-amz-id-2:
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Wed, 10 Mar 2021 03:33:33 GMT
Server: AmazonS3

<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>//a//</Key><RequestId></RequestId><HostId></HostId></Error>"""))
```

Note, the examples here were performed using the branch of `AWS.jl` from this PR: https://github.com/JuliaCloud/AWS.jl/pull/304

Hope that's clearer.

This may not be the right fix, perhaps it is better than if a user performs a `StreamRequest` and it throws an exception then the user should expect to not get any information and handle it themselves based on the status code. Or maybe this should be optional behaviour? Interested to hear your thoughts on this!

src/StreamRequest.jl
mmiller-max: I didn't want to alter `response_stream` in case it's used elsewhere. But I don't think it is, so could be safe to remove

quinnj:
Thanks for the detailed explanation, that helps explain the approach you're taking here. The main question I have is that if AWS.jl is passing in the Base.BufferStream, it seems like they should be able to read it to get the error response, even in the case of an error? I'm not sure if [this](https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/http.jl) is the right place, but at least from what I'm understanding, the response error is actually in the BufferStream, and it's just a matter of reading it. Let me know if that makes sense.

quinnj:
cc: @mattBrzezinski ^^

mmiller-max:
Oh yeah that completely makes sense, I hadn't thought about how the BufferStream would actually still be available, whoops!  Adding in something similar to this PR [here](https://github.com/JuliaCloud/AWS.jl/blob/b24e9cd54bc5e798919e4898217f5cc51866aa70/src/AWS.jl#L377) seems to do the trick. Will discuss over at `AWS.jl`. Cheers @quinnj 

As for this PR/the related issue, will close them both as I don't think makes sense to include this in HTTP with your solution above. That all good with you?

quinnj:
Yes, that all sounds fine to me

PR #701, opened 2021-04-22
Bugfixes and improvements to rate limit check

fredrikekre:
Bugfixes and improvements to rate limit check

 - use the correct IP adress of the client
 - use `DateTime(0)` for first time connections
 - don't warn when capping the rate limit allowance
 - fix rate_limit specification with non-1 denominator

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -36,6 +36,9 @@ function update!(rl::RateLimit, rate_limit)
     timepassed = float(Dates.value(current - rl.lastcheck)) / 1000.0
     rl.lastcheck = current
     rl.allowance += timepassed * rate_limit
+    if rl.allowance > rate_limit
+        rl.allowance = rate_limit
+    end
     return nothing
 end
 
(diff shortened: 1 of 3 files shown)


src/Servers.jl
```diff
@@ -51,14 +54,9 @@ soon, it is closed and discarded, otherwise, the timestamp for the
 ip address is updated in the global cache.
 """
 function check_rate_limit(tcp, rate_limit::Rational{Int})
-    ip = Sockets.getsockname(tcp)[1]
-    rate = Float64(rate_limit.num)
-    rl = get!(RATE_LIMITS[Threads.threadid()], ip, RateLimit(rate, Dates.now()))
+    ip = Sockets.getpeername(tcp)[1]
```
fredrikekre, line 57: How did this ever work before? It essentially acted like a global `rate_limiter` (same as `max_connections`) IIUC.

src/Servers.jl
```diff
@@ -51,14 +54,9 @@ soon, it is closed and discarded, otherwise, the timestamp for the
...
+    rl = get!(RATE_LIMITS[Threads.threadid()], ip, RateLimit(rate_limit, Dates.DateTime(0)))
```
fredrikekre, line 58: Use `DateTime(0)` here to not "punish" first time connections.

src/Servers.jl
```diff
@@ -51,14 +54,9 @@ soon, it is closed and discarded, otherwise, the timestamp for the
...
     update!(rl, rate_limit)
-    if rl.allowance > rate
-        @warn "throttling $ip"
-        rl.allowance = rate
-    end
```
fredrikekre, line 61: Moved into the `update!` function with the warning removed. Why should HTTP warn here at all?

src/Servers.jl
```diff
@@ -51,14 +54,9 @@ soon, it is closed and discarded, otherwise, the timestamp for the
 ip address is updated in the global cache.
 """
 function check_rate_limit(tcp, rate_limit::Rational{Int})
-    ip = Sockets.getsockname(tcp)[1]
-    rate = Float64(rate_limit.num)
```
fredrikekre, line 55: This also fixes the a bug if you specify `rate_limit = 3//2` or something (e.g. non-1 denominator).

src/Servers.jl
quinnj: Yeah, this was never very well-tested

src/Servers.jl
quinnj: I just thought it might be useful to know, server-side, if there are potentially malicious calls going on.

quinnj (approved):
Thanks!

src/Servers.jl
fredrikekre: But this is not the case here. For example if rate limiting allowed 2 connections per second, and it was 10 seconds since my last connection this would warn?

PR #703, opened 2021-04-28
Verify URL before making the request, fixes #527, fixes #545.

### LINKED ISSUE #527: URL scheme doesn't matter?
This does not seem like it should work:
```jl
julia> HTTP.request("HEAD", "floofy://www.google.com")
HTTP.Messages.Response:
"""
HTTP/1.1 200 OK
Date: Wed, 01 Apr 2020 15:58:39 GMT
Expires: -1
Cache-Control: private, max-age=0
Content-Type: text/html; charset=ISO-8859-1
P3P: CP="This is not a P3P policy! See g.co/p3phelp for more info."
Server: gws
X-XSS-Protection: 0
X-Frame-Options: SAMEORIGIN
Set-Cookie: 1P_JAR=2020-04-01-15; expires=Fri, 01-May-2020 15:58:39 GMT; path=/; domain=.google.com; Secure
Set-Cookie: NID=201=23KuvAtVAiJou_Mdnon-_d8_P32i-z72OaXTsyzbxPo0Uk7mYUfqdzFocUIU_38xGm7gF7hXPgZzLGVG8kNKr9QMlBWKAWojbZ0E5tcHr64qsYigx0xFBM5KJQksotDYnP7JbCvu1BmUncNtbj0DQTtDaSRfRJ5RTvgjAC4X7dc; expires=Thu, 01-Oct-2020 15:58:39 GMT; path=/; domain=.google.com; HttpOnly
Transfer-Encoding: chunked
Accept-Ranges: none
Vary: Accept-Encoding

"""
```
HTTP v0.8.13

### LINKED ISSUE #545: HTTP.request ignores /etc/hosts file
Julia  1.5.0-DEV.876
HTTP.jl 0.8.15

When requesting a URL that's only set via /etc/hosts and doesn't resolve via the default DNS server, I get a DNSError. The exact same URL works fine with `Base.download` and standard `curl` from the shell.

This is reproducible by setting an entry for a non-existing domain pointing to e.g. one of the servers of `dig google.com` in /etc/hosts, checking that something gets returned with `curl` (probably 404s), checking with `Base.download` that it also 404s and then trying the same with `HTTP.get(...)`, resulting in a DNSError.

fredrikekre:


src/RedirectRequest.jl
```diff
@@ -21,6 +21,11 @@ function request(::Type{RedirectLayer{Next}},
     count = 0
     while true
 
+        # Verify the url before making the request. Verification is done in the redirect
+        # loop to also catch bad redirect URLs.
+        # TODO: At first thought that seemed like a good idea, but perhaps it is not?
+        verify_url(url)
```
fredrikekre, line 27: Thoughts on this? Missing host will break DNS lookup so that should be safe, and I suppose it does not hurt to verify the scheme either.

src/RedirectRequest.jl
```diff
@@ -71,4 +76,13 @@ function isdomainorsubdomain(sub, parent)
     return sub[length(sub)-length(parent)] == '.'
 end
 
+function verify_url(url::URI)
+    if isempty(url.scheme) || !(url.scheme in ("http", "https", "ws", "wss"))
```
fredrikekre, line 80: Any more schemes that can be valid here?

quinnj: ftp maybe?

quinnj (approved):

fredrikekre:
I just noticed https://github.com/JuliaWeb/URIs.jl/blob/62a77a5b97ac1a5709c17baa1928b77d792d93b1/src/URIs.jl#L288 so perhaps that should be used? It checks a lot more than the bare minimum as this PR does, but perhaps that is fine.

src/RedirectRequest.jl
fredrikekre: Doesn't look like it, unless I am missing something. I suppose this can be relaxed later if it turns out it breaks some usecase.

PR #704, opened 2021-04-28
lazily allocate the threaded buffers and allocate them on the thread that will access it

KristofferC:
This package does quite a bit of work on load time since it allocates and compiles some regular expressions. That code needs to be inferred and run which causes some penalty to the load time of the package. In particular, the function `Threads.resize_nthreads!` seems to be slow to infer (https://github.com/JuliaLang/julia/issues/40630). In addition, it is generally better to allocate objects on the actual thread that they will be used. The pattern used in this PR is the same as the Random stdlib uses to allocate thread local RNGs: https://github.com/JuliaLang/julia/blob/e4fcdf5b04fd9751ce48b0afc700330475b42443/stdlib/Random/src/RNGs.jl#L369-L385.

Together with https://github.com/JuliaWeb/URIs.jl/pull/27 the timings for loading the package are:

Before:

```
julia> @time using HTTP
  0.428206 seconds (1.05 M allocations: 63.117 MiB, 2.00% gc time, 0.89% compilation time)
```

After:

```
julia> @time using HTTP
  0.113877 seconds (64.19 k allocations: 5.825 MiB, 7.65% compilation time)
```

### DIFF
--- a/src/CookieRequest.jl
+++ b/src/CookieRequest.jl
@@ -5,12 +5,12 @@ import ..Layer, ..request
 using URIs
 using ..Cookies
 using ..Pairs: getkv, setkv
-import ..@debug, ..DEBUG_LEVEL
+import ..@debug, ..DEBUG_LEVEL, ..access_threaded
 
-const default_cookiejar = [Dict{String, Set{Cookie}}()]
+const default_cookiejar = Dict{String, Set{Cookie}}[]
 
 function __init__()
-    Threads.resize_nthreads!(default_cookiejar)
+    resize!(empty!(default_cookiejar), Threads.nthreads())
     return
 end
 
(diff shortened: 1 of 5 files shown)


fredrikekre (approved):

src/CookieRequest.jl
```diff
@@ -5,12 +5,12 @@ import ..Layer, ..request
...
 using ..Pairs: getkv, setkv
-import ..@debug, ..DEBUG_LEVEL
+import ..@debug, ..DEBUG_LEVEL, ..access_threaded
 
-const default_cookiejar = [Dict{String, Set{Cookie}}()]
+const default_cookiejar = Dict{String, Set{Cookie}}[]
 
 function __init__()
-    Threads.resize_nthreads!(default_cookiejar)
+    resize!(empty!(default_cookiejar), Threads.nthreads())
```
quinnj, line 13: curious why the `empty!` call here?

quinnj (approved):

src/CookieRequest.jl
KristofferC: Just in case some garbage is left in it from the precompilation process that ends up getting serialized.

Issue #705, opened 2021-04-30
Bug for Proxy in requests?


liupgd:
Julia 1.6.0
HTTP.jl 0.9.7
WebSockets.jl 1.5.9

Hi, I'm using the separate package '[WebSockets.jl](https://github.com/JuliaWeb/WebSockets.jl)' which is based on HTTP.jl package. When I setted proxy for the websocket client, the client always returned 'Bad Request'. After my debug into HTTP.jl, I found maybe some code is wrong in *line 95~120 of ConnectionRequest.jl*.
* Ther original code:
```julia
    try
        if proxy !== nothing && target_url.scheme == "https"
            # tunnel request
            target_url = merge(target_url, port=443)
            r = connect_tunnel(io, target_url, req)
            if r.status != 200
                close(io)
                return r
            end
            io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
            req.headers = filter(x->x.first != "Proxy-Authorization", req.headers)
        end

        r =  request(Next, io, req, body; kw...)

        if (io.sequence >= reuse_limit
            || (proxy !== nothing && target_url.scheme == "https"))
            close(io)
        end

        return r
    catch e
        @debug 1 "❗️  ConnectionLayer $e. Closing: $io"
        try; close(io); catch; end
        rethrow(isioerror(e) ? IOError(e, "during request($url)") : e)
    end
```
The problem is that, if proxy is setted, the code only handle for https not for ws or wss.
* New code I modified:
```julia
try
      if proxy !== nothing# && target_url.scheme == "https"
      # tunnel request
          if target_url.scheme == "https" || target_url.scheme == "wss"  # !!! wss also go through ssl, right? !!!
              target_url = merge(target_url, port=443)
          end
            r = connect_tunnel(io, target_url, req)# !!! proxy for websockets also connect with tunnel.
          if r.status != 200
              close(io)
              return r
          end
          if target_url.scheme == "https" || target_url.scheme == "wss"   # !!! So the same case here?
              io = ConnectionPool.sslupgrade(io, target_url.host; kw...)
          end
          req.headers = filter(x -> x.first != "Proxy-Authorization", req.headers)
      end

      r =  request(Next, io, req, body; kw...)

      if (io.sequence >= reuse_limit
      || (proxy !== nothing && target_url.scheme == "https"))
          close(io)
      end

      return r
  catch e
      @debug 1 "❗️  ConnectionLayer $e. Closing: $io"
[...]

PR #708, opened 2021-05-02
fix bug in websocket with proxy

liupgd:
issue:https://github.com/JuliaWeb/HTTP.jl/issues/705
Now, both separate WebSockets.jl package and HTTP.jl package can use websocket with proxy now. 
The main bug is:
* for wss with proxy: it's not enabled with ssl
* for ws, maybe it is because the default 80 port not concated to the url end, so the request returned `Bad Request`


quinnj:
Sorry for the slow response/review here! I think this code looks good. It looks like this was originally a fork#master made as a PR to HTTP#master, which means I can't resolve the conflicts here, so I'm going to incorporate the code into an upcoming PR.

fonsp:
Implemented in https://github.com/JuliaWeb/HTTP.jl/pull/833, thanks @liupgd !

PR #713, opened 2021-05-17
Add access logging functionality

fredrikekre:
This adds access logging functionality to `HTTP.listen`/`HTTP.serve`. This includes a string macro that parses NGINX style format strings (chose that since it is pretty readable, self-explanatory, and looks more like Julia code compared to e.g. Apache format strings).

Example usage:
```julia
HTTP.listen(handler, host, port; access_log = logfmt"$remote_addr $request")
```
which would then log
```
[ Info: ::1 GET / HTTP/1.1
[ Info: ::1 POST /hello/post HTTP/1.1
[ Info: ::ffff:7f00:1 POST /hello/post HTTP/1.1
[ Info: ::ffff:7f00:1 HEAD /hello/post HTTP/1.1
```
etc.

Originally I thought this had to be implemented "on the inside" in order to get access to e.g. number of written bytes etc, but (as done in this PR) those things can technically be stored in the `http::Stream` object and you could do:
```julia
HTTP.listen(host, port) do http
    # handle stuff
    do_log(http) # do logging like this instead
end
```
but it is pretty convenient to just  have to pass the `access_log` keyword and it does not really cost anything. In addition, if you want manual control you can still do it like above.

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -74,7 +74,10 @@ struct Server{S <: Union{SSLConfig, Nothing}, I <: Base.IOServer}
     hostname::String
     hostport::String
     on_shutdown::Any
+    access_log::Union{Function,Nothing}
 end
+Server(ssl, server, hostname, hostport, on_shutdown) =
+    Server(ssl, server, hostname, hostport, on_shutdown, nothing)
 
 Base.isopen(s::Server) = isopen(s.server)
 Base.close(s::Server) = (shutdown(s.on_shutdown); close(s.server))
@@ -143,6 +146,10 @@ Optional keyword arguments:
     allowed per client IP address; excess connections are immediately closed.
     e.g. 5//1.
  - `verbose::Bool=false`, log connection information to `stdout`.
+ - `access_log::Function`, function for formatting access log messages. The
+    function should accept two arguments, `io::IO` to which the messages should
+    be written, and `http::HTTP.Stream` which can be used to query information
+    from. See also [`@logfmt_str`](@ref).
  - `on_shutdown::Union{Function, Vector{<:Function}, Nothing}=nothing`, one or
     more functions to be run if the server is closed (for example by an
     `InterruptException`). Note, shutdown function(s) will not run if a
@@ -234,6 +241,7 @@ function listen(f,
                 reuse_limit::Int=nolimit,
                 readtimeout::Int=0,
                 verbose::Bool=false,
+                access_log::Union{Function,Nothing}=nothing,
                 on_shutdown::Union{Function, Vector{<:Function}, Nothing}=nothing)
 
     inet = getinet(host, port)
@@ -262,7 +270,7 @@ function listen(f,
         x -> f(x) && check_rate_limit(x, rate_limit)
     end
 
-    s = Server(sslconfig, tcpserver, string(host), string(port), on_shutdown)
+    s = Server(sslconfig, tcpserver, string(host), string(port), on_shutdown, access_log)
     return listenloop(f, s, tcpisvalid, connection_count, max_connections,
                          reuse_limit, readtimeout, verbose)
 end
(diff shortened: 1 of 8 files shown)


src/Servers.jl
```diff
@@ -412,6 +418,9 @@ function handle_transaction(f, t::Transaction, server; final_transaction::Bool=f
         closeread(http)
         @debug 2 "server closewrite"
         closewrite(http)
+        if server.access_log !== nothing
+            @info sprint(server.access_log, http) _group=:access_log
```
oxinabox, line 422: might want to call the group just `:access` ?
Saying `:access_log` seems redudant.

src/Streams.jl
```diff
@@ -102,11 +105,15 @@ function Base.unsafe_write(http::Stream, p::Ptr{UInt8}, n::UInt)
...
+        return nw
     end
-    return write(http.stream, string(n, base=16), "\r\n") +
-           unsafe_write(http.stream, p, n) +
-           write(http.stream, "\r\n")
+    nw = write(http.stream, string(n, base=16), "\r\n") +
+         unsafe_write(http.stream, p, n) +
+         write(http.stream, "\r\n")
+    http.nwritten += nw
+    return nw
```
oxinabox, line 116: Would it be cleaner to restructure this to be:
```
    nw = if http.writechunked
        write(http.stream, string(n, base=16), "\r\n") +
        unsafe_write(http.stream, p, n) +
        write(http.stream, "\r\n")
    else
        unsafe_write(http.stream, p, n)        
    end
    http.nwritten += nw
    return nw
```

src/access_log.jl
```diff
@@ -0,0 +1,110 @@
...
+ - `\$status`: response status code
+ - `\$body_bytes_sent`: number of bytes in response body
+
+## Examples
+```julia
+logfmt"[\$time_iso8601] \\"\$request\\" \$status" # [2021-05-01T12:34:40+0100] "GET /index.html HTTP/1.1" 200
+
+logfmt"\$remote_addr \\"\$http_user_agent\\"" # 127.0.0.1 "curl/7.47.0"
+```
+"""
```
oxinabox, line 30: Is there a way wr can write `raw"""` for docstrings?
The multiple layers of escaping does my head-in.
and I know a surprising number of people read docstrings not through REPL or web, but by looking at the source

test/server.jl
```diff
@@ -251,24 +251,110 @@ end
     @test HTTP.Servers.shutdown(nothing) === nothing
 
     IOserver = Sockets.listen(Sockets.InetAddr(parse(IPAddr, "127.0.0.1"), 8052))
-    
+
     # Shutdown adds 1
     TEST_COUNT = Ref(0)
     shutdown_add() = TEST_COUNT[] += 1
-    server = HTTP.Servers.Server(nothing, IOserver, "host", "port", shutdown_add)
+    server = HTTP.Servers.Server(nothing, IOserver, "host", "port", shutdown_add, nothing)
```
oxinabox, line 258: Why does this not default to `nothing`?
Isn't making this argument required going to be a breaking change?

oxinabox (approved):
Basically LGTM.
Other than a few things, most importantly making sure it doesn't break existing code if it doesn't have to.

src/Servers.jl
fredrikekre: Sure, updated.

src/access_log.jl
fredrikekre: You still need `\\` for `\` etc, but not for `$` so updated.

test/server.jl
fredrikekre: I mean, this is pretty internal I would think? The next to last argument was added pretty recently too. Doesn't hurt to be conservative though

fredrikekre:
Thanks for the review, updated all of it I think (and fixed tests on Windows).

oxinabox:
You have merge rights here, yes?

quinnj:
Yeah, sorry for the slow review; this looks great!

PR #720, opened 2021-06-06
Add example on reducing time to first response

### LINKED ISSUE #579: The first request(as client) or response(as server) of Http.jl is very slow?  Initial connection should not be so slow
Julia 1.5.0
HTTP.jl v0.8.17
MbedTLS.jl v1.0.2

```sh
julia> using HTTP

julia> @time r = HTTP.request("GET", "http://0.0.0.0:9000/"; verbose=0)
  3.086142 seconds (8.95 M allocations: 450.041 MiB, 7.77% gc time)

julia> @time r = HTTP.request("GET", "http://0.0.0.0:9000/"; verbose=0)
  0.030034 seconds (25.44 k allocations: 1.303 MiB)

julia> @time r = HTTP.request("GET", "http://baidu.com"; verbose=0)
  0.118624 seconds (362 allocations: 85.906 KiB)
```
```sh
julia                                                                                                                                                                             (20m 5s 996ms)

               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0 (2020-08-01)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |


julia> using HTTP

julia> @time r = HTTP.request("GET", "http://baidu.com"; verbose=0)
  2.939621 seconds (8.95 M allocations: 450.028 MiB, 6.21% gc time)
HTTP.Messages.Response:
```

rikhuijzer:
In many situations, the HTTP server responds quite slow to the first request. For example, I've described [here](https://discourse.julialang.org/t/precompilation-techniques-for-speeding-up-the-first-run-heroku-deployments-and-packagecompiler/29327/18) that the first request took 3 seconds for a simple Julia server running on Heroku. Also, I notice this when using [LiveServer.jl](https://github.com/tlienart/LiveServer.jl).

To solve this, we could use [PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl). However, although this is the most reliable solution, this adds a lot of overhead to the end user. In this PR, a `trigger_compilation` optional keyword is added to the server, which has as benefit that it is easier to use than PackageCompiler. 

Basically, the trigger_compilation will spawn an async job which hits one of the server endpoints. That way, Julia is forced to compile a large part of HTTP. This reduces the time to first response, which is useful in situations where users interact with the server. For example, when locally starting a server, then getting a cup of coffee and opening a webpage on the server or when starting a server online.

I've disabled the keyword by default because enabling it could be breaking in some cases.

Also fixes #579.
 
### `trigger_compilation=false`
```
using HTTP
using Sockets

router = HTTP.Router()
respond(req) = HTTP.Response(200)
HTTP.@register(router, "GET", "/", respond)

host = "127.0.0.1"
port = 8080

@async HTTP.serve(router, host, port; trigger_compilation=false)
sleep(5)
@time HTTP.get("http://$host:$port")
```
```
3.987207 seconds (9.51 M allocations: 543.961 MiB, 5.56% gc time, 6.40% compilation time)
```

### `trigger_compilation=true`
```
using HTTP
using Sockets

router = HTTP.Router()
respond(req) = HTTP.Response(200)
HTTP.@register(router, "GET", "/", respond)

host = "127.0.0.1"
port = 8080

@async HTTP.serve(router, host, port; trigger_compilation=true)
sleep(5)
@time HTTP.get("http://$host:$port")
```
```
0.415061 seconds (591.34 k allocations: 33.470 MiB, 99.67% compilation time)
```

### `trigger_compilation=true` and HTTP.post
```
using HTTP
using Sockets

router = HTTP.Router()
respond(req) = HTTP.Response(200)
HTTP.@register(router, "POST", "/", respond)

host = "127.0.0.1"
port = 8080

@async HTTP.serve(router, host, port; trigger_compilation=true)
sleep(5)
@time HTTP.post("http://$host:$port")
```
```
0.483329 seconds (575.69 k allocations: 33.907 MiB, 1.37% gc time, 89.22% compilation time)
```

### DIFF
--- a/docs/examples/reduce_first_response_time.jl
+++ b/docs/examples/reduce_first_response_time.jl
@@ -18,12 +18,12 @@ const trigger_compilation = retry(; delays=ExponentialBackOff(n=5, first_delay=0
     HTTP.get(url)
 end
 
-function status(req::HTTP.Request)
-    HTTP.Response(200, "Ok")
+function hello(req::HTTP.Request)
+    HTTP.Response(200, "Hello")
 end
 
 const ROUTER = HTTP.Router()
-HTTP.@register(ROUTER, "GET", "/status", status)
+HTTP.@register(ROUTER, "GET", "/", hello)
 
 @async trigger_compilation()
 println("Starting server at http://$host:$port")
--- a/docs/examples/reduce_first_response_time.jl
+++ b/docs/examples/reduce_first_response_time.jl
@@ -1,0 +1,28 @@
+"""
+Due to compilation, it can take more than a second for the server to respond on the first request.
+There are two ways to reduce this time to a few milliseconds.
+One way is to use [PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl) and another is to trigger compilation, which is shown below.
+Note that this trigger is only useful in situations where the server is not connected to immediately after starting it.
+For example, this happens when running a server on Heroku.
+"""
+using HTTP
+using Sockets
+
+host = Sockets.localhost
+
+function trigger_compilation(port)
+    sleep(0.5)
+    url = "http://$host:$port/"
+    HTTP.get(url)
+end
+
+function status(req::HTTP.Request)
+    HTTP.Response(200, "Ok")
+end
+
+const ROUTER = HTTP.Router()
+HTTP.@register(ROUTER, "GET", "/status", status)
+
+port = 8080
+@async trigger_compilation(port)
+HTTP.serve(ROUTER, host, port)
--- a/docs/examples/reduce_first_response_time.jl
+++ b/docs/examples/reduce_first_response_time.jl
@@ -2,8 +2,8 @@
 Due to compilation, it can take more than a second for the server to respond on the first request.
 There are two ways to reduce this time to a few milliseconds.
 One way is to use [PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl) and another is to trigger compilation, which is shown below.
-Note that this trigger is only useful in situations where the server is not connected to immediately after starting it.
-For example, this happens when running a server on Heroku.
+Note that this trigger is only useful if there is time in between startup and the first request.
+For example, this happens when running a user facing server on Heroku.
 """
 using HTTP
 using Sockets
--- a/docs/examples/reduce_first_response_time.jl
+++ b/docs/examples/reduce_first_response_time.jl
@@ -9,10 +9,12 @@ using HTTP
 using Sockets
 
 host = Sockets.localhost
+port = 8080
 
-function trigger_compilation(port)
-    sleep(0.5)
-    url = "http://$host:$port/"
+# Retry returns an anonymous function.
+const trigger_compilation = retry(; delays=ExponentialBackOff(n=5, first_delay=0.5)) do
+    url = "http://$host:$port"
+    # Hit "GET /" (and possibly more routes if needed)
     HTTP.get(url)
 end
 
@@ -23,6 +25,6 @@ end
 const ROUTER = HTTP.Router()
 HTTP.@register(ROUTER, "GET", "/status", status)
 
-port = 8080
-@async trigger_compilation(port)
+@async trigger_compilation()
+println("Starting server at http://$host:$port")
 HTTP.serve(ROUTER, host, port)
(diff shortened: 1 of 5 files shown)


rikhuijzer:
Possibly, it would even be best to set `trigger_compilation=true` by default and make the next HTTP.jl release a major one. That way, new(ish) Julia users would benefit from having a more responsive server by default. In the worst case, `@async trigger` would fail silently.

rikhuijzer:
@quinnj Could you take a look at this PR or ping the appropriate person? :) I still think this PR is a great idea, but am open for suggestions.

fredrikekre:
I think this is little strange. It seems trivial to just do this yourself? I don't see the reason this should be done from within HTTP.jl

rikhuijzer:
> It seems trivial to just do this yourself? 

Fair point. At the same time, it isn't completely trivial to come up with it, I think. I concluded that I'm not the only one who likes to use this small trick based on the Discourse post that I've linked in the first post.

Therefore, I changed it now to a mention in the docs. @fredrikekre, what do you think? If you're still not convinced feel free to close the issue.

rikhuijzer:
I've just implemented this on another project of mine and the difference on Heroku is **massive**. It goes from 1.2 seconds to 0.025 milliseconds (almost 50 times faster) on the first hit on one of the endpoints and from 7 seconds to 2.6 seconds on another endpoint.

@fredrikekre, I really don't get why this isn't something which could at least end up in the docs? For anyone who isn't experienced with Julia, this is **not trivial** to come up with. Also, even with PackageCompiler, you would have to specify which parts of the codebase need to be compiled, so the small trick included in this PR is super easy and works super well. 

Or, what am I missing? Is there another way to ensure that the first response of a (Heroku) server remains below 5 seconds?

fredrikekre:
> @fredrikekre, I really don't get why this isn't something which could at least end up in the docs?

I don't think I said this? I think it is a good example to include.

The reason why *I* prefer to not build it into the `listen` function is that perhaps `GET /` is not the best way entrypoint for example, and perhaps you want to precompile multiple endpoints. With examples like the PR currently does it is easy to extend to your own usecase.

rikhuijzer:
> I don't think I said this? I think it is a good example to include.

You're right. I sort of inferred it from the lack of reply.

> I think it is a good example to include.

Great. Thanks. Feel free to update the text if you want

fredrikekre:
Perhaps it could be nice to retry, if the server has some startup time, for example:
```julia
using HTTP

const host = "0.0.0.0"
const port = 8080

const trigger_compilation = retry(; delays=ExponentialBackOff(n=5, first_delay=0.5)) do
    url = "http://$host:$port"
    # Hit "GET /" (and possibly more ones if needed)
    HTTP.get(url)
end

@async trigger_compilation()

HTTP.listen(host, port; access_log=combined_logfmt) do http
    HTTP.setstatus(http, 200)
    msg = "hello, world\n"
    HTTP.setheader(http, "Content-Length" => string(sizeof(msg)))
    HTTP.startwrite(http)
    HTTP.write(http, msg)
    return
end
```

rikhuijzer:
Good idea. I've added it.

rikhuijzer:
I'm closing this because I've walked into some practical problems with it:

1. The `trigger_compilation` method needs to be tested outside of an async. Otherwise, it can fail without showing an error.
1. It can be difficult to trigger all the routes in a production server, because it would alter the production database and logs.

I'm planning to make something like https://github.com/Optomatica/heroku-buildpack-julia/ with PackageCompiler. That should be the best solution, because it allows triggering all the routes with a development server instead of a production server.

fredrikekre:
> The trigger_compilation method needs to be tested outside of an async. Otherwise, it can fail without showing an error.

One can reverse the order I guess, e.g.
```
server = @async listen(...)

trigger_compilation()

wait(server)
```

> It can be difficult to trigger all the routes in a production server

But even if you hit some route all of the HTTP.jl machinery should be compiled.

rikhuijzer:
> One can reverse the order I guess, e.g.

Yes, exactly. That's what I did in my tests. It's okay, but something that one has to take into account.

> But even if you hit some route all of the HTTP.jl machinery should be compiled.

In practice, I do see delays of about a second on the first response of all routes that I didn't hit with `trigger_compilation`. Specifically, I'm hitting 3 routes in the trigger. Some of these routes include some database work even. Still, when another route is hit by users, the service time is about one second. The next 10+ hits are all around 100 ms, so I'm quite sure that compilation is causing the delay.

PR #727, opened 2021-06-28
hide IOErrors when remote closes the connection

pfitzseb:
because these aren't actionable anyways.

Fixes https://github.com/JuliaWeb/HTTP.jl/issues/392.


tanmaykm:
Should it be logged at `@debug` or `@info` level instead of completely ignoring? I believe servers like nginx do log a "upstream prematurely closed connection" when this happens. May be there are situations when it would be useful to know that this happened at the server?

pfitzseb:
Makes sense.

src/Servers.jl
```diff
@@ -421,7 +421,11 @@ function handle_transaction(f, t::Transaction, server; final_transaction::Bool=f
         @debug 2 "server closewrite"
         closewrite(http)
     catch e
-        @error "error handling request" exception=(e, stacktrace(catch_backtrace()))
+        # The remote can close the stream whenever it wants to, but there's nothing
+        # anyone can do about it on this side. No reason to log an error in that case.
+        level = e isa Base.IOError && !isopen(http) ? Logging.Debug : Logging.Error
```
fredrikekre, line 426: Since only the loglevels are needed you can directly use (and remove the Logging dep):
```suggestion
        level = e isa Base.IOError && !isopen(http) ? Base.CoreLogging.Debug : Base.CoreLogging.Error
```

fredrikekre (commented):
LGTM, add a note in the changelog?

src/Servers.jl
pfitzseb: Yes, but isn't that technically internal? Unlikely to break, I suppose, but still...

CHANGELOG.md
```diff
@@ -5,19 +5,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- "Connection closed by peer" errors are now emitted as `Debug`-level messages ([#727]).
```
fredrikekre, line 9:
```suggestion
- "Connection closed by peer" errors are now emitted as `Debug`-level messages (instead of `Error`-level) ([#727]).
```

src/Servers.jl
fredrikekre: I guess, although `@logmsg` is from Base so :shrug: Anyway, a stdlib dependency costs nothing.

pfitzseb:
Wanna tag a new patch release,  @fredrikekre? :)

fredrikekre:
Was already working on it: https://github.com/JuliaWeb/HTTP.jl/pull/729

PR #739, opened 2021-08-02
Set timeout limit for tests

rikhuijzer:
In response to recent runs taking 6 hours on nightly.


DilumAluthge:
@quinnj @StefanKarpinski It would be good to get this merged.

The specific value of the timeout can always be increased later, but anything is better than GitHub's default of 6 hours.

DilumAluthge (approved):
60 minutes would also be a reasonable timeout, if you wanted to go a little bit higher.

quinnj:
The immediate concern has been fixed in #745, where I fixed the websocket tests that were just hanging forever.

PR #741, opened 2021-08-13
Nothing comparison improvement

AtsushiSakai:
I'm not sure this is intended or not, but some codes are using equality comparison (==) for nothing, not identical comparison (===).
If this is not intended, I think later one should be used for small performance improvement.


quinnj (approved):
Thanks! I don't think it matters as much in the tests, but in download.jl, yes.

PR #744, opened 2021-08-24
Dummy commit to trigger CI

DilumAluthge:


DilumAluthge:
Why are all of the jobs timing out? The nightly jobs I expected to time out, but the other jobs should be passing.

fredrikekre:
Looks like they aren't even starting? Well, some have come to the test stage, but that stage has not started.

DilumAluthge:
Looks like they all eventually made it to the test stage, and then timed out in the test stage.

PR #745, opened 2021-08-31
Improve memory use and performance of multipart parsing

quinnj:
Memory usage and performance noted in
https://discourse.julialang.org/t/http-parse-multipart-form-does-a-lot-of-memory-allocations/66661.
This uses some of the same regex tricks we use in the other http parsing
code that operates on SubStrings and avoids excessive allocations. We
also streamline things to avoid allocating when comparing byte buffers
and the unnecessary allocated arrays of the content disposition
processing.

On my machine for current master, I see `35.022 μs (866 allocations:
70.27 KiB)` for parsing a simple multipart request from the tests, and
with this PR, I see `3.603 μs (42 allocations: 2.55 KiB)`.

### DIFF
--- a/src/IOExtras.jl
+++ b/src/IOExtras.jl
@@ -66,14 +66,37 @@ _doc = """
 
 Signal start/end of write or read operations.
 """
-"$_doc"
-startwrite(io) = nothing
-"$_doc"
-closewrite(io) = nothing
-"$_doc"
-startread(io) = nothing
-"$_doc"
-closeread(io) = nothing
+if isdefined(Base, :startwrite)
+    "$_doc"
+    Base.startwrite(io) = nothing
+else
+    "$_doc"
+    startwrite(io) = nothing
+end
+
+if isdefined(Base, :closewrite)
+    "$_doc"
+    Base.closewrite(io) = nothing
+else
+    "$_doc"
+    closewrite(io) = nothing
+end
+
+if isdefined(Base, :startread)
+    "$_doc"
+    Base.startread(io) = nothing
+else
+    "$_doc"
+    startread(io) = nothing
+end
+
+if isdefined(Base, :closeread)
+    "$_doc"
+    Base.closeread(io) = nothing
+else
+    "$_doc"
+    closeread(io) = nothing
+end
 
 using MbedTLS: SSLContext
 tcpsocket(io::SSLContext)::TCPSocket = io.bio
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -11,7 +11,7 @@ const DASH_BYTE = 0x2d # -
 const HTAB_BYTE = 0x09 # \t
 const SPACE_BYTE = 0x20
 const SEMICOLON_BYTE = UInt8(';')
-const CRLFCRLF = [CR_BYTE, LF_BYTE, CR_BYTE, LF_BYTE]
+const CRLFCRLF = (CR_BYTE, LF_BYTE, CR_BYTE, LF_BYTE)
 
 "compare byte buffer `a` from index `i` to index `j` with `b` and check if they are byte-equal"
 function byte_buffers_eq(a, i, j, b)
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -1,9 +1,27 @@
+module MultiPartParsing
+
+import ..access_threaded, ..Request, ..Multipart, ..payload
+using ..Parsers
+
+export parse_multipart_form
+
 const CR_BYTE = 0x0d # \r
 const LF_BYTE = 0x0a # \n
 const DASH_BYTE = 0x2d # -
 const HTAB_BYTE = 0x09 # \t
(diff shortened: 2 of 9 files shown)


src/parsemultipart.jl
```diff
@@ -1,17 +1,35 @@
...
+end
 
 """
     find_multipart_boundary(bytes, boundaryDelimiter; start::Int=1)
 
 Find the first and last index of the next boundary delimiting a part, and if
 the discovered boundary is the terminating boundary.
 """
-function find_multipart_boundary(bytes::AbstractVector{UInt8}, boundaryDelimiter::AbstractVector{UInt8}; start::Int=1)
+@inline function find_multipart_boundary(bytes::AbstractVector{UInt8}, boundaryDelimiter::AbstractVector{UInt8}; start::Int=1)
```
KristofferC, line 32: I'm kind of amazed that inlining this function has any performance improvement.

quinnj: Oh, it doesn't. This was from an iteration where I was returning a SubString instead of indices and it could avoid the allocation if it was inlined.

mbauman:
I can confirm this **significantly** improves the original request that was being tested/reported on Discourse — it is much larger and has an even larger benefit.

Before:

```Julia
julia> @benchmark HTTP.parse_multipart_form($r)
BenchmarkTools.Trial: 26 samples with 1 evaluation.
 Range (min … max):  181.312 ms … 235.950 ms  ┊ GC (min … max): 3.11% … 3.33%
 Time  (median):     195.296 ms               ┊ GC (median):    3.16%
 Time  (mean ± σ):   197.687 ms ±  13.029 ms  ┊ GC (mean ± σ):  3.20% ± 0.16%

 Memory estimate: 584.19 MiB, allocs estimate: 4785830.
```

After:

```Julia
julia> @benchmark HTTP.parse_multipart_form($r)
BenchmarkTools.Trial: 917 samples with 1 evaluation.
 Range (min … max):  5.131 ms …   9.564 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     5.309 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.447 ms ± 448.252 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

 Memory estimate: 2.03 KiB, allocs estimate: 32.
```

src/parsemultipart.jl
```diff
@@ -1,17 +1,35 @@
...
+export parse_multipart_form
+
 const CR_BYTE = 0x0d # \r
 const LF_BYTE = 0x0a # \n
 const DASH_BYTE = 0x2d # -
 const HTAB_BYTE = 0x09 # \t
 const SPACE_BYTE = 0x20
-const RETURN_BYTES = [CR_BYTE, LF_BYTE]
+const SEMICOLON_BYTE = UInt8(';')
+const CRLFCRLF = [CR_BYTE, LF_BYTE, CR_BYTE, LF_BYTE]
```
KristofferC, line 14: Make into a tuple (allows the compiler to know it's length etc)? I don't think it is mutated anywhere?

quinnj:
Ok, I unexported `closewrite` (and friends) so nightly isn't broken due to Julia Base now exporting `closewrite`. I also fixed the websocket tests which were failing because echo.websocket.org doesnt' exist anymore. Let's see how CI fares now.

src/IOExtras.jl
```diff
@@ -66,14 +66,37 @@ _doc = """
...
-startwrite(io) = nothing
-"$_doc"
-closewrite(io) = nothing
-"$_doc"
-startread(io) = nothing
-"$_doc"
-closeread(io) = nothing
+if isdefined(Base, :startwrite)
+    "$_doc"
+    Base.startwrite(io) = nothing
```
fredrikekre, line 71: This is type piracy now

quinnj: Yeah, we'll need to make a plan around this since Base doesn't define a fallback for `closewrite`. Either we add `closewrite(io) = nothing` to Base and we can check if it's defined here, or it's also possible we don't need the fallback anymore. I left this as-is for now since it ensures non-breaking.

src/parsemultipart.jl
KristofferC: So should it have been removed?

src/IOExtras.jl
quinnj: Because then that's breaking; we have all these code examples all over the package where we tell people to write streaming servers doing `closewrite(stream)`.

KristofferC: You can still export it. Downstreams users just must then decide if they want to use `Base.closewrite` or `HTTP.closewrite`.

DilumAluthge: Personally, I would strongly prefer that we remove the type piracy and just make a breaking release of HTTP.jl.

DilumAluthge: https://github.com/JuliaWeb/HTTP.jl/pull/749

PR #749, opened 2021-09-02
[Breaking] Remove the type piracy of several `Base` functions

DilumAluthge:
This pull request removes the type piracy of several `Base` functions. Specifically, it removes the following pirated methods:
1. `Base.startwrite(io::Any)`
2. `Base.closewrite(io::Any)`
3. `Base.startread(io::Any)`
4. `Base.closeread(io::Any)`

According to https://github.com/JuliaWeb/HTTP.jl/pull/745#discussion_r699633657, this is a breaking change. Therefore, I took advantage of the opportunity to bump the version number to `1.0.0`.

For context, see: https://github.com/JuliaWeb/HTTP.jl/pull/745#pullrequestreview-743118202 and https://github.com/JuliaLang/julia/pull/42080.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "HTTP"
 uuid = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 authors = ["Jacob Quinn", "Sam O'Connor", "contributors: https://github.com/JuliaWeb/HTTP.jl/graphs/contributors"]
-version = "0.9.14"
+version = "1.0.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/src/IOExtras.jl
+++ b/src/IOExtras.jl
@@ -66,37 +66,18 @@ _doc = """
 
 Signal start/end of write or read operations.
 """
-if isdefined(Base, :startwrite)
-    "$_doc"
-    Base.startwrite(io) = nothing
-else
-    "$_doc"
-    startwrite(io) = nothing
-end
 
-if isdefined(Base, :closewrite)
-    "$_doc"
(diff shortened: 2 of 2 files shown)


DilumAluthge:
cc: @fredrikekre @KristofferC @quinnj

DilumAluthge:
@quinnj I'll probably need your help to fix the tests.

c42f:
None of `startwrite`, `startread`, `closeread` are defined in Base yet and IIUC there's no particular plan to add them. (I could potentially see `closeread` being added for symmetry with `closewrite` in the future. Not so sure about startread/startwrite.)

From my point of view I like @quinnj's current solution better for `closewrite`, somewhat hacky as it may be. The point is that `closewrite` and the new `Base.closewrite` have *exactly* the same semantics. Thus, they deserve to share method tables.

SemVer is a blunt tool as usual... but if we need to release a breaking change anyway, I'd suggest we use it to figure out how to make the semantics of `HTTP.closewrite` compatible with `Base`.

KristofferC:
> The point is that closewrite and the new Base.closewrite should have exactly the same semantics because they solve the same problem. Thus, they deserve to share method tables.

Why should `closewrite(io) = nothing` exist then? And in that case, it should be moved to Base.

c42f:
> Why should closewrite(io) = nothing exist? And in that case, it should be moved to Base.

Well I agree you're right about that! If this is a nonsensical method, we should just remove this fallback method from HTTP.jl and declare a breaking change.

What I disagree about is something subtler: this PR makes `HTTP.closewrite` and `Base.closewrite` into separate generic functions with disjoint method tables and clashing exported names. I don't think we should do this because they have the same semantics.

PR #751, opened 2021-09-07
remove an inline macro from a big function

KristofferC:


quinnj:
Anybody happen to know why my `PIE_SOCKET_API_KEY` environment variable doesnt' seem to be getting set correctly? It worked in the PR I did recently and all tests passed, but not it seems like there _is_ an environment variable (if not, it will skip those tests), but now we're getting tests failed because the api_key isnt' correct. @DilumAluthge ?

DilumAluthge:
Is the `PIE_SOCKET_API_KEY` environment variable a secret?

For security reasons, GitHub does not make secrets available on PRs from forks. (This PR was made from a fork, presumably because Kristoffer does not have write access on this repo.)

In some of my repos (RegistryCI.jl and CompatHelper.jl), I use Bors to solve this limitation.

quinnj:
Ah, that must be it. How does Bors solve the issue here?

DilumAluthge:
So, roughly speaking:
- If the branch is in the main `JuliaWeb/HTTP.jl` repo, then GitHub secrets are available.
- If the branch is from a fork (like in this case), then GitHub secrets are not available.

With Bors, the idea is this:
1. First, you review the PR, and make sure that the PR author is not doing anything shady, like trying to steal your secrets.
2. Once you are confident that the PR is safe, you do either `bors try` (which runs CI, but does not merge the PR) or `bors merge` (which runs CI, and merges the PR if CI passes). The key here is that `bors try` pushes to a branch named `trying` within the main `JuliaWeb/HTTP.jl` repo, and `bors merge` pushes to a branch named `staging` within the main `JuliaWeb/HTTP.jl` repo. Because the branch is inside the main repo, secrets are available to the builds that Bors triggers.

Take a look at the CI configuration in RegistryCI.jl or CompatHelper.jl.

quinnj:
Ok, that sounds useful, though it does mean one extra step. I wonder if we should just make the tests more robust and check that not only is the environment variable defined, but it has a non-empty value before attempting those tests. That would mean simple PRs like this would "just work", but there would be potentially more overhead to consider whether a PR should have a `bors try` or not depending whether they're from a fork and do something with websockets (the only tests that currently rely on pie socket).

quinnj:
Merging this for now, since it's trivial

PR #753, opened 2021-09-14
Fix error messages in @ require precondition testing macro

c42f:
With a little hackery we can extract the function name from the internal
`#self#` without the need to pattern match backtraces.

Copies fixes from https://github.com/JuliaWeb/URIs.jl/pull/33


c42f:
The test failures here seem due to the ongoing flakiness of the external websockets test host.

fredrikekre (approved):
These macros are pretty strange IMO, but easier to fix than to rewrite the code that uses them I guess.

c42f:
Yeah fair comment :-)

PR #754, opened 2021-09-15
Make sure that the API secret is not empty.

fredrikekre:
Making this PR from my fork to test this out.


DilumAluthge:
Maybe your fork is still getting the secrets because you are a committer?

fredrikekre:
But then tests should pass :)

PR #756, opened 2021-09-15
Skip the authenticated tests (tests that require an API secret) on PRs from forks

DilumAluthge:
Closes #754

### DIFF
--- a/test/client.jl
+++ b/test/client.jl
@@ -178,7 +178,7 @@ end
     end
 end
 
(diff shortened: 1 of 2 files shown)


DilumAluthge:
cc: @fredrikekre

fredrikekre:
Oh wow, two identical places with the same `println` output...

quinnj (commented):
Thanks for doing this @DilumAluthge!

Issue #760, opened 2021-09-18
Wrong file name for download from a FigShare URL


briochemc:
```
Julia 1.6.2
HTTP v0.9.14
MbedTLS v1.0.3
```

This minimal example,
```julia
using DataDeps, XLSX

register(DataDep(
            "FigShare_dataset",
            "No description needed",
            "https://ndownloader.figshare.com/files/6294558",
            "ba250b1b64b8c43d1130a7aee39674e9ee5936ebb49050c4c224c740bff588b0"
))

XLSX.read(joinpath(datadep"FigShare_dataset", "rsta20150293_si_001.xlsx"))
```

downloads a table from FigShare (doi: https://doi.org/10.6084/m9.figshare.3980064.v1) and tries to read it (using DataDeps.jl) but fails with `No such file or directory`. It fails because instead of the correct name (`"rsta20150293_si_001.xlsx"`), the file is saved as `6294558` (which is the last "word" of the URL). However, this exact same snippet used to work sometime in the past year, so something changed since then. Talking briefly with @oxinabox, he suggested that this could be an issue for HTTP.jl or an issue from FigShare itself. I tried to dig when/where something changed using blame here, but I failed to figure it out, and pinning earlier package versions did not work either. 

I don't come with just a problem, FWIW, a solution/workaround (thanks to Lyndon as well) is to rename the file in post-processing after download. Thus at this stage this is not an issue for me anymore, but hopefully posting all these details will help someone here find a fix! 😃

fredrikekre:
What filename does curl give?

briochemc:
Apologies if this is dumb, but I'm not sure how to answer that! 😅

oxinabox:
Don't worry, I got this.

```sh
❱ curl -Li "https://ndownloader.figshare.com/files/6294558"                    
HTTP/1.1 302 Found
Server: nginx
Date: Sat, 18 Sep 2021 10:28:44 GMT
Content-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
Content-Length: 0
Connection: keep-alive
...
Location: https://s3-eu-west-1.amazonaws.com/pstorage-rs-4828782598/6294558/rsta20150293_si_001.xlsx?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-
...
Content-Disposition: attachment;filename=rsta20150293_si_001.xlsx
...
```

That content-disposition is what we should (and normally do) use to determine file name.

[...]

PR #761, opened 2021-09-18
Make HTTP.download support redirects when the redirected URL does not set Content-Disposition

oxinabox:
Closes https://github.com/JuliaWeb/HTTP.jl/issues/760

I do not love this solution as it used HEAD requrests.
And I am pretty sure those are not supported everywhere.
Even though the standard requires that they are supported everywhere `GET` is.
But I don't have a counter-example handy.

Possibly the better way to do this is to first try and fully do the download with redirect off.
If that fails, remember the filename,
and then retry with redirect on.

Does anyone have thoughts (or a URL I can use to break this?)

This uses the URL from #760 as go-httpbin doesn't have a way to create a redirect chain that has the first with the content-disposition, and the last without.
https://github.com/postmanlabs/httpbin/issues/652

### DIFF
--- a/src/download.jl
+++ b/src/download.jl
@@ -16,9 +16,8 @@ function safer_joinpath(basepart, parts...)
     joinpath(basepart, parts...)
 end
 
-function try_get_filename_from_headers(resp)
-    content_disp = header(resp, "Content-Disposition")
-    if content_disp != ""
+function try_get_filename_from_headers(hdrs)
+    for content_disp in hdrs
         # extract out of Content-Disposition line
         # rough version of what is needed in https://github.com/JuliaWeb/HTTP.jl/issues/179
         filename_part = match(r"filename\s*=\s*(.*)", content_disp)
@@ -55,16 +54,16 @@ function try_get_filename_from_request(req)
 end
 
 
-determine_file(::Nothing, resp) = determine_file(tempdir(), resp)
+determine_file(::Nothing, resp, hdrs) = determine_file(tempdir(), resp, hdrs)
 # ^ We want to the filename if possible because extension is useful for FileIO.jl
 
-function determine_file(path, resp)
+function determine_file(path, resp, hdrs)
     # get the name
     name = if isdir(path)
         # we have been given a path to a directory
         # got to to workout what file to put there
         filename = something(
-                        try_get_filename_from_headers(resp),
+                        try_get_filename_from_headers(hdrs),
                         try_get_filename_from_request(resp.request),
                         basename(tempname())  # fallback, basically a random string
                     )
@@ -107,11 +106,15 @@ function download(url::AbstractString, local_path=nothing, headers=Header[]; upd
 
     @debug 1 "downloading $url"
     local file
+    hdrs = String[]
     HTTP.open("GET", url, headers; kw...) do stream
         resp = startread(stream)
+        # Store intermediate header from redirects to use for filename detection
+        content_disp = header(resp, "Content-Disposition")
+        !isempty(content_disp) && push!(hdrs, content_disp)
         eof(stream) && return  # don't do anything for streams we can't read (yet)
 
-        file = determine_file(local_path, resp)
+        file = determine_file(local_path, resp, hdrs)
         total_bytes = parse(Float64, header(resp, "Content-Length", "NaN"))
         downloaded_bytes = 0
         start_time = now()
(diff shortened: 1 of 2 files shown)


oxinabox:
Perhaps the ideal solution would be for a variant of `request` to return a collection of all headers from redirects

fredrikekre:
Would it be better to just move it if the initial guess was wrong?

fredrikekre:
I think I misunderstood. Is the case that the correct filename is only in the original 302 headers but *not* the actual 200 response header? That seems a bit strange of the server, no?

oxinabox:
> Is the case that the correct filename is only in the original 302 headers but not the actual 200 response header?

Correct.

>  That seems a bit strange of the server, no?

I am honestly not sure.
On the surface it sounds like one should expect the redirect to have this name.
But perhaps not.

The 302 comes from a FigShare server, it has a pretty rich header, with a fair bit of metadata.
It is probably running a bespoke webserver so they have a lot of control over what it tells us.
It redirected to web-facing AWS S3 Bucket, which are not very configurable at all.
Though it is possible to set it via the AWS commandline tool or API.

302 is nominally a temporary redirect.
So it could be that the AWS bucket is not the final storage location for this, and is a stop-gap til they set something up properly.


Regardless of reasonable or not, when i go to https://ndownloader.figshare.com/files/6294558
in FireFox, it does want to download with the filename `rsta20150293_si_001.xlsx`
and matching behavior of web-browser is basically my goal for `HTTP.download`

fredrikekre:
I believe you have access to the original headers before the early return btw, so don't think you need the HEAD request.

oxinabox:
> I believe you have access to the original headers before the early return btw, so don't think you need the HEAD request.

Really?
I couldn't find a way to get them
It looked to me like the redirect layer eats them up.

https://github.com/JuliaWeb/HTTP.jl/blob/81ac5042dd284dc0525702982f71683b2e168046/src/RedirectRequest.jl#L28

Notes that we need the `headers(res)` which are the **response** headers,
not the `headers` which are the **request** headers (which are forwarded onwards by default)

fredrikekre:
`resp` just before contains the intermediate headers:
```diff
diff --git a/src/download.jl b/src/download.jl
index f75be99..1d7bdef 100644
--- a/src/download.jl
+++ b/src/download.jl
@@ -16,9 +16,8 @@ function safer_joinpath(basepart, parts...)
     joinpath(basepart, parts...)
 end
 
-function try_get_filename_from_headers(resp)
-    content_disp = header(resp, "Content-Disposition")
-    if content_disp != ""
+function try_get_filename_from_headers(hdrs)
+    for content_disp in hdrs
         # extract out of Content-Disposition line
         # rough version of what is needed in https://github.com/JuliaWeb/HTTP.jl/issues/179
         filename_part = match(r"filename\s*=\s*(.*)", content_disp)
@@ -55,16 +54,16 @@ function try_get_filename_from_request(req)
 end
 
 
-determine_file(::Nothing, resp) = determine_file(tempdir(), resp)
+determine_file(::Nothing, resp, hdrs) = determine_file(tempdir(), resp, hdrs)
 # ^ We want to the filename if possible because extension is useful for FileIO.jl
 
-function determine_file(path, resp)
+function determine_file(path, resp, hdrs)
     # get the name
     name = if isdir(path)
         # we have been given a path to a directory
         # got to to workout what file to put there
         filename = something(
-                        try_get_filename_from_headers(resp),
+                        try_get_filename_from_headers(hdrs),
                         try_get_filename_from_request(resp.request),
                         basename(tempname())  # fallback, basically a random string
                     )
@@ -107,11 +106,15 @@ function download(url::AbstractString, local_path=nothing, headers=Header[]; upd
 
     @debug 1 "downloading $url"
     local file
+    hdrs = String[]
     HTTP.open("GET", url, headers; kw...) do stream
         resp = startread(stream)
+        # Store intermediate header from redirects
+        content_disp = header(resp, "Content-Disposition")
+        !isempty(content_disp) && push!(hdrs, content_disp)
         eof(stream) && return  # don't do anything for streams we can't read (yet)
 
-        file = determine_file(local_path, resp)
+        file = determine_file(local_path, resp, hdrs)
         total_bytes = parse(Float64, header(resp, "Content-Length", "NaN"))
         downloaded_bytes = 0
         start_time = now()
```

oxinabox:
Oh wow cool.
I clearly have no idea how `HTTP.open(f, "GET", url)` works.

oxinabox:
Thanks @fredrikekre  I meant to push those changes into  this branch, but hasn't got around to it yet.

Issue #547, opened 2020-06-23
Meaning and Handling of `connection reset by peer` Error


ueliwechsler:
I am using [https://github.com/wookay/Bukdu.jl](`Bukdu.jl`) for creating a REST web server.  

The application is running without problems but from time to time I get a`Base.IOError("read: connection reset by peer (ECONNRESET)", -4077)`  error message which was generated at https://github.com/JuliaWeb/HTTP.jl/blob/b9b42c3e4c75fe27f5126dbc61ae5c93aee9f56a/src/Servers.jl#L264 (see full error log below).

I have posted already an issue [https://github.com/wookay/Bukdu.jl/issues/105](url) in the `Bukdu` repo and also read this post on discourse  [https://discourse.julialang.org/t/econnreset-error-with-http-jl/11228](url) but I didn't understand the scope of the problem entirely  since my understanding of web protocol is basic at best.

Even though the application runs, I do feel uneasy to use it in production just like this. Do you have any more resources on why this error happens (apart from the source code 😅). Do I need to refactor the webserver or is it safe to use with this message occurring?

The versions are:
```julia
Julia 1.4.2
HTTP v0.8.15
MbedTLS v1.0.2
```
and the full error message is
```julia
(Base.IOError("read: connection reset by peer (ECONNRESET)", -4077), 
Base.StackTraces.StackFrame[wait_readnb(::Sockets.TCPSocket, ::Int64) at stream.jl:350, eof at stream.jl:48 [inlined], 
read_to_buffer(::HTTP.ConnectionPool.Transaction{Sockets.TCPSocket}, ::Int64) at ConnectionPool.jl:253, 
readuntil(::HTTP.ConnectionPool.Transaction{Sockets.TCPSocket}, ::Function, ::Int64) at ConnectionPool.jl:273, readuntil at ConnectionPool.jl:271
 [inlined], readheaders(::HTTP.ConnectionPool.Transaction{Sockets.TCPSocket}, ::HTTP.Messages.Request) at Messages.jl:471, 
startread(::HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPool.Transaction{Sockets.TCPSocket}}) at Streams.jl:155, 
[...]

PR #764, opened 2021-09-21
Catch ECONNRESET error codes from other platforms

### LINKED ISSUE #763: Server emitting errors and traceback on ECONNRESET when it should be a warning.
Using versions:
Julia 1.6
HTTP.jl 0.9.14

It appears the intention of this code in [Servers.jl is to throw a warning rather than an error when the connection is reset by the peer](https://github.com/JuliaWeb/HTTP.jl/blob/87d29d9e11730b8ccefce487c437e8d8d490f249/src/Servers.jl#L293) (ie, client closed the connection to our server).

However this is done by a check on the error code. And this code seems to change based on the architecture of the system running. On my system, ECONNRESET corresponds with e.code of `-104`. 

See this list of search results for "connection reset by peer" in the linux source code:
https://github.com/torvalds/linux/search?p=1&q=%22connection+reset+by+peer%22&type=code

Notice lots of different `errno.h` files.
* For SPARC and ALPHA architectures, ECONNRESET is `54` (as per the code in Julia)
* For Generic architecture, ECONNERESET is `104` (this is what my system is, and probably 99% of users)
* For PARISC architecture, ECONNRESET is `232`
* For MIPS architecture, ECONNRESET is `131`

Note, when the TCP driver throws these exceptions, it makes them negative, so Julia must match on the negative value, eg `-104`.

I don't know why `-4077` is in the code, it doesn't match any error number I could find, maybe thats the one windows uses?

Related to https://github.com/JuliaWeb/HTTP.jl/issues/547
Tangentially related to https://github.com/JuliaWeb/HTTP.jl/issues/657 (in the sense that an error is propagated in a non-actionable way)

ashleysommer:
Windows uses -4077
SPAC and ALPHA architectures on linux use -54
Generic architectures on linux uses -104 (this includes x86, x86_64, arm, and arm64)
PARSIC on linux uses -232
MIPS on linux uses -131
I'm guessing MacOS uses -54 too, as this was one code that already worked for a large group of people.
Fixes #763 #547

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -303,7 +303,8 @@ function listenloop(f, server, tcpisvalid, connection_count,
                 handle_connection(f, conn, server, reuse_limit, readtimeout)
                 # verbose && @info "Closed ($count):  $conn"
             catch e
-                if e isa Base.IOError && (e.code == -54 || e.code == -4077)
+                if e isa Base.IOError &&
+                    (e.code == -54 || e.code == -4077 || e.code == -104 || e.code == -131 || e.code == -232)
                     verbose && @warn "connection reset by peer (ECONNRESET)"
                 else
                     @error "" exception=(e, stacktrace(catch_backtrace()))


quinnj (approved):
Thanks for this!

Issue #765, opened 2021-09-24
Cookies not saved because case-sensitive Set-Cookie header?


RuiRojo:
Julia 1.6.2 
HTTP.jl 0.9.14
MbedTLS.jl 1.0.3

I make a POST request to a site (to start a session) but the returned cookie isn't saved. I don't know much about all this so, I may be making a dumb mistake. But, after debugging, I think the issue is that the function `setcookies` is expecting the `Response` header to be named "Set-Cookie" when what I'm getting is not capitalized, "set-cookie". Maybe this should be case-insensitive?

(If I'm off the mark but it's obvious to someone what I may be doing wrong, I appreciate)

pfitzseb:
Yes, header names are supposed to be case insensitive. Wanna open a PR?

RuiRojo:
Ok, I [gave it a shot](https://github.com/JuliaWeb/HTTP.jl/pull/766)

PR #766, opened 2021-09-24
Treat set-cookie header as case insensitive

RuiRojo:
Fix https://github.com/JuliaWeb/HTTP.jl/issues/765

### DIFF
--- a/src/CookieRequest.jl
+++ b/src/CookieRequest.jl
@@ -4,6 +4,7 @@ import ..Dates
 import ..Layer, ..request
 using URIs
 using ..Cookies
+using ..Messages: ascii_lc_isequal
 using ..Pairs: getkv, setkv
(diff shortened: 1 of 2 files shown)


src/CookieRequest.jl
```diff
@@ -72,7 +72,7 @@ function getcookies(cookies, url)
 end
 
 function setcookies(cookies, host, headers)
-    for (k,v) in filter(x->x[1]=="Set-Cookie", headers)
+    for (k,v) in filter(x->lowercase(x[1])=="set-cookie", headers)
```
quinnj, line 75:
```suggestion
    for (k,v) in filter(x->ascii_lc_isequal(x[1], "set-cookie"), headers)
```
we have an optimized function for comparing two ascii strings case insensitively

fredrikekre:
Possible to add a test?

PR #752, opened 2021-09-08
Do not call `close()` on the `response_stream`

c42f:
This allows a plain `IOBuffer` to be used with `response_stream`.

It seems unnecessary to be calling `close(response_stream)` inside
HTTP.request() - given that HTTP.request is a blocking operation, the
caller can easily close the stream themselves after `HTTP.request()`
returns.

This could be considered a breaking change for users who use the
internal IO type Base.BufferStream with the response_stream keyword.

Fixes #543

PR #773, opened 2021-09-29
Revert "Do not call close..." and release 0.9.16

### LINKED ISSUE #772: Requests to AWS S3 hang with HTTP.jl v0.9.15
With `HTTP@0.9.15`, i'm seeing GET requests to S3 hang indefinitely.

The actual behaviour i am seeing is using AWSS3.jl v0.9.2 (example at https://github.com/JuliaCloud/AWSS3.jl/issues/215), but the behaviour is only present when using HTTP v0.9.15. 

I haven't quite narrowed down what's going on... but here's a bit of thinking out loud:
- `AWSS3.s3_get_file`  calls `get_object` (which is a AWS.jl call) 
- which in turn calls `AWS.s3("GET", "/$(Bucket)/$(Key)"; aws_config)`
- which is calling a `AWS.RestXMLService` object...
- which finally constructs a `HTTP.Request` and calls `HTTP.request`
    - construct `HTTP.Request`: https://github.com/JuliaCloud/AWS.jl/blob/564f9c55291a29c6994bf39625912256bf2e9c4f/src/AWS.jl#L223-L232) 
    - via `submit_request`: https://github.com/JuliaCloud/AWS.jl/blob/9a37112f63cc38a59f6b255bb1aff74fc697eec5/src/utilities/request.jl#L59-L84
    - calls `HTTP.request`: https://github.com/JuliaCloud/AWS.jl/blob/9a37112f63cc38a59f6b255bb1aff74fc697eec5/src/utilities/request.jl#L186-L196
    
Looking at the 0.9.15 changes (https://github.com/JuliaWeb/HTTP.jl/compare/v0.9.14...v0.9.15) i'd guess the most related change might be https://github.com/JuliaWeb/HTTP.jl/pull/752 ?

Needs further investigation :)

fredrikekre:
Since #752 turned out to be rather breaking I propose a new release with that change reverted. We should also yank 0.9.15 from the registry.

Fixes #772 

cc @oxinabox @mattBrzezinski @nickrobinson251 @c42f @quinnj @omus


oxinabox (approved):
I approve this.
To be clear I think it is entirely possible that #752 **is** the correct thing to do.
But I think we would like have a bit more time to think about it, 
and reverting it and tagging a patch gives us that time.

CHANGELOG.md
```diff
@@ -6,11 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.9.16] - 2021-09-29
+See changes for 0.9.15: this release is equivalent to 0.9.15 with [#752] reverted.
```
omus, line 10: Mention the follow up issue here

omus (approved):
I'm okay with reverting this change as long as we create an issue for removing `close` as part of a breaking release and mention it in the change logs.

oxinabox:
> long as we create an issue for removing close as part of a breaking release and mention it in the change logs.

That is one possibility, but let's take the time to discuss it and think it through fully.
Which can wait til we open that issue.

omus:
> That is one possibility, but let's take the time to discuss it and think it through fully. Which can wait til we open that issue.

Basically I want to make that issue now

c42f:
This is fine by me in terms of keeping the stable releases going.

I still think the original patch was the correct approach, so I'll propose it again as a PR, and we can mark it breaking, and decide when to merge it.

c42f:
See https://github.com/JuliaWeb/HTTP.jl/pull/775

PR #778, opened 2021-11-15
Fix chance of incomplete read in partial response stream reading

IanButterworth:
This appears to fix a bug seen over in https://github.com/JuliaCloud/AWS.jl/issues/515 where a 206 partial request response has the correct Content-Length in the header but HTTP.jl returns fewer bytes with no error.

The bug is seen most frequently when running many `@async` `HTTP.request`s each requesting a data `Range` on slower internet connections.

The problem appears to be that a simple `write(response_stream, http)` bypassed the `ntoread` handling [here](https://github.com/JuliaWeb/HTTP.jl/blob/f359667aa6a7ea3c0436f1922c0b2f65d0717d81/src/Streams.jl#L312-L324) that ensures successive reads were done if the promised content length wasn't fully retrieved the first time, which allows for the buffer to populate slowly on slower connections, so it needs to be `write(response_stream, read(http))` instead

### DIFF
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -311,6 +311,15 @@ function Base.readbytes!(http::Stream, buf::IOBuffer, n=bytesavailable(http))
     buf.size += n
 end
 
+# used in julia v1.0
+function Base.readbytes!(http::Stream, buf::IOStream, n=bytesavailable(http))
+    nread = 0
+    while nread < n
+        nread += write(buf, readavailable(http, n - nread))
+    end
+    nread
+end
+
 function Base.read(http::Stream)
     buf = PipeBuffer()
     if ntoread(http) == unknown_length
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -303,6 +303,8 @@ function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
     nothing
 end
 
+Base.readbytes!(http::Stream, buf::Base.BufferStream, n=bytesavailable(http)) = Base.readbytes!(http, buf.buffer, n)
+
 function Base.readbytes!(http::Stream, buf::IOBuffer, n=bytesavailable(http))
     Base.ensureroom(buf, n)
     unsafe_read(http, pointer(buf.data, buf.size + 1), n)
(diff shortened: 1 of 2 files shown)


fredrikekre:
Doesn't this "undo" the streaming feature?

IanButterworth:
I see. Perhaps there's something else needed to prevent the write from finishing too early if more content is expected?

IanButterworth:
Changing this to
```
function readbody(http::Stream, res::Response, response_stream, reached_redirect_limit)
    if response_stream === nothing
        res.body = read(http)
    else
        if reached_redirect_limit || !isredirect(res)
            res.body = body_was_streamed
            while true
                write(response_stream, http)
                Streams.incomplete(http) || break
                @info "looping" http.ntoread bytesavailable(http.stream) isopen(http.stream)
                yield()
            end
            close(response_stream)
        end
    end
end
```

results in good behavior until it gets stuck with
```
┌ Info: looping
│   http.ntoread = 15441
│   bytesavailable(http.stream) = 0
└   isopen(http.stream) = false
┌ Info: looping
│   http.ntoread = 15441
│   bytesavailable(http.stream) = 0
└   isopen(http.stream) = false
┌ Info: looping
│   http.ntoread = 15441
│   bytesavailable(http.stream) = 0
└   isopen(http.stream) = false
```

IanButterworth:
@fredrikekre @iamed2 
> Doesn't this "undo" the streaming feature?

I think I just updated the fix to allow for streaming? Could someone run the CI please 🙏🏻

IanButterworth:
Tests passing on my fork https://github.com/IanButterworth/HTTP.jl/pull/1

quinnj (approved):
This LGTM; thanks @IanButterworth.

src/Streams.jl
```diff
@@ -303,12 +303,23 @@ function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
     nothing
 end
 
+Base.readbytes!(http::Stream, buf::Base.BufferStream, n=bytesavailable(http)) = Base.readbytes!(http, buf.buffer, n)
```
fredrikekre, line 306: Why did you have to special case `BufferStream` and `IOBuffer`? Will this fail with other IO now?

PR #780, opened 2021-11-16
Generalize readbytes!

IanButterworth:
Intending to address @fredrikekre's comment on https://github.com/JuliaWeb/HTTP.jl/pull/778/files#r750049062

Fixes unintentional IO handling narrowing that #778 introduced

Given `Base.readbytes!(http::Stream, buf::IOBuffer, n=bytesavailable(http))` already existed, I thought it best to leave `IOBuffer`-backed IO's to use that method, but I wasn't sure if there's a generic way to do that?

### DIFF
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -303,8 +303,6 @@ function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
     nothing
 end
 
-Base.readbytes!(http::Stream, buf::Base.BufferStream, n=bytesavailable(http)) = Base.readbytes!(http, buf.buffer, n)
-
 function Base.readbytes!(http::Stream, buf::IOBuffer, n=bytesavailable(http))
     Base.ensureroom(buf, n)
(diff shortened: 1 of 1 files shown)


IanButterworth:
Looking at the definition of `BufferStream` https://github.com/JuliaLang/julia/blob/bc8337ae1406a03a82cc21a8ddaa1ee93fa4151e/base/stream.jl#L1469 
I think it was wrong to bypass to the internal IOBuffer, so I've removed that method, and more can rely on the generic `readbytes!`

fredrikekre:
If you want to add a test: This errors on HTTP master (after https://github.com/JuliaWeb/HTTP.jl/pull/778) and seems to deadlock on this PR:
```
using HTTP, SimpleBufferStream
bs = BufferStream()
HTTP.get("https://julialang.org"; response_stream=bs)
```

fredrikekre:
Also seem to hang in tests, so I cancelled them.

IanButterworth:
Replaced by #781

PR #781, opened 2021-11-17
Revert #778. Make incomplete content-length responses throw an EOFError

IanButterworth:
- Reverts #778 (flawed.. this approach is cleaner)
- ~~Implements an `EOFError` throw if the `Base.eof(http::Stream)` method itself detects that eof has been reached before a promised `content-length` response has been delivered. That allows the existing `Base.write` methods to be used, so apart from the revert, the code change is much more minimal and general~~
- Fixes the `closeread` `iscomplete` check and EOFError throw to detect incomplete responses, which involves exposing if `closewrite` or `closeread` throw an EOFError, which was suppressed in https://github.com/JuliaWeb/HTTP.jl/pull/529

Should probably not be squash-merged, to keep the revert separate

Closes #780

### DIFF
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -205,10 +205,7 @@ function Base.eof(http::Stream)
     if http.ntoread == 0
         return true
     end
-    if eof(http.stream)
-        return true
-    end
-    return false
+    return eof(http.stream)
 end
 
 @inline function ntoread(http::Stream)
@@ -355,6 +352,8 @@ function IOExtras.closeread(http::Stream{Response})
         # Close conncetion if server sent "Connection: close"...
         @debug 1 "✋  \"Connection: close\": $(http.stream)"
         close(http.stream)
+        # Error if Message is not complete...
+        incomplete(http) && throw(EOFError())
     else
 
         # Discard body bytes that were not read...
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -303,21 +303,10 @@ function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
     nothing
 end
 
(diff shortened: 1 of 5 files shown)


fredrikekre:
Looks good Here is a test that you could add
```
using HTTP, Sockets, Test

server = Sockets.listen(ip"0.0.0.0", 8080)
task = @async HTTP.listen("0.0.0.0", 8080; server=server) do http
    HTTP.setstatus(http, 200)
    HTTP.setheader(http, "Content-Length" => "64") # Promise 64 bytes...
    HTTP.startwrite(http)
    HTTP.write(http, rand(UInt8, 63)) # ...but only send 63 bytes.
    # Close the stream so that eof(stream) is true and the client isn't
    # waiting forever for the last byte.
    HTTP.close(http.stream)
end

err = try
    HTTP.get("http://localhost:8080"; retry=false)
catch err
    err
end
@test err isa HTTP.IOError
@test err.e isa EOFError

# Shutdown
try; close(server); wait(task); catch; end
```

curl errors with
```
curl: (18) transfer closed with 1 bytes remaining to read
```
for this case so seems correct that HTTP.jl also errors.

src/Streams.jl
```diff
@@ -202,13 +202,20 @@ function Base.eof(http::Stream)
...
+        return eof(http.stream)
+    end
     if http.ntoread == 0
         return true
     end
     if eof(http.stream)
-        return true
+        # ntoread is neither 0 nor unknown_length here
+        # so eof was reached before fixed content-length delivered
+        @debug 1 "❗️  EOFError: Transfer closed with $(http.ntoread) bytes remaining to read   $(http.stream)"
```
fredrikekre, line 214: Does `http.stream` contain anything useful?
```suggestion
        @debug 1 "❗️  EOFError: Transfer closed with $(http.ntoread) bytes remaining to read"
```

IanButterworth:
Just tidied up commits so that this can be regular merged.

PR #789, opened 2021-12-15
Change internal layers stack to be value-based instead of type-based

quinnj:
This is very proof-of-concept currently, but wanted to put the code up
to get initial feedback/ideas. Here's the gist:

  * I define 3 abstract subtypes of `Layer` in the Layers module:
  `InitialLayer`, `RequestLayer`, and `ConnectionLayer`; these
  correspond to the types of arguments the layer would receive when
  overloading `request`; `InitialLayer` => `request(method, url,
  headers, body), `RequestLayer` => `request(url, req, body)`, and
  `ConnectionLayer` => `request(io, req, body)`. I feel like these are
  useful abstract types to help distinguish the different _kinds_ of
  layers based on the arguments they receive in different parts of the
  overall stack. I feel like it hopefully also solves the "how do I
  insert my layer in the appropriate level of the stack" because you
  just pick which abstract layer to subtype
  * Custom layers are then required to:
    * be concrete types
    * must subtype one of the 3 abstract layer types mentioned
    * must have a constructor of the form: `Layer(next::Layer; kw...)
    * must have a field to store the `next::Layer` argument
    * must overload: `request(x::MyLayer, args...; kw...)` where `args`
    will depend on which abstract layer `MyLayer` subtypes
    * in the overloaded `request` method, it must, at some point, call
    `request(layer.next, args...; kw...)` to move to the next layer in
    the stack
    * the final requirement is the custom layer must overload
    `Layers.keywordforlayer(::Val{:kw}) = MyLayer` where `:kw` is a
    keyword argument "hook" or trigger that will result in `MyLayer`
    being inserted into the stack

What I like about this approach so far:
  * It feels more straightforward to define my own custom layer: pick
  which level of the stack I want it in, subtype that abstract layer,
  register a keyword argument that will include my layer, then define my
  own `request` overload; we can define all the current HTTP layers in
  terms of this machinery, custom packages could define their own layers
  and they all are treated just like any HTTP.jl-baked layer
  * It also feels like it would be easier to create my own custom "http
  client"; what I mean by this is that I feel like it's common for API
  packages to basically have their own `AWS.get`/`AWS.post`/`AWS.put`
  wrapper methods that eventually call `HTTP.get`/`HTTP.post`/etc., but
  in between, they're adding custom headers, validating targets, and
  whatever. A lot of that is _like_ a custom layer, but by having your
  own `AWS.get`/`AWS.post` wrapper methods, you can _ensure_ certain
  layers are included; and with this approach that's straightforward
  because they can indeed make their own custom layer as described
  above, and then just ensure every `HTTP.request` call includes the new
  keyword argument to "hook in" their custom layer

Potential downsides:
  * Relying on a global keyword registry; i.e. there's potential for
  clashes if two different packages try to overload
  `Layers.keywordforlayer(::Val{:readtimeout}) = ...`. Whichever package
  is loaded last will "pirate" the keyword arg and overwrite the
  previous method. We could potentially add some extra safety around
  this by having another `clientenv=:default` keyword and then devs
  could overload `Layers.keywordforlayer(::Val{:default},
  ::Val{:readtimeout}) = ...` but that feels a tad icky. I also don't
  expect there to be even dozens of total keywords/layers to process, so
  with that in mind, it also lowers the risk of clash.
  * The previous layer functionality allowed specifying the _exact_
  layer you wanted your custom layer to be inserted before; do we think
  that level of specificity/granularity is really necessary? I'm kind of
  going on a gut feel here that what you really want control over is
  _what kind of args_ your overloaded `request` method will receive and
  that's what my 3 new abstract layer subtypes aim to solve. So you can
  pick 1 of 3 different _general_ layers of where your custom layer gets
  inserted, but don't have control over anything more specific in terms
  of what layer comes before/after
  * One thing I noticed is that the current `stack` function actually
  has some hard-coded logic that checks keyword arguments and
  _conditionally_ will include layers based on the _value_ of the
  keyword arg. We could potentially handle that in the proposed scheme
  by saying that your custom `Layer(next; kw...)` constructor can
  return the `nothing` value, in which case, no layer will be inserted.
  That would give, for example, the `TimeoutLayer` the chance to check
  if the `readtimeout > 0` before returning a `TimeoutLayer` or
  `nothing`

### DIFF
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -453,7 +453,6 @@ include("ConnectionRequest.jl");        using .ConnectionRequest
 include("DebugRequest.jl");             using .DebugRequest
 include("StreamRequest.jl");            using .StreamRequest
 include("ContentTypeRequest.jl");       using .ContentTypeDetection
-include("exceptions.jl")
 
 """
 The `stack()` function returns the default HTTP Layer-stack type.
@@ -569,39 +568,23 @@ relationship with [`HTTP.Response`](@ref), [`HTTP.Parsers`](@ref),
 ```
 *See `docs/src/layers`[`.monopic`](http://monodraw.helftone.com).*
 """
-function stack(;redirect=true,
-                aws_authorization=false,
-                cookies=false,
-                canonicalize_headers=false,
-                retry=true,
-                status_exception=true,
-                readtimeout=0,
-                detect_content_type=false,
-                verbose=0,
-                kw...)
+function stack(; kw...)
 
-    NoLayer = Union
-    stack =                 TopLayer{
-    (redirect             ? RedirectLayer             : NoLayer){
-                            BasicAuthLayer{
-    (detect_content_type  ? ContentTypeDetectionLayer : NoLayer){
-    (cookies === true || (cookies isa AbstractDict && !isempty(cookies)) ?
-                            CookieLayer               : NoLayer){
-    (canonicalize_headers ? CanonicalizeLayer         : NoLayer){
-                            MessageLayer{
-    (aws_authorization    ? AWS4AuthLayer             : NoLayer){
-    (retry                ? RetryLayer                : NoLayer){
-    (status_exception     ? ExceptionLayer            : NoLayer){
-                            ConnectionPoolLayer{
-    (verbose >= 3 ||
-     DEBUG_LEVEL[] >= 3   ? DebugLayer                : NoLayer){
-    (readtimeout > 0      ? TimeoutLayer              : NoLayer){
-                            StreamLayer{Union{}}
-    }}}}}}}}}}}}}
+    layers = stacklayertypes(Layers.ConnectionLayer, StreamLayer(); kw...)
+    layers = ConnectionPoolLayer(layers; kw...)
+    layers = stacklayertypes(Layers.RequestLayer, layers; kw...)
+    layers = MessageLayer(layers; kw...)
+    return stacklayertypes(Layers.InitialLayer, layers; kw...)
+end
 
-    reduce(Layers.EXTRA_LAYERS; init=stack) do stack, (before, custom)
-        insert(stack, before, custom)
+function stacklayertypes(::Type{T}, layers; kw...) where {T}
+    for (k, _) in pairs(kw)
+        layer = Layers.keywordforlayer(Val(k))
+        if layer !== nothing && layer <: T
+            layers = layer(layers; kw...)
+        end
     end
+    return layers
 end
 
 include("download.jl")
--- a/src/MessageRequest.jl
+++ b/src/MessageRequest.jl
@@ -23,14 +23,17 @@ resource(uri::URI) = string( isempty(uri.path)     ? "/" :     uri.path,
 
 Construct a [`Request`](@ref) object and set mandatory headers.
 """
-struct MessageLayer{Next <: Layer} <: Layer{Next} end
+struct MessageLayer{Next <: Layer} <: RequestLayer
+    next::Next
+end
 export MessageLayer
+MessageLayer(next; kw...) = MessageLayer(next)
 
-function request(::Type{MessageLayer{Next}},
+function request(layer::MessageLayer,
                  method::String, url::URI, headers::Headers, body;
                  http_version=v"1.1",
                  target=resource(url),
-                 parent=nothing, iofunction=nothing, kw...) where Next
+                 parent=nothing, iofunction=nothing, kw...)
 
     if isempty(url.port) ||
               (url.scheme == "http" && url.port == "80") ||
@@ -63,7 +66,7 @@ function request(::Type{MessageLayer{Next}},
     req = Request(method, target, headers, bodybytes(body);
                   parent=parent, version=http_version)
 
-    return request(Next, url, req, body; iofunction=iofunction, kw...)
+    return request(layer.next, url, req, body; iofunction=iofunction, kw...)
 end
 
 const USER_AGENT = Ref{Union{String, Nothing}}("HTTP.jl/$VERSION")
--- a/src/StreamRequest.jl
+++ b/src/StreamRequest.jl
@@ -19,15 +19,15 @@ immediately so that the transmission can be aborted if the `Response` status
 indicates that the server does not wish to receive the message body.
 [RFC7230 6.5](https://tools.ietf.org/html/rfc7230#section-6.5).
 """
-abstract type StreamLayer{Next <: Layer} <: Layer{Next} end
+struct StreamLayer <: ConnectionLayer end
 export StreamLayer
 
-function request(::Type{StreamLayer{Next}}, io::IO, req::Request, body;
+function request(::StreamLayer, io::IO, req::Request, body;
                  reached_redirect_limit=false,
                  response_stream=nothing,
                  iofunction=nothing,
                  verbose::Int=0,
-                 kw...)::Response where Next
+                 kw...)::Response
 
     verbose == 1 && printlncompact(req)
 
@@ -84,7 +84,7 @@ function request(::Type{StreamLayer{Next}}, io::IO, req::Request, body;
     verbose == 1 && printlncompact(response)
     verbose == 2 && println(response)
 
-    return request(Next, response)
+    return response
 end
 
 function writebody(http::Stream, req::Request, body)
--- a/src/ConnectionRequest.jl
+++ b/src/ConnectionRequest.jl
@@ -57,12 +57,16 @@ Otherwise leave it open so that it can be reused.
 `IO` related exceptions from `Base` are wrapped in `HTTP.IOError`.
 See [`isioerror`](@ref).
 """
-abstract type ConnectionPoolLayer{Next <: Layer} <: Layer{Next} end
+struct ConnectionPoolLayer{Next <: Layer} <: RequestLayer
+    next::Next
+    pool::ConnectionPools.Pool
+end
 export ConnectionPoolLayer
+ConnectionPoolLayer(next; connectionpool=ConnectionPool.POOL) = ConnectionPoolLayer(next, connectionpool)
 
-function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
+function request(layer::ConnectionPoolLayer, url::URI, req, body;
                  proxy=getproxy(url.scheme, url.host),
-                 socket_type::Type=TCPSocket, kw...) where Next
+                 socket_type::Type=TCPSocket, kw...)
 
     if proxy !== nothing
         target_url = url
@@ -99,7 +103,7 @@ function request(::Type{ConnectionPoolLayer{Next}}, url::URI, req, body;
             req.headers = filter(x->x.first != "Proxy-Authorization", req.headers)
         end
 
-        r =  request(Next, io, req, body; kw...)
+        r =  request(layer.next, io, req, body; kw...)
 
         if proxy !== nothing && target_url.scheme == "https"
             close(io)
--- a/src/TimeoutRequest.jl
+++ b/src/TimeoutRequest.jl
@@ -17,12 +17,17 @@ end
 
 Close `IO` if no data has been received for `timeout` seconds.
 """
-abstract type TimeoutLayer{Next <: Layer} <: Layer{Next} end
+struct TimeoutLayer{Next <: Layer} <: ConnectionLayer
+    next::Next
+    readtimeout::Int
+end
 export TimeoutLayer
+Layers.keywordforlayer(::Val{:readtimeout}) = TimeoutLayer
 
-function request(::Type{TimeoutLayer{Next}}, io::IO, req, body;
-                 readtimeout::Int=0, kw...) where Next
+TimeoutLayer(next; readtimeout::Int=0, kw...) = TimeoutLayer(next, readtimeout)
 
+function request(layer::TimeoutLayer, io::IO, req, body; kw...)
+    readtimeout = layer.readtimeout
     wait_for_timeout = Ref{Bool}(true)
     timedout = Ref{Bool}(false)
 
@@ -37,7 +42,7 @@ function request(::Type{TimeoutLayer{Next}}, io::IO, req, body;
     end
 
     try
-        return request(Next, io, req, body; kw...)
+        return request(layer.next, io, req, body; kw...)
     catch e
         if timedout[]
            throw(ReadTimeoutError(readtimeout))
--- a/src/TopRequest.jl
+++ b/src/TopRequest.jl
@@ -10,7 +10,7 @@ export TopLayer
 This layer is at the top of every stack, and does nothing.
 It's useful for inserting a custom layer at the top of the stack.
 """
-abstract type TopLayer{Next <: Layer} <: Layer{Next} end
+struct TopLayer{Next <: Layer} <: Layer{Next} end
 
 request(::Type{TopLayer{Next}}, args...; kwargs...) where Next =
     request(Next, args...; kwargs...)
--- a/src/exceptions.jl
+++ b/src/exceptions.jl
@@ -1,7 +1,0 @@
-struct LayerNotFoundException <: Exception
-    var::String
-end
-
-function Base.showerror(io::IO, e::LayerNotFoundException)
-    println(io, typeof(e), ": ", e.var)
-end
--- a/src/layers.jl
+++ b/src/layers.jl
@@ -1,64 +1,29 @@
 module Layers
-export Layer, next, top_layer, insert, insert_default!, remove_default!
+export Layer, keywordforlayer
 
-const EXTRA_LAYERS = Set{Tuple{Union{UnionAll, Type{Union{}}}, UnionAll}}()
+struct LayerNotFoundException <: Exception
+    var::String
+end
 
-include("exceptions.jl")
+function Base.showerror(io::IO, e::LayerNotFoundException)
+    println(io, typeof(e), ": ", e.var)
+end
 
-"""
-## Request Execution Stack
-
-The Request Execution Stack is separated into composable layers.
-
-Each layer is defined by a nested type `Layer{Next}` where the `Next`
-parameter defines the next layer in the stack.
-The `request` method for each layer takes a `Layer{Next}` type as
-its first argument and dispatches the request to the next layer
-using `request(Next, ...)`.
-
-The example below defines three layers and three stacks each with
-a different combination of layers.
-
-
-```julia
 abstract type Layer end
-abstract type Layer1{Next <: Layer} <: Layer end
-abstract type Layer2{Next <: Layer} <: Layer end
-abstract type Layer3 <: Layer end
 
-request(::Type{Layer1{Next}}, data) where Next = "L1", request(Next, data)
-request(::Type{Layer2{Next}}, data) where Next = "L2", request(Next, data)
-request(::Type{Layer3}, data) = "L3", data
+abstract type InitialLayer <: Layer end
+abstract type RequestLayer <: Layer end
+abstract type ConnectionLayer <: Layer end
 
-const stack1 = Layer1{Layer2{Layer3}}
-const stack2 = Layer2{Layer1{Layer3}}
-const stack3 = Layer1{Layer3}
-```
+function keywordforlayer end
 
-```julia
-julia> request(stack1, "foo")
-("L1", ("L2", ("L3", "foo")))
-
-julia> request(stack2, "bar")
-("L2", ("L1", ("L3", "bar")))
-
-julia> request(stack3, "boo")
-("L1", ("L3", "boo"))
-```
-
-This stack definition pattern gives the user flexibility in how layers are
-combined but still allows Julia to do whole-stack compile time optimisations.
-
-e.g. the `request(stack1, "foo")` call above is optimised down to a single
-function:
-```julia
-julia> code_typed(request, (Type{stack1}, String))[1].first
-CodeInfo(:(begin
-    return (Core.tuple)("L1", (Core.tuple)("L2", (Core.tuple)("L3", data)))
-end))
-```
-"""
-abstract type Layer{Next} end
+keywordforlayer(kw) = nothing
+# custom layers must subtype one of above
+# must register a keyword arg for layer
+# must have a layer constructor like: Layer(next; kw...)
+# must have a field to store `next` layer
+# must overload: request(layer::MyLayer, args...; kw...)
+# in `request` overload, must call: request(layer.next, args...; kw...)
 
 """
     next(::Type{S}) where {T, S<:Layer{T}}


src/HTTP.jl
```diff
@@ -569,39 +568,23 @@ relationship with [`HTTP.Response`](@ref), [`HTTP.Parsers`](@ref),
...
+    layers = stacklayertypes(Layers.ConnectionLayer, StreamLayer(); kw...)
+    layers = ConnectionPoolLayer(layers; kw...)
+    layers = stacklayertypes(Layers.RequestLayer, layers; kw...)
+    layers = MessageLayer(layers; kw...)
+    return stacklayertypes(Layers.InitialLayer, layers; kw...)
+end
+
+function stacklayertypes(::Type{T}, layers; kw...) where {T}
+    for (k, _) in pairs(kw)
+        layer = Layers.keywordforlayer(Val(k))
```
fredrikekre, line 582: I am not sure if I understood the design here, but in order to insert a new layer you need to 1) overload this method, and 2) pass the matching kwarg?

fredrikekre:
> The previous layer functionality allowed specifying the exact layer you wanted your custom layer to be inserted before; do we think that level of specificity/granularity is really necessary? [...]

I think it is nice that the order is deterministic at least.

quinnj:
@fredrikekre, the other consideration is I haven't looked into how exactly this interacts with #469; I admit to not fully remembering the motivation for that issue or if that can be folded into the changes here somehow.

fredrikekre:
The motivation was that #463 where calls to `HTTP.request` could unexpectedly dispatch directly to internal functions. Regardless of that, I think it makes sense to have a single entry point (HTTP.request) that is separated from the internal machiner (e.g. `Layers.request` or even `HTTP._request` or something like that).

quinnj:
Ok, sounds good. Do you mind if I hijack those changes and fold them into my PR here? It's pretty deeply tangled in the stuff I'm doing, so might as well?

fredrikekre:
Go ahead.

quinnj:
Ok, quick update. This PR now includes the main changes from #469 since the 2 PRs basically touch all the same files/lines. I also went ahead and removed the AWSLayer (#486, cc: @mattBrzezinski ) since it was also intertwined with all the layer code updates.

I like how the code has shaped up; all the test suite passes locally for me, so let's see how CI does. My plan from here is to request feedback/review from folks (@christopher-dG, @mattBrzezinski, and @c42f), I'm going to take a deeper look at BrokenRecord.jl, which seems to be the biggest public package that heavily uses the current-master stack machinery, and then I also want to do a thorough update of the examples code and README where the custom layer/stack functionality are mentioned, so update all the docs around this proposed new stack architecture.

src/HTTP.jl
quinnj: Yes, that's the idea. The only "baked in" layers then are those where the keyword args included in this `stack` function call. I've also updated the mechanism (see below) so that if the layer constructor returns `nothing` it isn't included in the stack.

fredrikekre: I find it slightly strange that you need to do both of these things. If you need to modify the call to HTTP.request anyway (to add the kwarg) why not pass the layer directly?

quinnj: That would be pretty inconvenient I think; it would be the difference between:
```julia
HTTP.get(...; retry=true) # includes the RetryLayer based on retry keyword "hook"
```
vs.
```julia
HTTP.get(...; layer=HTTP.RetryLayer(; retry=true))
```
There's awkwardness there because how do we "pass" the layer? We can't do `retry=HTTP.RetryLayer()` unless we loop through all keyword arguments and check if the value is a `Layer` or not. And then there's awkwardness because what if the layer author wants to use a keyword argument as an argument to their layer constructor? Then users have to be burdened with the responsibility of knowing whether to pass args/keyword args to the layer constructor or as a keyword arg to `HTTP.request`.

In my proposal, everything is keyword args and users don't have to be aware of layers at all.

fredrikekre: Right, but it becomes a bit of "keyword argument piracy". Right now AWS calls `HTTP.request` with their own stack I think? So in that case, even if I have loaded the aws package I won't accidentally use that layer. BrokenRecords.jl already does "layer piracy" I think?

quinnj: I think cases like AWS though shouldn't _need_ to provide an entire custom stack. They should be able to just define their own custom layer or two and then it all composes with the "base" stack in HTTP.jl. The problem with everyone providing their own stacks is there isn't any composability. There isn't a nice way to use AWS custom layer along with my own custom layer in my company's auth system.

I'm going to take some more time to study the AWS and BrokenRecord setups now that this PR is "done", because I want to make sure I understand those use-cases thoroughly and how they fit in this potential new paradigm.

fredrikekre: Right, the current solution isn't ideal either, just not sure the kwargs based one is better.

quinnj: Looking back over the previous layer-manipulation code vs. the proposed here and there _is_ a key important difference: the previous code led to the scenario where you were modifying a single, global, and I'll add, non-threadsafe, client request stack by inserting/removing a custom layer for a request. Ok, you weren't modifying the _whole_ stack, but it was still a single global "extra custom layers" stack. That doesn't lend itself to concurrent code needs at all!

In my proposal there _is_ no single global stack; but rather on a per request basis, and with the user-supplied keyword arguments, the stack is dynamically generated and used for the duration of that request, in isolation from any other callers.

I partly discovered this by coming back to this PR and wondering if we shouldn't just use a single global `Dict` to "register" keyword arguments instead of overloading the `Layers.keywordforlayer` with a `Val`, but then again, we could get into thread-safety issues if that keyword registry is being modified concurrently.

So while I still agree the `Layers.keywordforlayer` mechanism is somewhat ugly, it _does_ have several advantages now that make it simpler in terms of implmentation.

quinnj:
Ok, I've been mulling the entire client request mechanisms again and came up w/ a modified proposal than initially posted in the PR above. I spent some time detailing everything in the new `Layers` module docs in this PR if you want to dive deep, but roughly, it works as follows:
  * Still have the new concept of layer "kinds": `InitialLayer`, `RequestLayer`, `ConnectionLayer`, and `ResponseLayer`
  * "Builtin" and custom layers choose one of those to subtype; must have constructor like `CustomLayer(next; kw...)` where `next` is the next layer in the stack, layer MUST have a field to store this `next` layer, MUST overload like `Layers.request(layer::CustomLayer, args...)`, and must call `Layers.request(layer.next, args...)` at some point in their overload (you can look at each XRequest.jl file for how these requirements look for existing internal HTTP.jl layers)
  * Once a custom layer is properly defined, we really just need a way to "hook it" into the stack; you could do it manually by calling something like
  
  ```julia
  resp = HTTP.request(HTTP.stack(CustomLayer), "GET", "https://google.com")
  ```
  
  But it obviously would get annoying having to include the `HTTP.stack(CustomLayer)` piece in every call.
  * A `HTTP.@client CustomLayer` macro is provided that allows quickly/easily making your own "custom" client that will include your `CustomLayer` in every call; so you would use `HTTP.@client` like:
  
  ```julia
  module CustomClient

using HTTP
include("customlayer_definitions.jl")
HTTP.@client CustomLayer

end
  ```
and users could then easily tab into the new `CustomLayer` functionality by doing:
```julia
using CustomClient

resp = CustomClient.get("https://google.com")
```
Basically, `HTTP.@client` makes it really easy to "make your own client" by providing those common, user-facing methods that will just call back into the generic `HTTP.request` method and will include any custom layer automatically.

So notably this removes the `Layers.keywordforlayer` "registry" function I had proposed above and in place either requires passing a custom layer manually (to `HTTP.stack` and `HTTP.request`), or by creating your own little "client" that will wrap all the client calls w/ your custom layer.

christopher-dG:
So if I'm reading this right, anyone wanting to use BrokenRecord would need to update their source code to use BrokenRecord.get instead of HTTP.get? My ideal case is that the test dependency doesn't leak into the package the way Mocking.jl does.

quinnj:
> So if I'm reading this right, anyone wanting to use BrokenRecord would need to update their source code to use BrokenRecord.get instead of HTTP.get? My ideal case is that the test dependency doesn't leak into the package the way Mocking.jl does.

Can you expound on what you mean by "doesn't leak into the package"? Perhaps share a small example of code and where it would live (either `/src/` or `/test`)?

Part of the motivation in this latest proposal is to make it very explicit when custom layers are being integrated into the request stack. You can imagine this being a security issue in the future if someone were able to "inject" a malicious layer. In the latest proposal, layers must be explicitly included either via manual call to `HTTP.stack` or by using a custom client. I personally like the transparency that will come w/ this new mechanism.

iamed2:
> You can imagine this being a security issue in the future if someone were able to "inject" a malicious layer.

I believe this kind of injection would be what @christopher-dG wants to do with BrokenRecord.jl, where it would be possible to monitor a package using HTTP.get without that package having to depend on BrokenRecord.jl.

quinnj:
> I believe this kind of injection would be what @christopher-dG wants to do with BrokenRecord.jl, where it would be possible to monitor a package using HTTP.get without that package having to depend on BrokenRecord.jl.

I just don't follow what this is supposed to mean. Monitoring a package without that package taking a dependency sounds......malicious? Undesirable? Wrong? What am I missing here.

iamed2:
> I just don't follow what this is supposed to mean. Monitoring a package without that package taking a dependency sounds......malicious? Undesirable? Wrong? What am I missing here.

Difference of philosophy? This is very common in dynamic languages, it's just [monkey patching](https://en.wikipedia.org/wiki/Monkey_patch). But this BrokenRecord package would likely not be part of a typical dependency tree, it would sit as a test dependency. Compare https://github.com/vcr/vcr and (coincidentally?) Cassette.jl. AWS also provides a [Python tool](https://docs.aws.amazon.com/xray-sdk-for-python/latest/reference/thirdparty.html) that patches requests and httplib to add X-Ray support, which is pretty similar to this. 

An actually malicious Julia package could pretty easily replace all code in another package anyway.

clarkevans:
In this proposal would "body" be an IOStream?  In this way, a client or server could process data incrementally.

clarkevans:
I'd shy away from naming filters in each layer by keyword and just have the programmer manually specify the stack.  If one goes with kw based naming of filters within a given layer, one probably has to specify a way for one filter to mark that it comes before/after another. For example, server side, some filters in the response layer may require that the user is already authenticated, and hence, those layers may want to have a "after=auth" mark, so that they are invoked after authorization.

c42f:
> I just don't follow what this is supposed to mean. Monitoring a package without that package taking a dependency sounds......malicious? Undesirable? Wrong? What am I missing here.

It's about composability.

Suppose you're using `BrokenRecord` to generate test cases for `SomeApplication`. Within `SomeApplication` you make some requests directly using `HTTP`, but also use several third-party modules `OtherModule1` `OtherModule2` etc which also make `HTTP` requests.

The proposal with `HTTP.@client` yields a customized HTTP client which is lexically scoped within the module of `SomeApplication`. So it's fairly easy to add `BrokenRecord` as a dependency within `SomeApplication` itself. Though I think undesirable if it can't be made optional — it's a package for testing which you don't want to load the code for in production.

The bigger problem is capturing HTTP requests from `OtherModule1` and `OtherModule2`. If these other modules use a separate `HTTP.@client` which is module-scoped it's impossible to inject the `BrokenRecord` dependency.

If we removed the default global HTTP client entirely and encouraged all libraries to accept a HTTP client context as a parameter this would "fix" the problem. But obviously that's intrusive on APIs using HTTP anywhere.


An analogy for the dependency injection which is desired here is logging: it's nice to have a logging frontend which is independent of the logging backends. Authors of various modules can independently use `@info` without caring about the fact that some users want to send their logs to the console, while other users want to send them to a file and others want to store them on AWS.

c42f:
> An analogy for the dependency injection which is desired here is logging

Taking this analogy further, the best way to manage the default HTTP stack may be with dynamic scope (the same as the logging backend). Then things like `BrokenRecord` could even be used concurrently without disturbing other tasks using HTTP in the same Julia process. Unfortunately Julia's task local storage isn't inherited across `@async` / `@spawn` which makes this idea a bit useless :-(  So I think we'll have to settle for global state for this.

By the way, I can see the benefit of a lexically scoped `HTTP.@client` for many cases. I just think it fails for things like BrokenRecord.jl and HTTPDownloads.jl which seem to require more dynamic dependency injection.

quinnj:
> Suppose you're using `BrokenRecord` to generate test cases for `SomeApplication`. Within `SomeApplication` you make some requests directly using `HTTP`, but also use several third-party modules `OtherModule1` `OtherModule2` etc which also make `HTTP` requests.

Aha. Ok. This finally makes more sense. Sorry for being slow and not understanding exactly how packages use BrokenRecord.jl. Ok, let me think on this some more and how it plays into the new architecture i'm playing with. Stay tuned. I think there's probably still a way we can have a global state option for changing things while keeping the proposed changes.

quinnj:
Ok, so for those following along this saga, let me give an update as I just pushed yet another big refactor of how the client-side "layers" architecture works. I wasn't satisfied with how complicated the "layer" interface was from my most recent proposal; having to subtype a layer "kind", have a field to store the next layer, have a specific constructor, overload `Layers.request`, etc. I also have started mulling over a new "middleware" interface on the server side and couldn't help but notice the similarities between client-side layers and server-side middleware. Indeed, they're actually very similar problem spaces in that you're trying to compartmentalize code into separate block (layers, middlewares), that all have a common interface and can be "chained" together and each have control flow when it's passed to them.

Given that context, I realized that moving to a more "functional" approach to middlewares and handlers could bring a nice consistency between the client-side layers vs. server-side middlewares. The concept is fairly simple:
  * the "bottom" of your stack must be a handler of the form `f(request) -> response)` (it's a bit more advanced than this, but for simplicity sake, we'll leave it at that for both client/server)
  * any _other_ components in the stack must follow the "middleware" interface, which is `f(handler) -> handler`. i.e. a "middleware" (client or server) takes a "handler" function, and can/must return a modified "handler" function. For examples sake, let's consider the following simplified setup:

```julia
# handler function that is the "bottom" of our stack
function handler(request)
    send(request)
    return readresponse()
end

# client-side middleware that adds custom logic to our base handler by adding a specific auth header when user provides credentials in keyword argument
function adds_auth_layer(handler)
    return function(req; auth_credentials=nothing, kw...)
        if auth_credentials !== nothing
            setheader(req.headers, "Custom-Authentication" => get_token(auth_credentials))
        end
        return handler(req; kw...)
    end
end

# now we componse our middleware with the handler by calling
const stack = adds_auth_layer(handler)
```

So what's going on here? We have our `handler` function, which does the actual work of sending the request and reading/returning the response. Then our `adds_auth_layer` middleware function _takes a handler_ as its only argument, and returns a _*function*_ that is also a handler, while "passing control" to the next layer by calling the `handler` argument. We then see how these compose together by "stacking" them by passing our `handler` to our `adds_auth_layer` to form a stack.

So why this approach?
  * It drastically simplifies the interface requirements for "layers" or middleware
  * It will unify the interfaces, whether you're modifying things on client side or server side (server-side code still needs to be implemented)
  * It provides roughly the same if not more compiler-friendly "stacking" by nesting these function calls rather than generating deeply nested layer parameterized types
  * It gets rid of the awkwardness of needing a separate `Layers.request` method vs. `HTTP.request`
  * It's forced a big cleanup of layer "arguments" so that things are much more simple and consistent; you can write middleware that either takes a single `req::Request` argument, or go one level deeper and take a single `stream::Stream` argument. This will also be consistent on the server-side and removes much of the awkwardness of the 3-6 arguments layers were previously receiving.

src/CookieRequest.jl
```diff
@@ -15,38 +13,47 @@ function __init__()
...
+            url = req.url
+            hostcookies = get!(cookiejar, url.host, Set{Cookie}())
+            cookiestosend = getcookies(hostcookies, url)
+            if !(cookies isa Bool)
+                for (name, value) in cookies
+                    push!(cookiestosend, Cookie(name, value))
+                end
+            end
+            if !isempty(cookiestosend)
+                existingcookie = header(req.headers, "Cookie")
```
quinnj, line 36: This is net-new code; I realized our test [here](https://github.com/JuliaWeb/HTTP.jl/blob/8743671f8e208b856defb17110ce2cd6705213c4/test/client.jl#L53) was wrong because by default `cookies=false` and so that request wasn't actually _deleting_ any cookies, but just didn't include them at all. This is apparent because it should only be deleting the single `hey` cookie, yet the test was testing that there were _no_ cookies.

The solution here is that we need to track when we automatically include cookies on a request, because in the case of a redirect, the initial response may return a set-cookie that expires or otherwise modifies a cookie we've already added to the request. When redirecting, we re-use the request headers that were set, but for cookies, we want to filter out any cookies that were automatically included on the 1st request and just let the "auto" logic do its thing by including them again _if they're applicable_. This lets cookies that have expired by removed and not sent in subsequent redirect requests.

It's unfortunate it's kind of lost in such a big refactor PR, but oh well.

src/DefaultHeadersRequest.jl
```diff
@@ -0,0 +1,62 @@
+module DefaultHeadersRequest
```
quinnj, line 1: This is a new "layer" that splits some of the logic out from MessageRequest.jl layer to a separate layer since the MessageRequest.jl layer is now purely for constructing the `Request` object at the start of the stack

src/HTTP.jl
```diff
@@ -323,14 +327,90 @@ end
...
+    return request(HTTP.stack(), method, url, headers, body, query; kw...)
 end
-function request(stack::Type{<:Layer}, method, url, h=Header[], b=nobody;
-                 headers=h, body=b, query=nothing, kw...)::Response
-    return request(stack, string(method), request_uri(url, query), mkheaders(headers), body; kw...)
+
+const STREAM_LAYERS = [timeoutlayer, exceptionlayer, debuglayer]
+const REQUEST_LAYERS = [messagelayer, redirectlayer, defaultheaderslayer, basicauthlayer, contenttypedetectionlayer, cookielayer, retrylayer, canonicalizelayer]
+
+pushlayer!(layer; request::Bool=true) = push!(request ? REQUEST_LAYERS : STREAM_LAYERS, layer)
```
quinnj, line 336: @christopher-dG, @c42f, here's the code for how I'm imagining we can do global stack modifications in BrokenRecord and elsewhere. Basically we have our 2 arrays of layer types above and we can document these `pushlayer!`/`poplayer!` functions as the way to add/remove layers globally.

src/HTTP.jl
```diff
@@ -323,14 +327,90 @@ end
...
+    return foldr((x, y) -> x(y), REQUEST_LAYERS; init=layers3)
+end
+
+function request(stack::Base.Callable, method, url, h=Header[], b=nobody, q=nothing;
+                 headers=h, body=b, query=q, kw...)::Response
+    return stack(string(method), request_uri(url, query), mkheaders(headers), body; kw...)
 end
 
-request(::Type{Union{}}, resp::Response) = resp
+macro client(requestlayers, streamlayers=[])
```
quinnj, line 360: New lexically-scoped client macro generator that provides the most common/convenient user-facing functions that can be defined in your own module.

src/IOExtras.jl
```diff
@@ -10,25 +10,37 @@ module IOExtras
 using ..Sockets
 using MbedTLS: MbedException
 
-export bytes, ByteView, nobytes, CodeUnits, IOError, isioerror,
+export bytes, isbytes, nbytes, ByteView, nobytes, IOError, isioerror,
```
quinnj, line 13: The changes here are a consolidation of the previous `bodybytes`/`bytes` and `bodylength` functions into a single place. There was some unfortunate mixups of when to use `bodybytes` vs `bytes` and we don't really need both.

src/MessageRequest.jl
```diff
@@ -1,102 +1,23 @@
...
-"request-target" per https://tools.ietf.org/html/rfc7230#section-5.3
-"""
-resource(uri::URI) = string( isempty(uri.path)     ? "/" :     uri.path,
-                            !isempty(uri.query)    ? "?" : "", uri.query,
-                            !isempty(uri.fragment) ? "#" : "", uri.fragment)
+export messagelayer
 
 """
-    request(MessageLayer, method, ::URI, headers, body) -> HTTP.Response
+    messagelayer(method, ::URI, headers, body) -> HTTP.Response
```
quinnj, line 12: We now require `messagelayer` to be the 1st/outermost layer in our stack as it translates the method, url, headers, and body into the `Request` object that all other layers consume.

src/Messages.jl
```diff
@@ -103,26 +105,25 @@ Represents a HTTP Response Message.
 You can get each data with [`HTTP.status`](@ref), [`HTTP.headers`](@ref), and [`HTTP.body`](@ref).
 
 """
-mutable struct Response <: Message
+mutable struct Response{T} <: Message
```
quinnj, line 108: The `Response` and `Request` objects are now both parameterized on their `body` fields. We previously had some slight awkwardness of having `body` be hard-coded as `Vector{UInt8}`, but then needing to pass a separate `body` argument to all layer functions in case the `body` was an `IO` or some other supported format. We now just parameterize `body` on whatever the user passes and can remove the `body_was_streamed` business and extra argument passing.

src/Messages.jl
```diff
@@ -140,19 +141,19 @@ HTTP.Response(200, headers; body = "Hello")
 Response() = Request().response
 
 Response(s::Int, body::AbstractVector{UInt8}) = Response(s; body=body)
-Response(s::Int, body::AbstractString) = Response(s, bytes(body))
+Response(s::Int, body::AbstractString) = Response(s; body=bytes(body))
```
quinnj, line 144: I'm pretty sure this and the method just below were just wrong by not passing `body` as a keyword argument

src/Messages.jl
```diff
@@ -197,29 +199,30 @@ Represents a HTTP Request Message.
...
     method::String
     target::String
     version::VersionNumber
     headers::Headers
-    body::Vector{UInt8}
+    body::T # Vector{UInt8} or some kind of IO
     response::Response
-    txcount::Int
-    parent
+    url::URI
```
quinnj, line 223: We add a new `url` field because there are many times throughout the client layers where we want access to the original url provided by the user and not just the `resource(url)` that we store as the `target` field.

src/Messages.jl
```diff
@@ -197,29 +199,30 @@ Represents a HTTP Request Message.
...
+    parent::Union{Response, Nothing}
+    context::Context
```
quinnj, line 225: We add a new `context` field that is a `Dict{Symbol, Any}` that allows some state sharing between layers. It's not used a ton, but it makes a more streamlined/formal way layers can share state if needed instead of relying on some of the other hacks we were doing before (passing state via keyword arguments, or having the `txcount` field on the `Request` object, requiring extra positional args to layers, etc.).

src/StreamRequest.jl
```diff
@@ -19,27 +21,18 @@ immediately so that the transmission can be aborted if the `Response` status
...
-export StreamLayer
-
-function request(::Type{StreamLayer{Next}}, io::IO, req::Request, body;
-                 reached_redirect_limit=false,
-                 response_stream=nothing,
-                 iofunction=nothing,
-                 verbose::Int=0,
-                 kw...)::Response where Next
-
+function streamlayer(stream::Stream; iofunction=nothing, verbose=0, redirect_limit::Int=3, kw...)::Response
```
quinnj, line 24: This is our only client-side "handler" function that is _not_ a middleware; i.e. it actually takes the `stream`, sends the request and reads the response. So all other client layers can be seen as "wrapping" this layer that does the actual send/receive operation.

c42f:
Cool I haven't yet dived into the details here. But the layers your suggesting remind me of Mux.jl's middleware. I think this can be powerful (IIUC express.js uses a similar design too?).

So this is interesting and I'll try to look a bit closer at it soon.

I do think closures can be a nice way to define these things, but I'd say that in practice I've found Mux pretty annoying to use! Some issues I've observed:
* Mux tries to do *everything* by stacking middleware.  There's some nice minimalism to this, but in practice it leads to problems.
* For example, Mux's routing wants you to add a middleware layer for each route... this leads to really awful stack traces!
* Having a big stack of closures as data rather than generic functions confuses Revise

I'm not saying these things apply to the design here which I think is a little different from Mux (?) But they're aspects to be aware of: do we want to optimize so that it's easiest to define new middleware vs reuse existing middleware?

quinnj:
Ok, I'd like to merge this soon, so I'm going to throw out a: merge-in-24-hours call to any potential reviewers. I think I've addressed everything from previous reviews, and provided a necessary "escape hatch" for the global stack manipulation if needed. I'm punting on the docs for now, because I'm already sketching out a bigger docs overhaul to do before the next release.

src/ContentTypeRequest.jl
```diff
@@ -1,30 +1,26 @@
...
 import ..sniff
 import ..Form
 using ..Messages
-import ..MessageRequest: bodylength, bodybytes
+import ..IOExtras
 import ..@debug, ..DEBUG_LEVEL
 
-abstract type ContentTypeDetectionLayer{Next <: Layer} <: Layer{Next} end
-export ContentTypeDetectionLayer
+export contenttypedetectionlayer
```
mattBrzezinski, line 10: I feel like these layer names are getting a bit lengthy to read without some way of breaking apart words such as,

> content_type_detection_layer

src/DefaultHeadersRequest.jl
```diff
@@ -0,0 +1,62 @@
...
+Set the default User-Agent string to be used in each HTTP request.
+Can be manually overridden by passing an explicit `User-Agent` header.
+Setting `nothing` will prevent the default `User-Agent` header from being passed.
+"""
+function setuseragent!(x::Union{String, Nothing})
+    USER_AGENT[] = x
+    return
+end
+
+end # module
```
mattBrzezinski, line 62:
```suggestion
end # module

```

src/DefaultHeadersRequest.jl
```diff
@@ -0,0 +1,62 @@
...
+        defaultheader!(headers, "Accept" => "*/*")
+        if USER_AGENT[] !== nothing
+            defaultheader!(headers, "User-Agent" => USER_AGENT[])
+        end
+
+        if !hasheader(headers, "Content-Length") &&
+        !hasheader(headers, "Transfer-Encoding") &&
+        !hasheader(headers, "Upgrade")
+            l = nbytes(req.body)
+            if l !== nothing
```
mattBrzezinski, line 34:
```suggestion
            if nbytes(req.body) !== nothing
```

src/DefaultHeadersRequest.jl
```diff
@@ -0,0 +1,62 @@
...
+        end
+        if !hasheader(headers, "Content-Type") && req.body isa Form && req.method in ("POST", "PUT")
+            # "Content-Type" => "multipart/form-data; boundary=..."
+            setheader(headers, content_type(req.body))
+        end
+        return handler(req; iofunction=iofunction, kw...)
+    end
+end
+
+const USER_AGENT = Ref{Union{String, Nothing}}("HTTP.jl/$VERSION")
```
mattBrzezinski, line 48: Const definitions under `using/import`?

src/Messages.jl
```diff
@@ -74,6 +75,7 @@ import ..bytes
 
 include("ascii.jl")
 
+const nobody = UInt8[]
```
mattBrzezinski, line 78: Uppercase to match style in other files?

test/chunking.jl
```diff
@@ -59,4 +61,7 @@ using BufferedStreams
 
         @test r == decoded_data
     end
+    close(server)
 end
+
+end # module
```
mattBrzezinski, line 67:
```suggestion
end # module

```

test/client.jl
```diff
@@ -489,3 +494,5 @@ end
         end
     end
 end
+
+end # module
```
mattBrzezinski, line 498:
```suggestion
end # module

```

test/cookies.jl
```diff
@@ -173,3 +178,5 @@
         @test istaskdone(tsk)
     end
 end
+
+end # module
```
mattBrzezinski, line 182:
```suggestion
end # module

```

test/loopback.jl
```diff
@@ -350,3 +349,5 @@ end
             "Response: HTTP/1.1 200 OK <= (POST /delay1 HTTP/1.1)"]
     end
 end
+
+end # module
```
mattBrzezinski, line 353:
```suggestion
end # module

```

mattBrzezinski (commented):
Just some formatting comments, nothing too in-depth.

src/BasicAuthRequest.jl
```diff
@@ -1,32 +1,27 @@
...
-function request(::Type{BasicAuthLayer{Next}},
-                 method::String, url::URI, headers, body; kw...) where Next
-
-    userinfo = unescapeuri(url.userinfo)
-
-    if !isempty(userinfo) && getkv(headers, "Authorization", "") == ""
-        @debug 1 "Adding Authorization: Basic header."
-        setkv(headers, "Authorization", "Basic $(base64encode(userinfo))")
+function basicauthlayer(handler)
+    return function(req; basicauth::Bool=true, kw...)
```
fredrikekre, line 15: What does the stack traces look like now when everything are anonymous functions?

nickrobinson251: also would it improve stacktraces if these were named functions?

e.g. 
```suggestion
    return function basicauthhandler(req; basicauth::Bool=true, kw...)
```
so they have names like `(::var"#basicauthhandler#5") (generic function with 1 method)`

quinnj: I think the backtraces are as legible as they used to be: the use of closures means we still see a nested list of the layers, and the use of a module per layer means its pretty clear where the layer is coming from.

image

src/Messages.jl
quinnj: this is how its been defined for ages; but yeah, we should probably go through and update it.

src/DefaultHeadersRequest.jl
quinnj: We use the `l` variable in the next line; but this is probably due to being a really poor variable name more than anything.

Issue #787, opened 2021-12-08
Windows failure on Julia 1.7


fredrikekre:
See e.g. #785

@quinnj are these the same failures you see on your PR?

quinnj:
Yes, it looks like the same failures; though in my PR, I pumped minimum Julia up to 1.6 and I think I was seeing the same kind of failures for 1.6 as well, so it might not be 1.7-specific. I wonder if something changed with TCPSocket on windows in the last release or two.

fredrikekre:
Maybe a libuv update?

fredrikekre:
@vtjnash maybe you have some insight? I reduced the failure to the following testset, which fails on Windows only:

```julia
using Test, HTTP, Sockets

@testset "Incomplete response with known content length" begin
    server = Sockets.listen(ip"0.0.0.0", 8080)
    try
        task = @async HTTP.listen("0.0.0.0", 8080; server=server) do http
            HTTP.setstatus(http, 200)
            HTTP.setheader(http, "Content-Length" => "64") # Promise 64 bytes...
            HTTP.startwrite(http)
            HTTP.write(http, rand(UInt8, 63)) # ...but only send 63 bytes.
            # Close the stream so that eof(stream) is true and the client isn't
            # waiting forever for the last byte.
            HTTP.close(http.stream)
        end

        err = try
            HTTP.get("http://localhost:8080"; retry=false)
        catch err
            @error "error" exception=(err, catch_backtrace())
            err
        end
        @test err isa HTTP.IOError
        @test err.e isa EOFError

    finally
        # Shutdown
        try; close(server); wait(task); catch; end
        HTTP.ConnectionPool.closeall()
    end
end
```

PR #802, opened 2022-02-19
Two changes to fix windows tests

quinnj:
The first change is in our `Connection` `unsafe_read` method where we
wrap our `unsafe_read` call to the underlying socket in a try-catch and
if an `IOError` is thrown, just rethrow as `EOFError`. The reasoning
here is left as a code comment, but in short, at the `Connection` level,
we don't really care whether the underlying socket throws an `EOFError`
or the socket received an RST or whatever, we really just need to know
if *something* went wrong and bubble that up to the header/body parsing
code in a consistent way.

The 2nd change is around some failing server tests; the problem with
sending a request with larger-than-allowed headers is the server
immediately bails on parsing the headers (to not DOS itself) and
immediately closes the underlying socket. The closing of the socket in
this abrupt manner means the client may or may not actually recieve the
413 response, which as the server we don't really care since they were
probably sending the request maliciously anyway. So I'm not exactly sure
the best way to encode that in the testing framework, but for now, I'm
kind of marking them as "allowed to fail".

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -159,8 +159,20 @@ function Base.unsafe_read(c::Connection, p::Ptr{UInt8}, n::UInt)
         c.timestamp = time()
     end
     if n > 0
-        unsafe_read(c.io, p, n)
-        c.timestamp = time()
+        # try-catch underlying errors here
+        # as the Connection object, we don't really care
+        # if the underlying socket was closed/terminated
+        # or just plain reached EOF, so we wrap any
+        # Base.IOErrors and just throw as EOFError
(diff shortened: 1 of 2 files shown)


quinnj:
The piesocket tests are being pretty flaky today; lots of 502 bad gateways and such. This is the best result I've seen so far. @fredrikekre, any thoughts? Want to review?

@vtjnash, just FYI for you after our conversation.

fredrikekre (approved):
Fixes #787 ?

quinnj:
> Fixes #787 ?

Yes!

PR #810, opened 2022-03-12
WebSocket: test HTTP.jl server against Deno client

fonsp:
Deno is a lightweight TS/JS runtime that is easy to use inside Julia, thanks to https://github.com/JuliaBinaryWrappers/Deno_jll.jl

It has a WebSocket client [based](https://github.com/denoland/deno/blob/a0285e2eb88f6254f6494b0ecd1878db3a3b2a58/ext/websocket/lib.rs) on https://github.com/snapview/tungstenite-rs, which is a good reference point to test against, because it is heavily used, under active development, and tested rigorously, against https://github.com/crossbario/autobahn-testsuite .

## Context
Pluto.jl is a heavy user of the HTTP.jl server. Because we run a localhost server for all Pluto users (on all OSses, network configurations, browsers, proxies, etc), we are more affected by stability issues than most projects. 

After discussion with @quinnj yesterday, I think that the best way that the Pluto devs can contribute to the HTTP.jl server is by writing test suites for the HTTP.jl server against verified HTTP/WS implementations, using our experience in setting up tests against Deno and Chrome Puppeteer.

This first test is just a "Hello world" WS test, but it sets up the architecture for more interesting tests in the future! Including stress tests (WS and HTTP), concurrent connections (multiple Deno processes), large WS payloads (more than one frame).


fonsp:
It looks like this test is already catching a difference between Windows and Unix in the websockets timing. 🌟 

How about I make the test simpler so that it passes on Windows, and then we can work on the timing difference in a new issue/PR? This PR was mostly about the new testing strategy.

fonsp:
Done now!

quinnj (commented):
This is great; thanks @fonsp!

fonsp:
For future reference: a failure on windows after simplifying the tests: https://github.com/JuliaWeb/HTTP.jl/runs/5692539164?check_suite_focus=true

PR #812, opened 2022-03-16
Remove use of at-eval in handlers register macro

quinnj:


fonsp:
Removing the deprecated features is breaking, right?

quinnj:
Technically yes, though this code has been deprecated for quite a few releases. We've had a couple breaking changes on master, so they'll all be rolled up in the next release.

Issue #782, opened 2021-11-30
`parse_multipart_form` does not parse "Content-Type" correctly.


rgankema:
Julia 1.6.4

Forms with fields with content types other than `"text/plain"` don't seem to get parsed correctly. Here's an MRE:

```julia
julia> form = HTTP.Form(["foo" => HTTP.Multipart("", IOBuffer("foo"), "application/octet-stream")])
julia> headers = [HTTP.content_type(form)]
julia> request = HTTP.Request("GET", "foo.nl", headers, read(form))
julia> HTTP.parse_multipart_form(request)
1-element Vector{HTTP.Multipart}:
 HTTP.Multipart(filename="", data=::Base.GenericIOBuffer{SubArray{UInt8, 1, Vector{UInt8}, Tuple{UnitRange{Int64}}, true}}, contenttype="text/plain", contenttransferencoding=""))
```

Note that the parsed `Multipart` has `contenttype="text/plain"` instead of `"application/octet-stream"`

NHDaly:
Okay, after investigating in #815, I've found that the failure to match comes from the `x` at the end of the regex, which from what i've read online, puts the regex into "comment mode." I'm still not exactly clear on what regex comment mode means.. this is what it says in the julia docs from `?r""`:
```julia
    •  `x` enables "comment mode": whitespace is enabled except when escaped with `\`, and `#`
       is treated as starting a comment.
```
Somehow that makes this regex fail to match:

```julia
infil> re_ct.re
r"(?i)Content-Type: (\S*[^;\s])"x

infil> header
"Content-Disposition: form-data; name=\"namevalue\"; filename=\"multipart.txt\"\r\nContent-Type: text/plain\r\n\r\n"

infil> occursin(r"(?i)Content-Type: (\S*[^;\s])"x, header)
false

infil> occursin(r"(?i)Content-Type: (\S*[^;\s])", header)
true
```

Do you know what that `x` is there for? What is it meant to do in this case? Can we safely get rid of it? Thanks!

quinnj:
PR up: https://github.com/JuliaWeb/HTTP.jl/pull/832. I imagine the original code probably just copied the regex usage from places like [here](https://github.com/JuliaWeb/HTTP.jl/blob/72aceefb6530bc83ce38ebae591ec64d204a50b6/src/Parsers.jl#L63) where we _do_ include comments in the regex string that we'd like to ignore. I think in general this isn't necessary though.

[...]

PR #815, opened 2022-03-24
Prefactoring: Add broken test case for parsing multipart content type (#782)

NHDaly:
This PR makes the parsemultipart.jl test file a bit easier to edit, and then adds a broken test case for a form section with a non-text content type.

This can be merged as-is, since it only adds a `@test_broken`. I'll try to follow this up soon with a fix, but I'm done working for the day, and figured this is a nice place to stop already. :)

Part of #782.

### DIFF
--- a/test/parsemultipart.jl
+++ b/test/parsemultipart.jl
@@ -21,6 +21,11 @@ function generate_test_body()
     Content-Type: text/plain\r
     \r
     \nfile with leading newline\n\r
+    ----------------------------918073721150061572809433\r
+    Content-Disposition: form-data; name="json_file1"; filename="my-json-file-1.json"\r
+    Content-Type: application/json\r
+    \r
+    {"data": ["this is json data"]}\r
     ----------------------------918073721150061572809433--\r
     """)
 end
@@ -93,8 +98,13 @@ end
         @test (startIndex + endIndexOffset) == endIndex
 
         (isTerminatingDelimiter, startIndex, endIndex) = find_multipart_boundary(body, delimiter, start = startIndex + 3)
-        @test isTerminatingDelimiter
+        @test !isTerminatingDelimiter
         @test 600 == startIndex
+        @test (startIndex + endIndexOffset) == endIndex
+
+        (isTerminatingDelimiter, startIndex, endIndex) = find_multipart_boundary(body, delimiter, start = startIndex + 3)
+        @test isTerminatingDelimiter
+        @test 804 == startIndex
         # +2 because of the two additional '--' characters
         @test (startIndex + endIndexOffset + 2) == endIndex
     end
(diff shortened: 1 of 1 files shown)


NHDaly:
And actually, looking more at the parsing code, i'm not sure how easy it'll be to fix, so maybe i'll leave the fix to someone else?

Hopefully this is still helpful as the first part of test-driven-development :)

NHDaly:
🤔 the regex here seems to work..?:
https://github.com/JuliaWeb/HTTP.jl/blob/9666a9a1ea1f8449c13166812fb420d0dc5edb36/src/parsemultipart.jl#L196

```julia
julia> match(r"(?i)Content-Type: (\S*[^;\s])", String(generate_test_body())[600:804])
RegexMatch("Content-Type: application/json", 1="application/json")
```
but when i added this print out:
```julia
    @show re_ct
    @info Parsers.exec(re_ct, header)
 ```
it claims not to have worked:
```julia
re_ct = HTTP.Parsers.RegexAndMatchData(r"(?i)Content-Type: (\S*[^;\s])"x, Ptr{Nothing} @0x00007fb8bc4302a0)
[ Info: false
```
dunno why.

test/parsemultipart.jl
```diff
@@ -3,7 +3,31 @@ using HTTP
 import HTTP.MultiPartParsing: find_multipart_boundary, find_multipart_boundaries, find_header_boundary, parse_multipart_chunk, parse_multipart_body, parse_multipart_form
 
 function generate_test_body()
-    Vector{UInt8}("----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"namevalue\"; filename=\"multipart.txt\"\r\nContent-Type: text/plain\r\n\r\nnot much to say\n\r\n----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"key1\"\r\n\r\n1\r\n----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"key2\"\r\n\r\nkey the second\r\n----------------------------918073721150061572809433\r\nContent-Disposition: form-data; name=\"namevalue2\"; filename=\"multipart-leading-newline.txt\"\r\nContent-Type: text/plain\r\n\r\n\nfile with leading newline\n\r\n----------------------------918073721150061572809433--\r\n")
+    Vector{UInt8}("""----------------------------918073721150061572809433\r
```
fredrikekre, line 6: Looks scary to rely on `\n`s here -- does string literals use `\n` regardless of the line endings of the file?
Perhaps write this as
```
Vector{UInt8}(join([
    "line1",
    "line2",
], "\r\n"))
```
or something?

NHDaly: Great point! Love that suggestion. Thanks 👍
Will change this when I'm back to the computer later.

NHDaly: Done.

NHDaly:
Thanks @fredrikekre. Applied your suggested changes. This is now ready for review!

fredrikekre (approved):

PR #819, opened 2022-04-02
Delete unused streaming_combinations.jl

lucifer1004:


test/streaming_combinations.jl
```diff
@@ -92,7 +92,7 @@ end
 send_small_chunk(http, 10)
 send_large_chunk(http, 1000)
 
-send_chunk(http, size)
+send_chunk(http, size) = begin
```
fredrikekre, line 95:
```suggestion
function send_chunk(http, size)
```

fredrikekre (commented):
Is this file not executed in the tests?

lucifer1004:
> Is this file not executed in the tests?

It is certainly not executed. And it seems to be an ancient draft and has not been updated for years. Can we just delete it?

quinnj:
Yeah, I would say we just delete it.

lucifer1004:
So it is deleted.

PR #820, opened 2022-04-08
Allow `HTTP.header` default to be anything

omus:
I wanted to be able to set the default to `nothing` to be able to distinguish from header which is set to the empty string.


quinnj (approved):
SGTM; note we currently have a stack of breaking changes on master; so if this is needed in a sooner release, you could propose a branch off the lastest 0.9.X release. I'm hoping to get the breaking release out in the next month or so.

omus:
> SGTM; note we currently have a stack of breaking changes on master; so if this is needed in a sooner release, you could propose a branch off the lastest 0.9.X release. I'm hoping to get the breaking release out in the next month or so.

I'm not in a rush. This has been a minor annoyance and it's good to just know it will be addressed in a future release.

PR #822, opened 2022-04-29
Fix method invalidations via ENV

### LINKED ISSUE #712: Method invalidations
HTTP.jl is a large cause of method invalidations in JuMP
```Julia
julia> using SnoopCompileCore

julia> invalidations = @snoopr begin
       using HTTP
       end;

julia> using SnoopCompile

julia> trees = invalidation_trees(invalidations)
4-element Vector{SnoopCompile.MethodInvalidations}:
 inserting joinpath(uri::URIs.URI, parts::String...) in URIs at /Users/oscar/.julia/packages/URIs/o9DQG/src/URIs.jl:509 invalidated:
   mt_backedges: 1: signature Tuple{typeof(joinpath), Any, String} triggered MethodInstance for jointail(::Any, ::String) (0 children)
   1 mt_cache

 inserting write(iod::HTTP.DebugRequest.IODebug, a...) in HTTP.DebugRequest at /Users/oscar/.julia/dev/HTTP/src/IODebug.jl:30 invalidated:
   backedges: 1: superseding write(io::IO, x1, xs...) in Base at io.jl:636 with MethodInstance for write(::IO, ::Int64, ::Int64, ::Int64, ::Int64) (1 children)

 inserting readuntil(iod::HTTP.DebugRequest.IODebug, f) in HTTP.DebugRequest at /Users/oscar/.julia/dev/HTTP/src/IODebug.jl:58 invalidated:
   backedges: 1: superseding readuntil(io::IO, target::AbstractString; keep) in Base at io.jl:884 with MethodInstance for readuntil(::IO, ::String) (15 children)
              2: superseding readuntil(s::IO, delim::AbstractChar; keep) in Base at io.jl:771 with MethodInstance for readuntil(::IO, ::Char) (216 children)
   18 mt_cache

 inserting write(iod::HTTP.DebugRequest.IODebug, x::String) in HTTP.DebugRequest at /Users/oscar/.julia/dev/HTTP/src/IODebug.jl:38 invalidated:
   backedges: 1: superseding write(io::IO, s::Union{SubString{String}, String}) in Base at strings/io.jl:185 with MethodInstance for write(::IO, ::String) (261 children)


julia> length(invalidations)
1043
```

These seem fixable. I'll take a look in the next few days unless someone beats me to it.

x-ref: https://github.com/jump-dev/JuMP.jl/issues/2273

rikhuijzer:
This is a minimal possible solution to fix #712. It only adds the overrides for Base methods if

```julia
debug_mode()::Bool = get(ENV, "HTTP_DEBUG", "") == "true"
```

We could also let this check return `true` when `ENV["CI"] == "true"` to make this PR less breaking.

**With `HTTP_DEBUG`**
```julia
julia> ENV["HTTP_DEBUG"] = "true"

julia> using HTTP

julia> HTTP.get("https://example.com"; verbose=3);
DEBUG: 2022-04-29T11:40:56.801 759e ➡️   "716 ...
[...]
```

**Without `HTTP_DEBUG`:**
```julia
julia> using HTTP

julia> HTTP.get("https://example.com"; verbose=3);
ERROR: To enable debugging, set the environment variable "HTTP_DEBUG" to "true" and restart HTTP.jl
[...]
```

NB. Note that the method invalidations and setting a `const` can occur during precompilation too, so this PR is a bit tricky with state.


src/IODebug.jl
```diff
@@ -25,87 +25,92 @@ else
...
-Base.show(io::IO, iod::IODebug) = show(io, iod.io)
-
-function show_log(io::IO, iod::IODebug)
-    lock(io)
-    println(io, "$(typeof(iod)):\nio:     $(iod.io)")
-    prevop = ""
-    for (operation, f, bytes) in iod.log
-        if prevop != "" && prevop != operation
-            println(io)
+debug_mode()::Bool = get(ENV, "HTTP_DEBUG", "") == "true"
```
quinnj, line 28: Why introduce yet another "debug lever" in an ENV variable vs. just defining these if `DEBUG_LEVEL[] > 0`? I know it's a bit of a backward step since we changed `DEBUG_LEVEL` from a const to a Ref so it could be dynamic, but whatevs, it's just a debug mechanism. I'd rather use an existing lever vs. introduce a new one

rikhuijzer: I've just now tried that. It does work but I'm unable to allow downstream packages to enable debugging. They have to override the constant before the module is loaded.

rikhuijzer: The common trick to set the enviroment variable in the `__init__()` of the downstream package doesn't work either.

rikhuijzer: A more future-proof solution could be to define intermediate methods such as `httpwrite` where
```julia
http_write(io::IODebug) = ...
http_write(io::IO) = Base.write(...)
```
This way, `IODebug` doesn't need to be a subtype of `IO` and therefore the method invalidations in `Base` will also be avoided.

fredrikekre:
The package won't be re-precompiled for change in environment variables, right? So I don't think this works.

rikhuijzer:
> The package won't be re-precompiled for change in environment variables, right? So I don't think this works.

It does work. But well not very conveniently no.

As another possible solution, how about a `HTTPDebug.jl` package? That should work in combination with `Requires.jl`.

fonsp:
What if we move this into `__init__`, inside `@eval` if needed?

rikhuijzer:
> What if we move this into `__init__`, inside `@eval` if needed?

How do you mean? Feel free to change things in the PR

rikhuijzer:
> ```julia
> function __init__()
>     # Only enabling these overrides when `ENV["HTTP_DEBUG"] == "true"` to avoid many method invalidations.
>     @eval if debug_mode()
>         Base.wait_close(iod::IODebug) = Base.wait_close(iod.io)
>     
>         Base.write(iod::IODebug, a...) =
>             (logwrite(iod, :write, join(a));
>              write(iod.io, a...))
>     
>         Base.write(iod::IODebug, a::Array) =
>             (logwrite(iod, :write, join(a));
>              write(iod.io, a))
>     
>         Base.write(iod::IODebug, x::String) =
>             (logwrite(iod, :write, x);
>              write(iod.io, x))
>          # etc...
>     end
> end
> ```

I've tried unsuccessfully.

quinnj:
Let me take this on. I have ideas around removing the functionality all together and replacing it with a more cohesive debugging system. Give me a week or so to test out a few ideas. I just don't think any of the solutions here are going to be ideal long-term and we've already stacked up a bunch of breaking changes on `master`, so we might as well "overhaul" things while we're at it.

quinnj:
Superseded by https://github.com/JuliaWeb/HTTP.jl/pull/826

Issue #796, opened 2022-01-17
Inconsistent behaviour among Julia 1.6.5 (works ok) versus Julia 1.7.0 (error)


amartinhuertas:
Dear HTTP.jl developers, 

when I execute the following MWE on a Julia REPL:

```julia
using HTTP
doi="10.1016/j.cma.2021.114093"
a = String(HTTP.get("http://doi.org/$doi", ["Accept" => "application/x-bibtex"]).body)
```

I get the expected output with Julia 1.6.5, but an exception with Julia 1.7.0 (see details below).

Please note that the doi at hand (i.e., 10.1016/j.cma.2021.114093)  is tricky in the sense that one of the authors of the paper has an special character in it, namely **í** in Martín. If I use a doi associated to an article in which the author surnames do not have such character, both Julia 1.6.5 and 1.7.0 have the correct behaviour. I suspect this might be related to differences in the regex implementation of Julia 1.6.5 versus 1.7.0, but I am not certain.

Do you have any clue with regards to whether this is a bug or not, and a possible fix? 
Thanks in advance! 

cc @santiagobadia

Versions:
 - Julia v1.6.5 vs v1.7.0
 - HTTP.jl v0.9.17
 - MbedTLS.jl v1.0.3

```
a = String(HTTP.get("http://doi.org/$doi", ["Accept" => "application/x-bibtex"]).body)
[...]

PR #830, opened 2022-05-25
Attempt to reencode malformed headers from Latin-1 to UTF8

quinnj:
As brought up in #796, there may be scenarios where headers may contain
non-UTF8 characters (even though they're supposed to be ASCII). Appreciation
to @StefanKarpinski for the Latin-1 -> UTF-8 conversion code and the suggestion
to try reencoding before throwing an error. As proposed in this PR, the normal
header parsing path should be unaffected and only when we're unable to parse
a normal header will we attempt this reencoding.

Note that curl warns on the malformed header and filters it out.

### DIFF
--- a/src/Parsers.jl
+++ b/src/Parsers.jl
@@ -235,6 +235,28 @@ function parse_header_field(bytes::SubString{String})::Tuple{Header,SubString{St
         return (group(1, re, bytes) => unfold), nextbytes(re, bytes)
     end
 
+    # https://github.com/JuliaWeb/HTTP.jl/issues/796
+    # there may be legacy values encoded in latin-1 that will fail
+    # the utf8-expecting regexes above, before throwing the error here,
+    # we can do the ascii check and potentially re-encode the headers and
+    # try again
+    rawbytes= codeunits(bytes)
+    n = count(≥(0x80), rawbytes)
+    if n > 0
+        buf = Base.StringVector(length(rawbytes) + n)
+        i = 0
+        for byte in rawbytes
+            if byte ≥ 0x80
+                buf[i += 1] = 0xc0 | (byte >> 6)
+                buf[i += 1] = 0x80 | (byte & 0x3f)
+            else
+                buf[i += 1] = byte
+            end
+        end
+        @warn "malformed HTTP header; attempting to re-encode from Latin-1 to UTF8"
+        return parse_header_field(SubString(String(buf)))
+    end
+
     throw(ParseError(:INVALID_HEADER_FIELD, bytes))
 end
 
--- a/src/Parsers.jl
+++ b/src/Parsers.jl
@@ -221,6 +221,7 @@ function parse_header_field(bytes::SubString{String})::Tuple{Header,SubString{St
     # "malformed headers" and ignores them. we attempt to re-encode
     # these from latin-1 => utf-8 and then try to parse.
     if !isvalid(bytes)
+        @warn "malformed HTTP header detected; attempting to re-encode from Latin-1 to UTF8"
         rawbytes = codeunits(bytes)
         buf = Base.StringVector(length(rawbytes) + count(≥(0x80), rawbytes))
         i = 0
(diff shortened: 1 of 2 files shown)


src/Parsers.jl
```diff
@@ -235,6 +235,28 @@ function parse_header_field(bytes::SubString{String})::Tuple{Header,SubString{St
...
     end
 
+    # https://github.com/JuliaWeb/HTTP.jl/issues/796
+    # there may be legacy values encoded in latin-1 that will fail
+    # the utf8-expecting regexes above, before throwing the error here,
+    # we can do the ascii check and potentially re-encode the headers and
+    # try again
+    rawbytes= codeunits(bytes)
+    n = count(≥(0x80), rawbytes)
+    if n > 0
```
StefanKarpinski, line 245:
```suggestion
    if !isvalid(bytes)
        rawbytes = codeunits(bytes)
        n = count(≥(0x80), rawbytes)
```
I think this would fix the recursion problem. Since `!isvalid` implies that `n > 0` we don't need that check anymore.

quinnj:
@StefanKarpinski, thanks for the suggestions. I incorporated them locally since I forgot that we need to do the `isvalid` check _before_ we call the regexes, since that was the original issue we were running into (i.e. calling the regex on latin1 data). The strategy here then is:
  * Before parsing the first header, check if all headers are `isvalid`
  * If not, assume latin1 and re-encode
  * If still not `isvalid`, error out (as we do now on latin1); otherwise, parsing continues successfully

We could maybe try to match some client behavior here and just "skip" the header w/ latin1, parsing other headers, but it seems tricky to try and get exactly right because we're dealing w/ potentially corrupt data, but still trying to find the `\r\n` to move on to the next header.

quinnj:
Also, FYI, I found [this](https://www.jmix.io/cuba-blog/utf-8-in-http-headers/) as a decent discussion

src/Parsers.jl
```diff
@@ -215,6 +215,28 @@ Return `Pair(field-name => field-value)` and
...
+        for byte in rawbytes
+            if byte ≥ 0x80
+                buf[i += 1] = 0xc0 | (byte >> 6)
+                buf[i += 1] = 0x80 | (byte & 0x3f)
+            else
+                buf[i += 1] = byte
+            end
+        end
+        bytes = SubString(String(buf))
+        !isvalid(bytes) && @goto error
```
StefanKarpinski, line 237: There's no need to check this—the resulting string is guaranteed to be valid by construction.

quinnj: Good catch.

src/Parsers.jl
```diff
@@ -215,6 +215,28 @@ Return `Pair(field-name => field-value)` and
...
 function parse_header_field(bytes::SubString{String})::Tuple{Header,SubString{String}}
+    # https://github.com/JuliaWeb/HTTP.jl/issues/796
+    # there may be certain scenarios where non-ascii characters are
+    # included (illegally) in the headers; curl warns on these
+    # "malformed headers" and ignores them. we attempt to re-encode
+    # these from latin-1 => utf-8 and then try to parse.
+    if !isvalid(bytes)
+        @warn "malformed HTTP header detected; attempting to re-encode from Latin-1 to UTF8"
+        rawbytes = codeunits(bytes)
+        buf = Base.StringVector(length(rawbytes) + count(≥(0x80), rawbytes))
```
fonsp, line 226: Should we add a fallback for when `Base.StringVector` does not exist?

fonsp, line 226: Should we add a fallback for when `Base.StringVector` does not exist?

quinnj: We're requiring Julia 1.6 in HTTP.jl now, so I think we should be covered?

fonsp: It's just that this is not public Julia API (it has no docstring). But I guess that if Stefan recommends it then it should be fine?

Issue #573, opened 2020-08-28
HTTP.download() not honoring `local_path`


staticfloat:
Julia 1.5.1
HTTP.jl v0.8.17

Example:

```
julia> HTTP.download("https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/7858451b7a520344eb60354f69809d30a44e7dae", "/tmp/7858451b7a520344eb60354f69809d30a44e7dae")
┌ Info: Downloading
│   source = "https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/7858451b7a520344eb60354f69809d30a44e7dae"
│   dest = "/tmp/7858451b7a520344eb60354f69809d30a44e7dae.gz"
│   progress = 1.0
│   time_taken = "0.16 s"
│   time_remaining = "0.0 s"
│   average_speed = "12.986 MiB/s"
│   downloaded = "2.013 MiB"
│   remaining = "0 bytes"
└   total = "2.013 MiB"
"/tmp/7858451b7a520344eb60354f69809d30a44e7dae.gz"
```

Notice how it gets `.gz` appended to the end.

quinnj:
Assigning to @oxinabox to take a look if able; otherwise, I can dig in if needed.

PR #835, opened 2022-05-27
don't add gz if we are given a local filepath

oxinabox:
Closes https://github.com/JuliaWeb/HTTP.jl/issues/573
I am not sure if this is actually the correct thing to do though.
The content-encoding if gzipped, but i am not sure that it is correct that the file that is downloaded should just ende up gzipped.

I think the real solution is we should probably be decompressing it inside of HTTP.jl?
Not sure
If I open https://httpbin.org/gzip with my browser it is displayed in nongzipped form
OTOH if i open  https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/7858451b7a520344eb60354f69809d30a44e7dae
then a file id downloaded that is gzipped

`curl` doesn't seem to uncompress either (at least by default)

### DIFF
--- a/src/download.jl
+++ b/src/download.jl
@@ -58,8 +58,7 @@ determine_file(::Nothing, resp, hdrs) = determine_file(tempdir(), resp, hdrs)
 # ^ We want to the filename if possible because extension is useful for FileIO.jl
 
 function determine_file(path, resp, hdrs)
-    # get the name
-    name = if isdir(path)
+    if isdir(path)
         # we have been given a path to a directory
         # got to to workout what file to put there
         filename = something(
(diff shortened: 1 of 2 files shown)


quinnj (approved):
This LGTM; there's another open issue about auto-decompressing gzipped encodings. I think I'd be ok w/ taking a dependency on CodecZlib.jl to do the decompression. The only thing I worry about is production settings where you may want really fine-grained control over all that and making sure it's configurable enough to control.

quinnj:
Merging for now; happy to take enhancement PRs on top of this. As we have at least a little time before a new release though, I'm not worried about getting things into master sooner. The plan is to try and get a 1.0 release in the next few weeks.

fredrikekre:
Should this be done at all? `curl` doesn't append it, for example.

oxinabox:
Yeah, idk. It's unclear to me if it should be done at all.

Issue #256, opened 2018-06-19
Support transparent gzip decompression


Alexander-Barth:
Some server always respond with a compressed stream, e.g. the stackoverflow API:

```
import HTTP
r = HTTP.get("https://api.stackexchange.com/2.2/questions"; query = Dict("site" => "stackoverflow"));
``` 

It would be nice if I can directly access the decompressed response body. 
As a workaround, I can use CodecZlib

```
using CodecZlib
String(transcode(GzipDecompressor,r.body))
```
Would it be possible to support transparent decompression, or is it already implemented?

Julia 0.6.3
HTTP.jl master 61e6b05 
MbedTLS.jl 0.5.9

samoconnor:
Yes, this could be supported.
This is the place to hook it in:
https://github.com/JuliaWeb/HTTP.jl/blob/b602c4ea0cf346b35ae9bea13db2c960eee759de/src/Messages.jl#L395-L401

EricForgy:
This is cool. You guys ok to add `CodecZlib` as a dependency? Making the change is easy. Any idea what would be a good test?

Alexander-Barth:
A test case could be the use of `http://httpbin.org/gzip`: 

```
julia> r = Requests.get("http://httpbin.org/gzip")
Response(200 OK, 13 headers, 221 bytes in body)

julia> print(readstring(r))
{"gzipped":true,"headers":{"Accept":"text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8","Connection":"close","Host":"httpbin.org","User-Agent":"Requests.jl/0.0.0"},"method":"GET","origin":"139.165.106.124"}

julia> r.headers
Dict{String,String} with 13 entries:
  "Connection"                       => "keep-alive"
  "Via"                              => "1.1 vegur"
  "Access-Control-Allow-Credentials" => "true"
  "Date"                             => "Wed, 20 Jun 2018 08:13:50 GMT"
  "Access-Control-Allow-Origin"      => "*"
  "http_minor"                       => "1"
  "Keep-Alive"                       => "1"
  "status_code"                      => "200"
  "Server"                           => "gunicorn/19.8.1"
  "Content-Length"                   => "191"
  "http_major"                       => "1"
  "Content-Type"                     => "application/json"
  "Content-Encoding"                 => "gzip"
```

jakewilliami:
[...]

PR #838, opened 2022-05-28
Auto decompress gzip-encoding response bodies

quinnj:
Implements #256. If the content-encoding of a response is "gzip"
and the keyword argument `decompress === true`, then we'll
use CodecZlib.jl to decompress the response and set as the response
body. Passing `decompress=false` will leave the resposne body as-is.
We also support `HTTP.decode(::Request, "gzip")` which will do
the decompression.

### DIFF
--- a/src/StreamRequest.jl
+++ b/src/StreamRequest.jl
@@ -7,7 +7,7 @@ import ..ConnectionPool
 using ..MessageRequest
 import ..RedirectRequest: nredirects
 import ..sprintcompact
-using LoggingExtras
+using LoggingExtras, CodecZlib
 
 export streamlayer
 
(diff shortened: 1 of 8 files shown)


src/StreamRequest.jl
```diff
@@ -103,9 +103,13 @@ end
 writechunk(stream, body::IO) = writebodystream(stream, body)
 writechunk(stream, body) = write(stream, body)
 
-function readbody(stream::Stream, res::Response, redirectlimitreached)
+function readbody(stream::Stream, res::Response, redirectlimitreached, decompress)
     if isbytes(res.body)
-        res.body = read(stream)
+        if decompress && header(res, "Content-Encoding") == "gzip"
+            res.body = transcode(GzipDecompressor, read(stream))
```
fredrikekre, line 109: Is there a streaming version of this that does this in one go?

quinnj: Good point; there is. And that actually simplifies it because we can just wrap the `stream` in the decompressor and get the decompressing for both static byte vector and streaming response case.

fonsp:
This means that we could add `"Accept-Encoding" => "gzip"` to our default headers, if `decompress=true`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

quinnj:
Great idea! Mind making a PR @fonsp?

PR #841, opened 2022-06-08
proxy: islocalhost: ipv6 support

fonsp:
Follow-up of https://github.com/JuliaWeb/HTTP.jl/pull/833, I am guessing that this was the intended value for the third check.

I added two additional ways to format the same IPv6 address, is that enough?


quinnj (approved):

quinnj:
Yeah, good catch. Thanks @fonsp!

PR #843, opened 2022-06-10
Overhaul websockets code

quinnj:
Big update to websockets code. The work here includes:
  * Reading through [RFC 6455](https://www.rfc-editor.org/rfc/rfc6455.txt) and trying to make our implementation more compliant
  * Getting all tests passing from the https://github.com/crossbario/autobahn-testsuite, which is pretty much the industry standard for testing websocket spec compliance; running the tests currently requires docker, which is only available on ubuntu for github actions, so this PR is my attempt to at least run those tests on linux; they currently pass on my macOS machine running the docker images locally
  * An updated overall API for our websockets code, in particular:
    * `WebSocket` doesn't subtype IO anymore; it's _kind_ of an IO, but also not really in a lot of ways, and it actually makes it a little awkward because of all the ways it _isn't_ an IO, while not really getting any advantage from the subtyping
    * The API for using a `WebSocket` includes `send(ws, msg)` for sending messages, `receive(ws)` for receiving non-control messages. pings/pongs are handled automatically. You can also iterate the `WebSocket` object, where each iteration will yield the body of a non-control message, with iteration finishing when the websocket connection is closed. You can send TEXT messages by calling `send(ws, msg::AbstractString)`, BINARY messages by calling `send(ws, msg::AbstractVector{UInt8})`, and fragmented messages of either type by calling `send(ws, iterable)` where the iterable iterates elements of type `AbstractString` or `AbstractVector{UInt8}` (must all be one or the other, no intermixing). To close the websocket, you call `close(ws)`. To manually send ping/pong, you call `ping(ws)`, `pong(ws)`.

In terms of making/serving websocket connections, those APIs remain the same via `WebSockets.open`/`WebSockets.listen`.


quinnj:
Sigh......it's proving difficult to get the CI to run the autobahn testsuite here.

quinnj:
Ok, this is looking better; we're able to run the autobahn test suite for 64-bit linux & macOS. 32-bit python seems not possible. and windows doesn't seem possible because it requires python 2.7 and the windows "compiler tools" isn't supported for 2.7 any more.

Going to add docs here and then I think it's ready to merge.

rikhuijzer:
> Getting all tests passing from the https://github.com/crossbario/autobahn-testsuite, which is pretty much the industry standard for testing websocket spec compliance

Did this PR already find bugs? 🚀

quinnj:
Oh yes, there were PLENTY of bugs in the old implementation.

quinnj:
Alright, lots of inline docs added. I think this is ready to go. Again, merging semi-quickly because I have other work coming in, but I would love anyone to look/review things and we can address things in follow up PRs. I plan to review the ~30 open websockets-related issues, which I believe can mostly be closed from the updates here (but we still need to review each to make sure!).

BoZenKhaa:
I love this ♥. Looking at the inline docs and tests, the blocking `eof!` is no longer necessary? If so, the readme example is up for a facelift as well: https://github.com/JuliaWeb/HTTP.jl#websocket-examples

Can't wait for this to be released!

PR #845, opened 2022-06-11
Use `uint16`

rikhuijzer:
`FrameFlags` is used only inside the `WebSockets` module while extending `Base.UIn16` is global. This PR may avoid some unnecessary method invalidations.


fredrikekre (approved):

quinnj:
Thanks @rikhuijzer!

PR #857, opened 2022-06-19
Fix HTTP.listen with provided TCP server

fredrikekre:
Sockets.getsockname returns host as an IPAddr and port as Int, but
HTTP.Listener expects host and port as Strings.


quinnj (approved):
Thanks!

PR #866, opened 2022-06-22
Store the original registered route path when Router matches

quinnj:
This is a common feature of other router implementations, sometimes referred
to as "reverse routing". We don't go the full way in allowing you to do
`getregisteredroute(r::Router, url)`, but we could do that eventually. For now,
it's at least very useful to store the registered route that was matched
during request handling.

### DIFF
--- a/src/Handlers.jl
+++ b/src/Handlers.jl
@@ -148,7 +148,7 @@ end
 const VARREGEX = r"^{([[:alnum:]]+):?(.*)}$"
 
 function Variable(pattern)
-    re = match(VARREGEX, pattern)
+    re = Base.match(VARREGEX, pattern)
     if re === nothing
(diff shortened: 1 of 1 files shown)


src/Handlers.jl
```diff
@@ -254,7 +254,7 @@ function Base.insert!(node::Node, leaf, segments, i)
     end
 end
 
-function Base.match(node::Node, params, method, segments, i)
+function Base.match(node::Node, method, segments, i)
```
fredrikekre, line 257: Drive by comment, but if this is just an internal function, maybe it shouldn't overload Base.match?

quinnj: Yeah, that's a good thought; same with `insert!`.

PR #867, opened 2022-06-22
FIX UndefRefError to x in cookies.jl

raffopazzo:


raffopazzo:
Not tested and may not be the correct thing to do. Just wanted to point out the issue

fredrikekre:
LGTM but of course would be good with a test

raffopazzo:
I don't know exactly what's wrong. I just saw this happen once and I thought I'd report it to you but trying to write it up in an issue was more complicated then showing you where the issue is :D

quinnj (approved):
Yeah, good catch. This was ported directly over from the go code, so I must have just missed this in the translation. I also ported their tests over, so I guess they didn't have a test covering this. This is kind of a random feature though, because it's really just saying, hey, there was something in the `Cookie` that doesn't have a semantic meaning, so stick in this `unparsed` field. So probably not super common.

Issue #853, opened 2022-06-18
WebSocket permessage-deflate


Mixfair:
- Julia 1.7.2
 - HTTP.jl 0.9.17
 
 Hi, 
 
It would be great if HTTP.Websockets could support the permessage-deflate extension, which provides compression for websocket frames. It's described in [RFC 7692](https://datatracker.ietf.org/doc/html/rfc7692) and is currently implemented client-side in browsers. 

 For example, endpoint: 
 ```
using HTTP

url = "wss://stream.binance.com:9443/stream?streams=adausdt@depth20@100ms/btcusdt@depth20@100ms"
headers = [
    "sec-websocket-extensions" => "permessage-deflate",
]

HTTP.WebSockets.open(url, headers = headers) do ws
    while true
        data = readavailable(ws)
        println(String(data))
    end
end 
 ```

I have already implemented this feature locally with `CodecZlib`. If possible it would be great to see this feature by default

quinnj:
Sigh.....I thought I'd at least get to do the 1.0 release w/ the overhauled WebSockets code that "actually works" before getting a request for this, but that's how it goes sometimes!

Yeah, I don't think this should be too hard to support with the redesigned WebSockets code on HTTP#master. I've been planning to do this as a post-1.0 task however since I it shouldn't involve any breaking changes, so it's easy to add as an enhancement for, e.g., 1.1 release.

@Mixfair, if you'd like to have a go at implementing this, I'd be happy to provide guidance!

PR #868, opened 2022-06-27
permessage-deflate support

Mixfair:
Hi.
Sharing my example of implementing permessage-deflate using CodecZlib in continuation #853.
In the process, I encountered fragmentation difficulties, since the `opcode` is calculated during the process of sending data and we do not know about the type of compressed data. Therefore, I decided to separately compress a string into a string, an array into an array before sending. Also moved utf-8 check after unpacking. Tested with Chrome and Julia client.

Please, review it

Server: 
```
WebSockets.listen(host, port; binary=false, isdeflate=true) do ws
    @async while true
        sleep(2)
        payload = WebSockets.compress("Hello")                       # if fragmentation true
        WebSockets.send(ws, [payload[1:2], payload[3:end]])
        # payload = "Hello"                                          # if fragmentation false
        # WebSockets.send(ws, payload)
    end
    for msg in ws
        @info "Received: ", msg
    end
end
```
Client:
```
WebSockets.open("ws://127.0.0.1:8080") do ws
    for msg in ws
        @info "Client received: ", msg
        # msg = WebSockets.compress(msg)                            # if fragmentation true
        WebSockets.send(ws, msg)
    end
end
```

### DIFF
--- a/src/WebSockets.jl
+++ b/src/WebSockets.jl
@@ -90,27 +90,18 @@ function mask!(bytes::Vector{UInt8}, mask)
     end
     return
 end
-
-function compress(data::T) where T <: AbstractVector{UInt8}
-    compressed = transcode(DeflateCompressor, data)
-    push!(compressed, 0x00)
-    return compressed
+function final_deflate_codecs(t::Tuple)
+    CodecZlib.TranscodingStreams.finalize(t[1])
+    CodecZlib.TranscodingStreams.finalize(t[2])
 end
 
-function compress(data::String)
-    compressed = transcode(DeflateCompressor, data)
-    push!(compressed, 0x00)
-    return String(compressed)
-end
+function init_deflate_codecs()
+    codecco = DeflateCompressor()
+    CodecZlib.TranscodingStreams.initialize(codecco)
+    codecde = DeflateDecompressor()
+    CodecZlib.TranscodingStreams.initialize(codecde)
 
-function decompress(data::T) where T <: AbstractVector{UInt8}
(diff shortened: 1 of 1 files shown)


src/WebSockets.jl
```diff
@@ -91,6 +91,27 @@ function mask!(bytes::Vector{UInt8}, mask)
     return
 end
 
+function compress(data::T) where T <: AbstractVector{UInt8}
+    compressed = transcode(DeflateCompressor, data)
+    return vcat(compressed, 0x00)
```
fredrikekre, line 96: Looks like you don't need to allocate a full new vector here:
```suggestion
    compressed = transcode(DeflateCompressor, data)
    push!(compressed, 0x00)
    return compressed
```

src/WebSockets.jl
```diff
@@ -91,6 +91,27 @@ function mask!(bytes::Vector{UInt8}, mask)
...
+end
+
+function compress(data::String)
+    compressed = transcode(DeflateCompressor, data)
+    return String(vcat(compressed, 0x00))
+end
+
+function decompress(data::T) where T <: AbstractVector{UInt8}
+    decompressed = transcode(DeflateDecompressor, vcat(data, [0x00, 0x00, 0xff, 0xff, 0x03, 0x00]))
```
fredrikekre, line 105: Would be nice not to allocate a new vector here, but not sure if possible.

quinnj:
Implemented in https://github.com/JuliaWeb/HTTP.jl/pull/1308; sorry for the slow response/review here. As the websocket code has been completely re-done for the 2.0 release, I ended up doing the support along with the rewrite.

PR #870, opened 2022-07-03
CORS example improvement: added CORS response header to 200 server responses

BoZenKhaa:
The CORS example currently only handles the preflight CORS requests, but its 200 responses do not get through CORS on the client-side (on different origin, something not apparent from the current code). According to https://cors-errors.info/faq#bae3, the responses need to set their `Access-Control-Allow-Origin` header as well. This PR adds a response header to the 200 responses to make the example complete and viable in a real CORS situation.

However, the fix does not set the CORS header for other than 200 responses, so for example 404 still does not get through. 

I found out about this when trying to implement an HTTP.jl server for a javascript visualization running in a browser on a different origin (localhost, but different port).

### DIFF
--- a/docs/examples/cors_server.jl
+++ b/docs/examples/cors_server.jl
@@ -1,8 +1,9 @@
 """
 Server example that takes after the simple server, however,
 handles dealing with CORS preflight headers when dealing with more
-than just a simple request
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Thanks @BoZenKhaa!

PR #872, opened 2022-07-04
Allow any string for parameter names in router, fixes #871.

### LINKED ISSUE #871: Routing: path containing underscore "_" in path path variable fails to route
Using a path variable with an underscore registers a path successfully, but then fails to match the path to a request. For example, consider the [simple server example](https://github.com/JuliaWeb/HTTP.jl/blob/master/docs/examples/simple_server.jl):

Changing the path variable from `id` to `animal_id`:
image

leads to the following error:
image

I can try to prepare a PR, but I am not sure what should actually happen. In my opinion, either

1. underscores and other allowed Julia variable names should work,
2. there should be a mention of allowed characters in the docstring,
3. `register!` should throw an error when trying to register a path with not allowed characters path variables,
4. trying to fetch a route with a path using the incorrect characters in path variables should show some warning

Versions:
 - Julia 1.7.2
 - HTTP.jl 1.0.5
 - MbedTLS.jl 1.1.0

fredrikekre:


quinnj (approved):
Thanks @fredrikekre!

PR #877, opened 2022-07-12
Allow user-configurable retry delay and check

quinnj:
It's been reported that users sometimes feel the need to disable HTTP's
retry abilities and wrap HTTP methods with their own retry logic due
to the lack of configurability around delays and overriding the retry
check. Since there is some tricky logic in getting retries just right,
especially in the presence of streaming request or response bodies,
this PR proposes more configurability for retry logic in the hopes
that users won't opt to roll their own incorrect retry logic.

### DIFF
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -261,8 +261,9 @@ Whether a `Request` is eligible to be retried.
 """
 function retryable end
 
-retryable(r::Request) = (isbytes(r.body) || (r.body !== nothing && ismarked(r.body))) &&
-    allow_retries(r) && (isidempotent(r) || retry_non_idempotent(r)) && !retrylimitreached(r)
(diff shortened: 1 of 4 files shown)


quinnj:
Ok, challenges here:
  * We currently check `retryable(stream)` in `readbody` to see if we should write the response body out to a user-provided response stream or "save it" temporarily in the request context
  * But if we let the user provide the `retry_check`, then that needs to also be factored into the `retryable(stream)` check
  * But I would expect, as the user, for the `retry_check` interface to be something like `f(req, resp, body, ex) -> Bool` where I'm indicating, for a given request/response/body/exception whether the request should be retried or not.

So I'm not quite sure what to do. There are valid cases (like [here](https://github.com/JuliaCloud/AWS.jl/blob/d6838228a59b490286724bef73cab65f95f5c75c/src/utilities/request.jl#L154)), where you might want to read the response body before deciding whether to do a retry. But how do we make that work if you provided a `response_stream`?

PR #878, opened 2022-07-12
Updated CORS example with 404 and 405 responses

BoZenKhaa:
Updated example so that 404 and 405 responses are returned correctly instead of 500 errors. Also, they have the CORS headers now.


quinnj (approved):
Thanks @BoZenKhaa!

PR #880, opened 2022-07-14
Add getparam and empty Listener definitions for convenience

quinnj:
We have `HTTP.getparams`, but it can be convenient to have `HTTP.getparam`
when you only need one param. The params are also considered immutable
so we don't need much else in terms of an AbstractDict interface.

Also adds a missing Listener definition so we support the case
like `HTTP.serve(; server=Sockets.listen(...))`.


fredrikekre:
Could use a test

PR #881, opened 2022-07-15
wait on the correct object, instead of calling sleep

vtjnash:
Using sleep usually indicates a design issue with the code, and
introduced a few minor data races into this code. Provides various
opportunities for code cleanup also.

~~Seemed to be needed along with https://github.com/JuliaLang/MbedTLS.jl/pull/243, though I can't remember why.~~


src/Servers.jl
```diff
@@ -336,22 +324,21 @@ function listen!(f, listener::Listener;
...
-    ready_to_accept = Ref(false)
+    ready_to_accept = Base.Event()
     if verbose > 0
         tsk = @async LoggingExtras.withlevel(Logging.Debug; verbosity=verbose) do
             listenloop(f, listener, conns, tcpisvalid, max_connections, readtimeout, access_log, ready_to_accept)
         end
     else
         tsk = @async listenloop(f, listener, conns, tcpisvalid, max_connections, readtimeout, access_log, ready_to_accept)
     end
+    # This rationale is utter bullshit, but is our fake reason for this event to exist:
```
quinnj, line 335: oof. harsh.

src/Servers.jl
```diff
@@ -139,46 +139,34 @@ closedorclosing(st) = st == CLOSING || st == CLOSED
...
 end
 
-# graceful shutdown that waits for active connectiosn to finish being handled
+# graceful shutdown that waits for active connections to finish being handled
 function Base.close(s::Server)
     shutdown(s.on_shutdown)
     close(s.listener)
     # first pass to mark or request connections to close
-    for c in s.connections
+    for c in collect(s.connections)
```
quinnj, line 161: why `collect` here?

src/Servers.jl
quinnj:
```suggestion
```

src/Servers.jl
vtjnash: close is a yield point and you are missing a lock here. The collect helps minimize the data race window

src/Servers.jl
vtjnash: Shouldn't you delete the rest of the comment and the ready_to_accept event code simultaneously too?

quinnj: Definitely welcome to refactor things; the reason I added this was because we had all these cases in tests/scripts where people did:
```julia
@async run_localhost_server()
sleep(1)
resp = HTTP.get(localhost)
```
the `sleep` was needed because the `HTTP.get` call would return an error if the server hadn't fully started yet.

Now that I think about it, we're probably already in a better place here because I already refactored with the `Listener` piece which ensures `HTTP.listen!` doesn't return until we've already got a listening tcp server socket. So yeah, we can probably just remove this stuff.

vtjnash: yeah, as soon as the socket is open, it is ready to listen, so this isn't needed

quinnj:
I think this has been integrated piecemeal in other PRs.

PR #888, opened 2022-07-18
disallow bugs in `@try` usage

vtjnash:
Catching all exceptions is considered bad practice, which is why the
default `try` block forbids it.

### DIFF
--- a/src/Exceptions.jl
+++ b/src/Exceptions.jl
@@ -5,15 +5,20 @@ import ..HTTP # for doc references
 
 @eval begin
 """
-    @try expr
+    @try Permitted Error Types expr
 
 Convenience macro for wrapping an expression in a try/catch block
 where thrown exceptions are ignored.
 """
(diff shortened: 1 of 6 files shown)


vtjnash:
I noticed one more that is currently hiding a reliability problems in `Base.close(c::Connection)`

quinnj:
I'd like to merge this, but the 1.6 CI failures look consistent (something hanging in the server tests). I can probably try to debug later today, unless @vtjnash can take a look.

quinnj:
Tried some things to fix the CI tests for 1.6 in https://github.com/JuliaWeb/HTTP.jl/pull/908; but the ubuntu build is still hanging. Not sure why. It seems like the sockets are little sluggish or something where we expect them to have received data, but they need an extra `yield()` in order to get kickstarted.

quinnj:
Merged in #908

PR #891, opened 2022-07-23
Avoid warning - merge is deprecated

2005m:
I am getting a warning because merge is deprecated in URIs.jl.
I think that should fix it.


quinnj:
Yeah, this LGTM; I'm a bit concerned about the CI hang/timeout on nightlies though. Most likely not related to this PR though.

PR #893, opened 2022-07-25
Add $bytes_received to log number of (body) bytes read from client, fixes #892.

### LINKED ISSUE #892: How to log size of incoming request?
- Julia 1.7.3
 - HTTP.jl 1

The logging function supplied with HTTP.jl checks the `stream.nwritten` field to log the size of the data sent out.

https://github.com/JuliaWeb/HTTP.jl/blob/master/src/Streams.jl#L16

But I would like to also log what the size of the data coming in was. At the time when the logging function is called, the other field `ntoread` is always 0, which points to it being a state variable that is 0 once the whole incoming message has been read.

How would I correctly get the size of the received message? I've considered looking at `stream.message.body` but I'm not sure if that's the way to go.

fredrikekre:
NGINX `$bytes_received` includes header bytes as well, I think (http://nginx.org/en/docs/stream/ngx_stream_core_module.html#var_bytes_received) so perhaps this should be called `$body_bytes_receieved`? Or perhaps this should be updated to also count header bytes. (NGINX doesn't have `$body_bytes_receieved`.)

### DIFF
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -13,7 +13,8 @@ mutable struct Stream{M <: Message, S <: IO} <: IO
     readchunked::Bool
     warn_not_to_read_one_byte_at_a_time::Bool
     ntoread::Int
-    nwritten::Int
+    nread::Int    # number of read body bytes
+    nwritten::Int # number of written body bytes
 end
 
 """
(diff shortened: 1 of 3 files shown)


quinnj:
Closing as obsolete; the access logging wasn't ported forward in teh 2.0 release. I think we could still bring that forward relatively easily if we wanted, but this would have to be rebased or coudl be folded in anyway.

PR #897, opened 2022-07-30
Fix HTTP.download stack overflow error on "/" paths

quinnj:


quinnj:
cc: @oxinabox. I'm not super familiar with the download code, but this seems to be in the same spirit as what `file_from_target` is aiming to do. I'll try to figure out if I can see how we usually right tests for download stuff as well.

oxinabox (approved):

oxinabox:
> I'll try to figure out if I can see how we usually right tests for download stuff as well.

We usually write tests using http://test.greenbytes.de/tech/tc2231/ which is like reference API for this particular part of the HTTP spec.

I think this one actually corresponds to the absolute path part of spec which is tested at:
http://test.greenbytes.de/tech/tc2231/#attabspath

For which the correct behavour is:
> Either ignore the filename altogether, or discard the path information.

PR #901, opened 2022-08-09
fix monitor_idle_connection for MbedTLS.SSLContext

vtjnash:
The comments were wrong, and also lacked error messages / handling.

Closes: https://github.com/JuliaLang/MbedTLS.jl/issues/250


quinnj (approved):
Thanks @vtjnash

PR #920, opened 2022-09-01
Create Invalidations.yml

ranocha:
This is based on https://github.com/julia-actions/julia-invalidations. Adding such checks came up in https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359. I suggest to add this check here since this package is widely used as a dependency.

See also SciML/MuladdMacro.jl#26 and SciML/MuladdMacro.jl#29

### DIFF
--- a/.github/workflows/Invalidations.yml
+++ b/.github/workflows/Invalidations.yml
@@ -1,0 +1,40 @@
+name: Invalidations
+
+on:
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: always.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  evaluate:
+    # Only run on PRs to the default branch.
+    # In the PR trigger above branches can be specified only explicitly whereas this check should work for master, main, or any other default branch
+    if: github.base_ref == github.event.repository.default_branch
+    runs-on: ubuntu-latest
+    steps:
+    - uses: julia-actions/setup-julia@v1
+      with:
+        version: '1'
+    - uses: actions/checkout@v3
+    - uses: julia-actions/julia-buildpkg@v1
+    - uses: julia-actions/julia-invalidations@v1
+      id: invs_pr
+
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


quinnj (approved):

quinnj:
@ranocha, I'm partially regretting merging this. I have a PR open where this new action fails (https://github.com/JuliaWeb/HTTP.jl/actions/runs/3099619353/jobs/5018972872), but the output is......non-existant? Like, I can't find any output telling me why the job failed, or what to do about the failure. Am I missing something obvious in the output there? Can we improve the github action experience to be more useful?

ranocha:
The step ("report invalidation counts") just before the failing step reports the reason: https://github.com/JuliaWeb/HTTP.jl/actions/runs/3099619353/jobs/5018972872
The number of invalidations increased be 2, so it does not appear to be a big deal to me at first sight. If you want to dig deeper, you need to work with SnoopCompile.jl locally.

quinnj:
Yeah, my comment was more directed to the fact that
```
 echo "Invalidations on default branch: 723 (723 via deps)" >> $GITHUB_STEP_SUMMARY
  echo "This branch: 725 (725 via deps)" >> $GITHUB_STEP_SUMMARY
```
is, in my opinion, pretty unclear as to what is actually being communicated. What is "default branch"? What is "$GITHUB_STEP_SUMMARY"? What is the 732 vs. (723 via deps) and why would those be different?

Do you see what I mean? The invalidations job failed and I go look at this output and I just find it very unhelpful. Can it not spell out the individual invalidations that changed between default branch vs. PR? Can it organize the information in a table or something that would be more clear to someone passing by and wanting to see and understand quickly what went wrong?

ranocha:
You're definitely right, there is a lot of potential to improve the action. Some ideas are floating around, e.g., https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359/22 with an open PR https://github.com/timholy/SnoopCompile.jl/pull/303. I think it will take some time but we will be able to improve it

PR #929, opened 2022-10-01
debuglayer(): fix DEBUG_LEVEL handling

alyst:
If *verbose* kwarg was not specified, `DEBUG_LEVEL[]` should have been used instead, but it was not.


quinnj (approved):
Thanks!

PR #750, opened 2021-09-05
Add minor precompile workload

rikhuijzer:
Just like my attempt in #738, I'm on a mission to reduce the time to first response. I still think that PR has some potential, but adding precompilation is more low hanging fruit.

As a proof-of-concept, I've only added `HTTP.get` for now.

## Demo

I've run `@time using HTTP; @time HTTP.get("http://example.com")` twice on `master` and `precompile` with Julia 1.7-beta3.0:

```
rik@nixos ~/g/HTTP.jl (master)> julia --project -ie '@time using HTTP; @time HTTP.get("http://example.com")'
using Revise
[ Info: Precompiling HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
  2.374460 seconds (1.87 M allocations: 102.031 MiB, 0.87% gc time, 1.02% compilation time)
  3.711768 seconds (7.92 M allocations: 418.646 MiB, 6.11% gc time)

rik@nixos ~/g/HTTP.jl (precompile)> julia --project -ie '@time using HTTP; @time HTTP.get("http://example.com")'
using Revise
[ Info: Precompiling HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
  4.184486 seconds (1.99 M allocations: 109.878 MiB, 0.51% gc time, 0.58% compilation time)
  3.053110 seconds (4.49 M allocations: 239.125 MiB, 1.89% gc time)

rik@nixos ~/g/HTTP.jl (master)> julia --project -ie '@time using HTTP; @time HTTP.get("http://example.com")'
using Revise
[ Info: Precompiling HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
  2.416790 seconds (1.87 M allocations: 102.046 MiB, 0.88% gc time, 1.02% compilation time)
  3.486002 seconds (7.92 M allocations: 418.657 MiB, 4.36% gc time)

rik@nixos ~/g/HTTP.jl (precompile)> julia --project -ie '@time using HTTP; @time HTTP.get("http://example.com")'
using Revise
[ Info: Precompiling HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
  4.140951 seconds (1.99 M allocations: 109.938 MiB, 0.48% gc time, 0.59% compilation time)
  3.161522 seconds (4.49 M allocations: 239.152 MiB, 1.82% gc time)
```

[...]

PR #930, opened 2022-10-03
Added precompile workload

louisponet:
Hi All,

I've attempted to solve some of the TTFX things in HTTP.jl. By adding the small workload, the time to first return of the request in went down by a third (3.41s vs 5.1s) for this:
```julia
using HTTP
@time HTTP.post("http://httpbin.org/post"; body="hey")
```

I had to do the `empty!(resize!(...))` in order not to trigger the `access_threaded` assert.

Let me know what you think!


rikhuijzer:
Review comment from a similar PR: https://github.com/JuliaWeb/HTTP.jl/pull/750#issuecomment-927303657

louisponet:
Well, that particular comment won't happen with the `listenany` I guess

src/workload.jl
```diff
@@ -0,0 +1,13 @@
+using SnoopPrecompile
+@precompile_all_calls begin
+    resize!(empty!(Parsers.status_line_regex),           Threads.nthreads())
+    resize!(empty!(Parsers.request_line_regex),          Threads.nthreads())
+    resize!(empty!(Parsers.header_field_regex),          Threads.nthreads())
+    resize!(empty!(Parsers.obs_fold_header_field_regex), Threads.nthreads())
+    resize!(empty!(Parsers.empty_header_field_regex),    Threads.nthreads())
+    port, server = listenany(ip"0.0.0.0", 8080)
```
quinnj, line 8:
```suggestion
```

src/workload.jl
```diff
@@ -0,0 +1,13 @@
...
+    router = Router()
+    register!(router, "GET", "/read/**", _ -> Response(200))
+    t = @async serve(router, "0.0.0.0", port, server=server)
```
quinnj, line 11:
```suggestion
    server = HTTP.serve!(router, "0.0.0.0"; listenany=true)
    port = HTTP.port(server)
```

src/workload.jl
```diff
@@ -0,0 +1,13 @@
...
+    resp = get("http://localhost:$port/read//$(homedir())")
```
quinnj, line 12:
```suggestion
    try
        resp = get("http://localhost:$port/read//$(homedir())")
    finally
        close(server)
    end
```

quinnj:
I _think_ the issue w/ CI tests here is because we're starting the server on port 8080 during precompile and then it's not getting properly closed and then the tests are going haywire when they first try to run a server/request on port 8080. I've left some code review suggestions that should hopefully help w/ this.

louisponet:
I’ll have a look, thanks!

louisponet:
There are already some tests that actually rely on no server running on 8080, I'll add listenany to those as well if that's okay. (I incidentally had another server running on 8080 and so tests errored)

quinnj:
Hmmm, it looks like the client.jl tests are hanging, so maybe my theory wasn't quite right.

louisponet:
On my pc the same tests fail on this branch as on the main, not sure if that’s the same on the github ci though

src/workload.jl
Oddestduck: > 
@rikhuijzer 
Screenshot_20230413_211132_Samsung Internet

quinnj:
Closing as stale

PR #932, opened 2022-10-07
fixed index from length for a couple of cases

jdiaz97:
Hello. Updated the code according to: https://docs.julialang.org/en/v1/base/arrays/#Base.eachindex


quinnj (approved):

quinnj:
Thanks!

Issue #934, opened 2022-10-11
https server crashes on disconnect without ssl handshake


tanmaykm:
- Julia 1.7.3
 - HTTP.jl master
 - MbedTLS.jl 1.1.6

```julia
using HTTP, MbedTLS, Sockets

sslconfig = MbedTLS.SSLConfig("fullchain.pem", "privkey.pem")
server = HTTP.listen!(; sslconfig=sslconfig, verbose=true) do http::HTTP.Stream
    HTTP.setstatus(http, 200)
    HTTP.startwrite(http)
    write(http, "response body\n")
end

sleep(5)
sock = connect("127.0.0.1", 8081)
close(sock)

wait(server)
```

The above code results in the server exiting with:

```julia
┌ LogLevel(-1): Listening on: 127.0.0.1:8081
└ @ HTTP.Servers ~/.julia/packages/HTTP/VAOuj/src/Servers.jl:361
ERROR: LoadError: TaskFailedException
Stacktrace:
 [1] wait
   @ ./task.jl:334 [inlined]
 [2] wait(s::HTTP.Servers.Server{HTTP.Servers.Listener{SSLConfig, Sockets.TCPServer}})
   @ HTTP.Servers ~/.julia/packages/HTTP/VAOuj/src/Servers.jl:125
 [3] top-level scope
   @ /data/Work/julia/JuliaTeamTestDir/master/certs/testhttp.jl:19

    nested task error: type EOFError has no field code
    Stacktrace:
     [1] getproperty(x::EOFError, f::Symbol)
```

tanmaykm:
Should `EOFError` be handled separately [here](https://github.com/JuliaWeb/HTTP.jl/blob/4364f2a231f783fd91e877f9dad0d4fdef537fbd/src/Servers.jl#L387) ?

PR #935, opened 2022-10-11
do not access `code` field if error is EOFError

tanmaykm:
Check for EOFError before attempting to access the `code` field of exceptions in the listen loop.
Fixes: #934

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -384,7 +384,7 @@ function listenloop(f, listener, conns, tcpisvalid,
                 Base.release(sem)
             end
         catch e
-            if e isa IOError && e.code == Base.UV_ECONNABORTED
+            if e isa Base.IOError && e.code == Base.UV_ECONNABORTED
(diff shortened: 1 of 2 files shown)


src/Servers.jl
```diff
@@ -384,7 +384,7 @@ function listenloop(f, listener, conns, tcpisvalid,
                 Base.release(sem)
             end
         catch e
-            if e.code == Base.UV_ECONNABORTED
+            if !isa(e, EOFError) && (e.code == Base.UV_ECONNABORTED)
```
fredrikekre, line 387: There are many other error types without a `code` field, shouldn't this check if `e` is or correct type instead? I guess `IOError`?
```suggestion
            if e isa IOError && e.code == Base.UV_ECONNABORTED
```

tanmaykm: Yes, makes sense

quinnj (approved):
LGTM; thanks @tanmaykm! (and @fredrikekre for the review/suggestions)

PR #942, opened 2022-10-18
Try with timeout rethrow

gustafsson:
The previous behaviour was to return any exception as res which I believe was a bug.

### DIFF
--- a/src/Exceptions.jl
+++ b/src/Exceptions.jl
@@ -33,29 +33,33 @@ function try_with_timeout(f, shouldtimeout, delay, iftimeout=() -> nothing)
         notify(cond, f())
     catch e
         @debugv 1 "error executing f in try_with_timeout"
-        notify(cond, e)
+        isopen(timer) && notify(cond, e, error = true)
     end
     # start a timer
     timer = Timer(delay; interval=delay / 10) do tm
-        if shouldtimeout()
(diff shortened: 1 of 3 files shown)


quinnj:
Hmmm, can you clarify the issue you're seeing that's prompting the changes here? I'm not sure I follow your PR description details.

gustafsson:
If I recall corretly I ran into this a while back when the connection was closed during `sslupgrade` and something in MbedTLS threw an exception that didn't get rethrown by try_with_timeout.

src/Exceptions.jl
```diff
@@ -33,29 +33,33 @@ function try_with_timeout(f, shouldtimeout, delay, iftimeout=() -> nothing)
         notify(cond, f())
     catch e
         @debugv 1 "error executing f in try_with_timeout"
-        notify(cond, e)
+        isopen(timer) && notify(cond, e, error = true)
```
quinnj, line 36: Why do we need to check `isopen(timer)` here? It seems like it wouldn't matter whether it's open or not?

gustafsson: `iftimeout()` may cause `f` to throw. In that scenario I'd rather pick up the `TimeoutError` in `wait(cond)` and swallow the exception from `f`

PR #943, opened 2022-10-18
connectionlayer: sslupgrade(socket_type_tls) fixup (#941)

gustafsson:
The sockettype for ConnectionPool.sslupgrade should depend on target_url  rather than url. These are not necessarily on the same scheme. I.e if  "https_proxy=http://..."


quinnj (approved):
Ah, good catch. Thanks for the PR.

PR #949, opened 2022-10-31
Add an example to parse_multipart_form docs

cmcaine:


quinnj:
This LGTM; let me try to fix the CI so we can get a clean run here since these are doc tests.

quinnj:
Looks like there's some kind of syntax error in the multipart file that is causing tests to fail.

cmcaine:
I just copied and pasted the example from my REPL, but I wasn't on master, maybe the exact printed output has changed or something.

I can try again on Friday

quinnj:
Sorry for the slow response/review here; exampled added in #1311 for the 2.X release cycle

PR #950, opened 2022-10-31
Reduce slightly allocations when reading/writing HTTP messages

nickrobinson251:
1. write direct to io rather than first creating strings
    - we need to be careful to `string` any numbers before doing this, but doing this selectively still reduces allocations
2. don't create a `VersionNumber` every time we parse a message (in `parse_status_line!`/`parse_request_line!`), as the `VersionNumber` constructor relies on regex matching which allocates, instead we can define our own much simpler type with allocation-free parsing

Together this takes a simple GET request from 170 -> 155 allocations

### DIFF
--- a/src/Messages.jl
+++ b/src/Messages.jl
@@ -77,7 +77,7 @@ abstract type Message end
 
 Represents an HTTP response message with fields:
 
-- `version::VersionNumber`
+- `version::HTTPVersion`
    [RFC7230 2.6](https://tools.ietf.org/html/rfc7230#section-2.6)
 
 - `status::Int16`
@@ -94,14 +94,14 @@ Represents an HTTP response message with fields:
 
 """
 mutable struct Response <: Message
-    version::VersionNumber
+    version::HTTPVersion
     status::Int16
     headers::Headers
     body::Any # Usually Vector{UInt8} or IO
     request::Union{Message, Nothing} # Union{Request, Nothing}
 end
 
-function Response(status::Integer, headers, body; version::VersionNumber=v"1.1", request=nothing)
+function Response(status::Integer, headers, body; version=HTTPVersion(1,1), request=nothing)
     b = isbytes(body) ? bytes(body) : something(body, nobody)
     @assert (request isa Request || request === nothing)
     return Response(version, status, mkheaders(headers), b, request)
@@ -119,7 +119,7 @@ Response(body) = Response(200; body=body)
 Base.convert(::Type{Response}, s::AbstractString) = Response(s)
 
 function reset!(r::Response)
-    r.version = v"1.1"
+    r.version = HTTPVersion(1, 1)
     r.status = 0
     if !isempty(r.headers)
         empty!(r.headers)
@@ -136,8 +136,10 @@ body(r::Response) = getfield(r, :body)
 const Context = Dict{Symbol, Any}
 
 """
-    HTTP.Request(method, target, headers=[], body=nobody;
-        version=v"1.1", url::URI=URI(), responsebody=nothing, parent=nothing, context=HTTP.Context())
+    HTTP.Request(
+        method, target, headers=[], body=nobody;
+        version=v"1.1", url::URI=URI(), responsebody=nothing, parent=nothing, context=HTTP.Context()
+    )
 
 Represents a HTTP Request Message with fields:
 
@@ -147,7 +149,7 @@ Represents a HTTP Request Message with fields:
 - `target::String`
    [RFC7230 5.3](https://tools.ietf.org/html/rfc7230#section-5.3)
 
-- `version::VersionNumber`
+- `version::HTTPVersion`
    [RFC7230 2.6](https://tools.ietf.org/html/rfc7230#section-2.6)
 
 - `headers::HTTP.Headers`
@@ -170,7 +172,7 @@ Represents a HTTP Request Message with fields:
 mutable struct Request <: Message
     method::String
     target::String
-    version::VersionNumber
+    version::HTTPVersion
     headers::Headers
     body::Any # Usually Vector{UInt8} or some kind of IO
     response::Response
@@ -181,8 +183,10 @@ end
 
 Request() = Request("", "")
 
-function Request(method::String, target, headers=[], body=nobody;
-                 version=v"1.1", url::URI=URI(), responsebody=nothing, parent=nothing, context=Context())
+function Request(
+    method::String, target, headers=[], body=nobody;
+    version=HTTPVersion(1,1), url::URI=URI(), responsebody=nothing, parent=nothing, context=Context()
+)
     b = isbytes(body) ? bytes(body) : body
     r = Request(method, target == "" ? "/" : target, version,
                 mkheaders(headers), b, Response(0; body=responsebody),
@@ -463,12 +467,7 @@ function decode(m::Message, encoding::String="gzip")::Vector{UInt8}
 end
 
 # Writing HTTP Messages to IO streams
-"""
-    HTTP.writehttpversion(io, ::Message)
-
-e.g. `"HTTP/1.1"`
-"""
-writehttpversion(io, m::Message) = write(io, "HTTP/", m.version.major, ".", m.version.minor)
+Base.write(io::IO, v::HTTPVersion) = write(io, "HTTP/", string(v.major), ".", string(v.minor))
 
 """
     writestartline(::IO, ::Message)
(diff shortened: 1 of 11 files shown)


src/Parsers.jl
```diff
@@ -205,7 +205,7 @@ function parse_status_line!(bytes::AbstractString, response)::SubString{String}
     if !exec(re, bytes)
         throw(ParseError(:INVALID_STATUS_LINE, bytes))
     end
-    response.version = VersionNumber(group(1, re, bytes))
+    response.version = HTTPVersion(group(1, re, bytes))
```
nickrobinson251, line 208: here, and a few lines above, is where we were allocating as part of construction `VersionNumber`

src/Strings.jl
```diff
@@ -1,9 +1,98 @@
...
+        minor = parse(UInt8, d2)
+    else
+        return nothing
+    end
+    return HTTPVersion(major, minor)
+end
+
+# const RE_HTTP_VERSION = r"(\d+)(?:\.(\d+))?"
+# function Base.tryparse(::Type{HTTPVersion}, v::AbstractString)
+#     m = match(RE_HTTP_VERSION, String(v)::String)
```
nickrobinson251, line 88: this is how `VersionNumber` construction works, for comparison

I need to remove this

src/Strings.jl
```diff
@@ -1,9 +1,98 @@
 module Strings
 
-export escapehtml, tocameldash, iso8859_1_to_utf8, ascii_lc_isequal
+export HTTPVersion, escapehtml, tocameldash, iso8859_1_to_utf8, ascii_lc_isequal
```
nickrobinson251, line 3: this is in `src/Strings.jl` because it was the submodule that was imported in exactly the same places we need  `HTTPVersion` 

happy to move it somewhere else

src/Strings.jl
```diff
@@ -1,9 +1,98 @@
...
+# because we don't want to add overhead for a case that should never be relevant.
+Base.convert(::Type{HTTPVersion}, v::VersionNumber) = HTTPVersion(v.major, v.minor)
+Base.VersionNumber(v::HTTPVersion) = VersionNumber(v.major, v.minor)
+
+Base.show(io::IO, v::HTTPVersion) = print(io, "HTTPVersion(\"", string(v.major), ".", string(v.minor), "\")")
+
+Base.:(==)(va::VersionNumber, vb::HTTPVersion) = va == VersionNumber(vb)
+Base.:(==)(va::HTTPVersion, vb::VersionNumber) = VersionNumber(va) == vb
+
+Base.isless(va::VersionNumber, vb::HTTPVersion) = isless(va, VersionNumber(vb))
```
nickrobinson251, line 40: we either need to define these methods, or else overload `getproperty` such that `r.version` returns a `VersionNumber` and we do all comparisons with `VersionNumbers`. Either seems fine to me.

src/Messages.jl
```diff
@@ -94,14 +94,14 @@ Represents an HTTP response message with fields:
...
-    version::VersionNumber
+    version::HTTPVersion
     status::Int16
     headers::Headers
     body::Any # Usually Vector{UInt8} or IO
     request::Union{Message, Nothing} # Union{Request, Nothing}
 end
 
-function Response(status::Integer, headers, body; version::VersionNumber=v"1.1", request=nothing)
+function Response(status::Integer, headers, body; version=HTTPVersion(1,1), request=nothing)
```
quinnj, line 104:
```suggestion
function Response(status::Integer, headers, body; version=HTTPVersion(1, 1), request=nothing)
```

src/Messages.jl
```diff
@@ -182,8 +184,10 @@ end
 
 Request() = Request("", "")
 
-function Request(method::String, target, headers=[], body=nobody;
-                 version=v"1.1", url::URI=URI(), responsebody=nothing, parent=nothing, context=Context())
+function Request(
+    method::String, target, headers=[], body=nobody;
+    version=HTTPVersion(1,1), url::URI=URI(), responsebody=nothing, parent=nothing, context=Context()
```
quinnj, line 189:
```suggestion
    version=HTTPVersion(1, 1), url::URI=URI(), responsebody=nothing, parent=nothing, context=Context()
```

src/Messages.jl
```diff
@@ -455,25 +459,20 @@ function decode(m::Message, encoding::String="gzip")::Vector{UInt8}
...
+Base.write(io::IO, v::HTTPVersion) = write(io, "HTTP/", string(v.major), ".", string(v.minor))
 
 """
     writestartline(::IO, ::Message)
 
 e.g. `"GET /path HTTP/1.1\\r\\n"` or `"HTTP/1.1 200 OK\\r\\n"`
 """
 function writestartline(io::IO, r::Request)
-    write(io, "$(r.method) $(r.target) $(httpversion(r))\r\n")
+    write(io, r.method, " ", r.target, " ", r.version, "\r\n")
```
quinnj, line 470: So instead of creating a big string and writing that, we iterate over the args and each is written to io?

src/Strings.jl
quinnj: I think that's fine

src/Strings.jl
quinnj: This is not that much code, so this seems fine

quinnj (approved):
LGTM; thanks @nickrobinson251!

src/Messages.jl
nickrobinson251: yeah, exactly, we save on allocating that big string before writing it out
https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-string-interpolation-for-I/O

and we overload `Base.write` for `version` so it also writes what we want straight to IO

so each saving 1 string allocation

nickrobinson251:
tests pass locally. not sure if we want to resolve #951 and get a clean CI run before merging here?

src/clientlayers/MessageRequest.jl
```diff
@@ -12,7 +13,7 @@ Construct a [`Request`](@ref) object from method, url, headers, and body.
 Hard-coded as the first layer in the request pipeline.
 """
 function messagelayer(handler)
-    return function(method::String, url::URI, headers::Headers, body; response_stream=nothing, http_version=v"1.1", kw...)
+    return function(method::String, url::URI, headers::Headers, body; response_stream=nothing, http_version=HTTPVersion(1,1), kw...)
```
nickrobinson251, line 16:
```suggestion
    return function(method::String, url::URI, headers::Headers, body; response_stream=nothing, http_version=HTTPVersion(1, 1), kw...)
```

src/Messages.jl
```diff
@@ -119,7 +119,7 @@ Response(body) = Response(200; body=body)
 Base.convert(::Type{Response}, s::AbstractString) = Response(s)
 
 function reset!(r::Response)
-    r.version = v"1.1"
+    r.version = HTTPVersion(1, 1)
```
fredrikekre, line 122: Could construct the version number without going through a string instead maybe?

nickrobinson251: This case is actually fine, we could leave this as-is. 

The problematic case is `parse_status_line!`/`parse_request_line!` where we're parsing the version out of the response/request body

src/Messages.jl
```diff
@@ -455,25 +459,20 @@ function decode(m::Message, encoding::String="gzip")::Vector{UInt8}
...
 """
 function writestartline(io::IO, r::Request)
-    write(io, "$(r.method) $(r.target) $(httpversion(r))\r\n")
+    write(io, r.method, " ", r.target, " ", r.version, "\r\n")
     return
 end
 
 function writestartline(io::IO, r::Response)
-    write(io, "$(httpversion(r)) $(r.status) $(statustext(r.status))\r\n")
+    write(io, r.version, " ", string(r.status), " ", statustext(r.status), "\r\n")
```
quinnj, line 475: I think this is causing a test failure; the problem is that we're dispatching to this method in Base:
```julia
function write(io::IO, x1, xs...)
    written::Int = write(io, x1)
    for x in xs
        written += write(io, x)
    end
    return written
end
```
which isn't ideal when writing to a Socket since it might actually "flush" the writes out on the wire at any time, but it means the response might be received in fragments. This becomes worse when you have a system under heavy load and we have a scenario where it writes `write(socket, "HTTP/1.1")`, then gets task switched away for several minutes before switching back and writing `write(socket, 200)`.

But what's the best way to solve while trying to avoid allocations? Not super clear, but one idea is that we have a `writebuffer::IOBuffer` in the `Connection` object where these writes would go and then we make sure we have a dedicated `flush` operation at some point (like after `writestartline`, `writeheaders`, `writebody`, etc.).

PR #955, opened 2022-11-03
A few server cleanups

quinnj:
* Spawn server task on interactive threadpool if available
* Allow passing `verbose=-1` to silence initial server listening log
* Use Event instead of Ref for waiting when listenloop is ready

### DIFF
--- a/src/Servers.jl
+++ b/src/Servers.jl
@@ -324,6 +324,17 @@ function listen(f, listener::Listener; kw...)
     return server
 end
 
+# compat for `Threads.@spawn :interactive expr`
+@static if hasmethod(getfield(Threads, Symbol("@spawn")), Tuple{LineNumberNode, Module, Symbol, Expr})
+    macro _spawn_interactive(ex)
+        esc(:(Threads.@spawn :interactive $ex))
+    end
+else
+    macro _spawn_interactive(ex)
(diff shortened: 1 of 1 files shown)


src/Servers.jl
```diff
@@ -324,6 +324,17 @@ function listen(f, listener::Listener; kw...)
...
 
+# compat for `Threads.@spawn :interactive expr`
+@static if hasmethod(getfield(Threads, Symbol("@spawn")), Tuple{LineNumberNode, Module, Symbol, Expr})
+    macro _spawn_interactive(ex)
+        esc(:(Threads.@spawn :interactive $ex))
+    end
+else
+    macro _spawn_interactive(ex)
+        esc(:(Threads.@spawn $ex))
+    end
```
NHDaly, line 335: Should this be `@async` to keep it more similar to the behavior pre-1.9?

quinnj: Yeah, that's probably a good idea.

NHDaly:
Did you release this as a breaking change? IMO it is currently, a breaking change, since it will mean without interactive threads support, this update would cause HTTP to start spawning tasks on a different thread than it did before, which could break availability. (It would be a breaking change for us at RAI since it would make our "reserved thread" not work anymore for HTTP requests.)

PR #965, opened 2022-11-21
Allow silencing server closing log message as well

JackDunnNZ:
#955 allowed passing `verbose=-1` to silence the initial server listening log, it seems appropriate that this setting should also silence the server closing log


quinnj (approved):
Thanks!

PR #966, opened 2022-11-23
Keep HEAD request method when redirecting which follows curl behavior

quinnj:


quinnj:
Closed in favor of https://github.com/JuliaWeb/HTTP.jl/pull/967

PR #967, opened 2022-11-23
Do not turn HEAD requests to GET by default through redirects

giordano:
I don't think this is a breaking change, in fact I didn't have to change any existing tests: this is only a change in the default behaviour if `redirect_method` is not set, but the user can always override it by setting that keyword argument.

This is also consistent with what `Downloads.jl` does:
```julia
julia> using Downloads

julia> io = IOBuffer();

julia> Downloads.request("https://google.com/"; output=io, method="HEAD"); # The HEAD request produces an empty body also through redirect

julia> length(String(take!(io)))
0
```


quinnj (approved):
Thanks @giordano!

PR #972, opened 2022-11-30
Fix access_threaded in the forthcoming interactive threadpool world

### LINKED ISSUE #970: serve() fails when run from an :interactive thread (Julia >=1.9)
The minimal example fails to return a valid response (but doesn't throw any errors) when run from an `:interactive` thread, but works fine if `:interactive` is removed.  Julia should be started with `-t2,1`.
```julia
import HTTP

function run()
  HTTP.serve() do request::HTTP.Request
    return HTTP.Response(200, "hello")
  end
end

t = Threads.@spawn :interactive run()
wait(t)
```
```
$ curl -v 127.0.0.1:8081
*   Trying 127.0.0.1:8081...
* Connected to 127.0.0.1 (127.0.0.1) port 8081 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:8081
> User-Agent: curl/7.85.0
> Accept: */*
>
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```
```
(foo) pkg> st
Status `/tmp/foo/Project.toml`
  [cd3eb016] HTTP v1.5.5

julia> versioninfo()
Julia Version 1.10.0-DEV.46
Commit d0a211a9209 (2022-11-24 15:42 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × AMD Ryzen 7 4700U with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver2)
  Threads: 3 on 8 virtual cores
Environment:
  JULIA_DEPOT_PATH = /me/.julia
```
I also tested and got the same results with 1.9.0-alpha1.

quinnj:
Fixes #970.

The issue here is that our `access_threaded` function, and thread-specific arrays, all relied on `Threads.nthreads()`, which by default only returns the number of threads in the `:default` threadpool. In the new world where the `:interactive` threadpool exists, there can also be threads there.

There is a new function `Threads.maxthreadid()` which helpfully tells us the max id of any thread, regardless of threadpool. So the proposed fix here is that if that function is defined, use that, otherwise fallback to `Threads.nthreads()`.

### DIFF
--- a/src/Parsers.jl
+++ b/src/Parsers.jl
@@ -351,11 +351,12 @@ const unhex = Int8[
 function __init__()
     # FIXME Consider turing off `PCRE.UTF` in `Regex.compile_options`
     # https://github.com/JuliaLang/julia/pull/26731#issuecomment-380676770
-    resize!(empty!(status_line_regex),           Threads.nthreads())
-    resize!(empty!(request_line_regex),          Threads.nthreads())
-    resize!(empty!(header_field_regex),          Threads.nthreads())
-    resize!(empty!(obs_fold_header_field_regex), Threads.nthreads())
-    resize!(empty!(empty_header_field_regex),    Threads.nthreads())
+    nt = isdefined(Base.Threads, :maxthreadid) ? Threads.maxthreadid() : Threads.nthreads()
+    resize!(empty!(status_line_regex),           nt)
+    resize!(empty!(request_line_regex),          nt)
+    resize!(empty!(header_field_regex),          nt)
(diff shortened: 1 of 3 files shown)


quinnj:
cc: @KristofferC, since, if I remember correctly, he was the original author of this utility? Just as an FYI in case other packages will also need this change in the future (like Parsers.jl).

nickrobinson251 (approved):

PR #975, opened 2022-12-13
Fix setting of response body for StatusError exception

quinnj:
As reported in https://github.com/JuliaCloud/AWS.jl/issues/593, the recently refactored retry logic was failing to properly write the response body when a `response_stream` was used and a `StatusError` was thrown.


nickrobinson251 (approved):
Can we add a test?

JackDunnNZ:
Thanks @quinnj, this fixes all the problems we were seeing on our end!

PR #981, opened 2022-12-22
Set line info for `@client` macro as call site

nickrobinson251:
Make it easier to find where the `@client` was defined

e.g. 
Before 
```julia
julia> using CloudBase

julia> @which CloudBase.Azure.get()  # Points to where the `HTTP.@client` macro is defined
get(a...; kw...) in CloudBase.Azure at /Users/nickr/.julia/packages/HTTP/RyUY8/src/HTTP.jl:460
```
Now
```julia
julia> using CloudBase

julia> @which CloudBase.Azure.get()  # Points to where `HTTP.@client` was called
get(a...; kw...) in CloudBase.Azure at /Users/nickr/repos/CloudBase.jl/src/CloudBase.jl:187
```

I don't know if there is a more standard way to do this 😊

### DIFF
--- a/src/HTTP.jl
+++ b/src/HTTP.jl
@@ -422,6 +422,10 @@ function request(stack::Base.Callable, method, url, h=Header[], b=nobody, q=noth
     return stack(string(method), request_uri(url, query), mkheaders(headers), body; kw...)
 end
 
+macro remove_linenums!(expr)
+    return esc(Base.remove_linenums!(expr))
+end
+
 """
     HTTP.@client requestlayers
     HTTP.@client requestlayers streamlayers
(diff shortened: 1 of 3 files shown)


src/HTTP.jl
```diff
@@ -456,16 +460,18 @@ hard-codes the value of the `retry` and `redirect` keyword arguments. When we pa
...
+        post(a...; kw...) = ($__source__; request("POST", a...; kw...))
+        patch(a...; kw...) = ($__source__; request("PATCH", a...; kw...))
+        head(a...; kw...) = ($__source__; request("HEAD", a...; kw...))
+        delete(a...; kw...) = ($__source__; request("DELETE", a...; kw...))
+        open(f, a...; kw...) = ($__source__; request(a...; iofunction=f, kw...))
+        function request(method, url, h=HTTP.Header[], b=HTTP.nobody; headers=h, body=b, query=nothing, kw...)::HTTP.Response
+            $__source__
             HTTP.request(HTTP.stack($requestlayers, $streamlayers), method, url, headers, body, query; kw...)
+        end
     end)
```
NHDaly, line 475:
```suggestion
    return Base.remove_linenums(quote
        get(a...; kw...) = ($__source__; request("GET", a...; kw...))
        put(a...; kw...) = ($__source__; request("PUT", a...; kw...))
        post(a...; kw...) = ($__source__; request("POST", a...; kw...))
        patch(a...; kw...) = ($__source__; request("PATCH", a...; kw...))
        head(a...; kw...) = ($__source__; request("HEAD", a...; kw...))
        delete(a...; kw...) = ($__source__; request("DELETE", a...; kw...))
        open(f, a...; kw...) = ($__source__; request(a...; iofunction=f, kw...))
        function request(method, url, h=HTTP.Header[], b=HTTP.nobody; headers=h, body=b, query=nothing, kw...)::HTTP.Response
            $__source__
            HTTP.request(HTTP.stack($(esc(requestlayers)), $(esc(streamlayers)), method, url, headers, body, query; kw...)
        end
    end)
```

nickrobinson251: if doing more fine-grained `esc`-ing, i think we'd also need something like `$(esc(:get))(a...; kw...) = ...` so that the `get` is a new function (not gensym'd and not overloading `HTTP.get` nor `Base.get`) but 👍 i still think your proposal in https://github.com/JuliaLang/julia/pull/31335 is sound

NHDaly: Ah yes, you're right. 👍 
For some reason i thought I remembered there was an exception in the macroexpansion rules that they _don't_ mangle the names of top-level function defs in a quote expression... but it seems you are correct:
```julia
julia> macro foo(value)
           return quote
               new_func() = $(esc(value))
           end
       end
@foo (macro with 1 method)

julia> @foo 2+2
#16#new_func (generic function with 1 method)

julia> new_func()
ERROR: UndefVarError: new_func not defined
```

NHDaly, line 475: > but 👍 i still think your proposal in https://github.com/JuliaLang/julia/pull/31335 is sound

👍 But i actually kind of really like the solution you have taken in this PR. Maybe we should just move your macro into Base, and call it `@remove_macro_linenums!` or `@remove_new_linenums!` or something, since it only removes the linenums from the local quote expression but leaves any that are part of the data of the expression, which is actually a really nice combination.

The more I think about it, I actually do think that the macro version you have is really _exactly_ what you want! It says: remove the linenums added by the fact that I'm creating an expression here, but don't remove any of the linenums that are in the data of my expression, either because they were passed-in by the caller, or because i've added them via `$__source__`. That's exactly what you want!

So i think that your solution is maybe _even better_ and we should just push yours up into Base? What do you think?

quinnj (approved):

PR #983, opened 2023-01-05
Parametrize `Connections` to avoid type instabilities

Drvi:
I noticed that due to instabilities, all calls to `bytesavailable` were allocating for `Connection`s due to the `io` field being abstractly typed (`IO`). Now these allocations are gone; a little parsing benchmark of a 500MB gzipped file:

```julia
# master
julia> @time streamedng(bucket, credentials, 4)
 18.736654 seconds (910.61 k allocations: 1.695 GiB)

# PR
julia> @time streamedng(bucket, credentials, 4)
 18.612146 seconds (552.58 k allocations: 1.695 GiB)
```

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -345,7 +345,12 @@ getpool(::Type{OpenSSL.SSLStream}) = OpenSSL_SSL_POOL
 # to opt out from locking, define your own `Pool` and add a `getpool` method for your IO type
 const POOLS_LOCK = Threads.ReentrantLock()
 function getpool(::Type{T}) where {T}
-    @lock POOLS_LOCK get!(POOLS, T, Pool(Connection{T}))::Pool{Connection{T}}
+    lock(POOLS_LOCK)
+    try
+        get!(POOLS, T, Pool(Connection{T}))::Pool{Connection{T}}
+    finally
+        unlock(POOLS_LOCK)
+    end
 end
 
 """
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -33,9 +33,6 @@ set_default_connection_limit!(n) = default_connection_limit[] = n
 
 taskid(t=current_task()) = string(hash(t) & 0xffff, base=16, pad=4)
 
-include("connectionpools.jl")
-using .ConnectionPools
-
 """
     Connection
 
@@ -58,7 +55,7 @@ Fields:
 - `readable`, whether the Connection object is readable
 - `writable`, whether the Connection object is writable
 """
-mutable struct Connection <: IO
+mutable struct Connection{IO_t <: IO} <: IO
     host::String
     port::String
     idle_timeout::Int
@@ -66,7 +63,7 @@ mutable struct Connection <: IO
     peerip::IPAddr # for debugging/logging
     peerport::UInt16 # for debugging/logging
     localport::UInt16 # debug only
-    io::IO
+    io::IO_t
     clientconnection::Bool
     buffer::IOBuffer
     timestamp::Float64
@@ -75,6 +72,9 @@ mutable struct Connection <: IO
     writebuffer::IOBuffer
     state::Any # populated & used by Servers code
 end
+
+include("connectionpools.jl")
+using .ConnectionPools
 
 """
     connectionkey
@@ -89,8 +89,8 @@ connectionkey(x::Connection) = (typeof(x.io), x.host, x.port, x.require_ssl_veri
 
 Connection(host::AbstractString, port::AbstractString,
            idle_timeout::Int,
-           require_ssl_verification::Bool, io::IO, client=true) =
-    Connection(host, port, idle_timeout,
+           require_ssl_verification::Bool, io::T, client=true) where {T}=
+    Connection{typeof(io)}(host, port, idle_timeout,
                 require_ssl_verification,
                 safe_getpeername(io)..., localport(io),
                 io, client, PipeBuffer(), time(), false, false, IOBuffer(), nothing)
@@ -340,7 +340,11 @@ end
 
 Global connection pool keeping track of active connections.
 """
-const POOL = Pool(Connection)
+const TCP_POOL = Pool(Connection{Sockets.TCPSocket})
+const SSL_POOL = Pool(Connection{MbedTLS.SSLContext})
+getpool(::Type{Sockets.TCPSocket}) = TCP_POOL
+getpool(::Type{MbedTLS.SSLContext}) = SSL_POOL
+getpool(::Connection{T}) where T = getpool(T)
 
 """
     newconnection(type, host, port) -> Connection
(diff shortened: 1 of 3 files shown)


src/connectionpools.jl
```diff
@@ -242,9 +245,16 @@ created and passed the `max`, `idle_timeout`, and `reuse` keyword arguments if p
 The provided function `f` must create a new connection instance of type `C`.
 The acquired connection MUST be returned to the pool by calling `release(pool, key, conn)` exactly once.
 """
-function acquire(f, pool::Pool{C}, key; forcenew::Bool=false, kw...) where {C}
+function acquire(f, pool::Pool{Connection{Sockets.TCPSocket}}, key; forcenew::Bool=false, kw...)
     pod = lock(pool.lock) do
-        get!(() -> Pod(C; kw...), pool.pods, key)
+        get!(() -> Pod(Connection{Sockets.TCPSocket}; kw...), pool.pods, key)
```
nickrobinson251, line 250: i think nothing in ConnectionPools is supposed to be `Connection` specific -- `Connection` is just one implementation on the required interface, but a user could provide an alternative one -- so i think this should stay as is... does it need to change? isn't `C` enough info anyway?

src/connectionpools.jl
```diff
@@ -3,6 +3,9 @@ module ConnectionPools
 export Pod, Pool, acquire, release
 
 import Base: acquire, release
+import Sockets
+import MbedTLS
+using ..ConnectionPool: Connection
```
nickrobinson251, line 8: i think this makes the modules mutually dependent on each other, which isn't right (see previous comment)

src/ConnectionPool.jl
```diff
@@ -89,8 +89,8 @@ connectionkey(x::Connection) = (typeof(x.io), x.host, x.port, x.require_ssl_veri
 
 Connection(host::AbstractString, port::AbstractString,
            idle_timeout::Int,
-           require_ssl_verification::Bool, io::IO, client=true) =
-    Connection(host, port, idle_timeout,
+           require_ssl_verification::Bool, io::T, client=true) where {T}=
+    Connection{typeof(io)}(host, port, idle_timeout,
```
nickrobinson251, line 93:
```suggestion
    Connection{T}(host, port, idle_timeout,
```

src/ConnectionPool.jl
```diff
@@ -58,15 +55,15 @@ Fields:
 - `readable`, whether the Connection object is readable
 - `writable`, whether the Connection object is writable
 """
-mutable struct Connection <: IO
+mutable struct Connection{IO_t <: IO} <: IO
```
nickrobinson251, line 58: this now make `Connection` a non-concrete type, which could be problematic if we ever have a collection of `Connection`s -- i see this PR address one such case by splitting `POOL` into two different `TCP_POOL` and `SSL_POOL` -- but i wonder whether or not `Connection` is public API and if we need to check this doesn't affect users? I _think_ it's not public API and we should be fine, but i know packages adding type params has caused me many headaches in the past

src/connectionpools.jl
nickrobinson251: i.e. i think `ConnectionPools` should not depend on `ConnectionPool` (which is a specific implementation of the `ConnectionPools` interface)

src/ConnectionPool.jl
Drvi: Good points, yeah. Looking at the docstring, `io` is meant to be one of the two types, but it is definitely a good idea to check how breaking this is.

Drvi:
@quinnj re the CI test failure on MacOS... locally, when I try running the `Autobahn WebSocket Tests` `Servers` I run into the following issue 
```
...
Servers = [u'ws://127.0.0.1:9002']
Connection to ws://127.0.0.1:9002 failed (Connection was refused by other side: 111: Connection refused.)
...
```
and no report gets generated.
However, when I change the URL in `test/websockets/config/fuzzingclient.json` from `"ws://127.0.0.1:9002"` to `"ws://host.docker.internal:9002"`, the tests are passing.

I don't understand why would this start failing as a result of the changes introduced in this PR, though... any ideas?

src/ConnectionPool.jl
Drvi: I did a quick scan through AWS.jl, Azure.jl, RAICode, Pluto.jl, Genie.jl and Github.jl and it seems that no one is really touching `Connection`s. It is also not exported by HTTP.jl. I'm hopeful that most people won't notice this change, and those who do more involved work regarding connections might appreciate fewer allocs.

src/ConnectionPool.jl
```diff
@@ -325,22 +325,40 @@ function purge(c::Connection)
...
+getpool(::Type{Sockets.TCPSocket}) = TCP_POOL
+getpool(::Type{MbedTLS.SSLContext}) = MbedTLS_SSL_POOL
+getpool(::Type{OpenSSL.SSLStream}) = OpenSSL_SSL_POOL
+# Fallback for custom connection io types
+# to opt out from locking, define your own `Pool` and add a `getpool` method for your IO type
+const POOLS_LOCK = Threads.ReentrantLock()
+function getpool(::Type{T}) where {T}
+    lock(POOLS_LOCK)
+    try
+        get!(POOLS, T, Pool(Connection{T}))::Pool{Connection{T}}
```
nickrobinson251, line 350: we probably don't want to actuall call `Pool` if we don't need to, as that's going to allocate a Dict 

```suggestion
        get!(() -> Pool(Connection{T}), POOLS, T)::Pool{Connection{T}}
```

src/ConnectionPool.jl
nickrobinson251: great- thanks for checking!

src/ConnectionPool.jl
Drvi: Ooof, good catch

src/ConnectionPool.jl
```diff
@@ -325,22 +325,40 @@ function purge(c::Connection)
...
+# Fallback for custom connection io types
+# to opt out from locking, define your own `Pool` and add a `getpool` method for your IO type
+const POOLS_LOCK = Threads.ReentrantLock()
+function getpool(::Type{T}) where {T}
+    lock(POOLS_LOCK)
+    try
+        get!(() -> Pool(Connection{T}), POOLS, T)::Pool{Connection{T}}
+    finally
+        unlock(POOLS_LOCK)
+    end
```
quinnj, line 353:
```suggestion
    Base.@lock POOLS_LOCK get!(() -> Pool(Connection{T}), POOLS, T)::Pool{Connection{T}}
```

quinnj (approved):
Thanks @Drvi; this looks great. I'm going to commit the one suggestion I made so we get one more CI run, but then I'll merge.

src/ConnectionPool.jl
Drvi: This broke Julia 1.6, where `@lock` doesn't exist.

quinnj: It wasn't exported yet in 1.6, so doing `Base.@lock` is a backwards compatible way to use it

nickrobinson251:
Is there a simple allocations test we can add for this PR, so we don't lose these gains over time? something along the lines of `@test @allocations(bytesavailable(conn)) == 0` ?

nickrobinson251 (approved):
an allocations-related test would be nice to have

otherwise looks great to me

PR #984, opened 2023-01-06
Allow passing pre-allocated buffer for response body

quinnj:
This is one piece in some efforts we're going to make to drastically reduce the # of allocations required for making HTTP requests.

This PR allows the user to pass a pre-allocated `Vector{UInt8}` via the `response_stream` keyword arg (previously not allowed), which will then be used directly for writing the response body.

cc: @nickrobinson251, @Drvi

Haven't added tests/docs yet, but wanted to at least get this up in case it overlaps w/ @Drvi's allocation hunting.

### DIFF
--- a/src/Streams.jl
+++ b/src/Streams.jl
@@ -1,6 +1,6 @@
 module Streams
 
-export Stream, closebody, isaborted, setstatus
+export Stream, closebody, isaborted, setstatus, readall!
 
 using Sockets, LoggingExtras
 using ..IOExtras, ..Messages, ..ConnectionPool, ..Conditions, ..Exceptions
@@ -268,7 +268,7 @@ end
 function Base.readbytes!(http::Stream, buf::AbstractVector{UInt8},
                                        n=length(buf))
     @require n <= length(buf)
-    return http_unsafe_read(http, pointer(buf), UInt(n))
+    return GC.@preserve buf http_unsafe_read(http, pointer(buf), UInt(n))
 end
 
 function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
@@ -282,13 +282,20 @@ function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
     nothing
 end
 
-function Base.readbytes!(http::Stream, buf::IOBuffer, n=bytesavailable(http))
+@noinline bufcheck(buf, n) = ((buf.size + n) <= buf.maxsize) || throw(ArgumentError("Unable to grow response stream IOBuffer large enough for response body size"))
+
+function Base.readbytes!(http::Stream, buf::Base.GenericIOBuffer, n=bytesavailable(http))
     Base.ensureroom(buf, n)
-    unsafe_read(http, pointer(buf.data, buf.size + 1), n)
+    # check if there's enough room in buf to write n bytes
+    bufcheck(buf, n)
+    data = buf.data
+    GC.@preserve data unsafe_read(http, pointer(data, buf.size + 1), n)
     buf.size += n
 end
 
-function Base.read(http::Stream, buf::IOBuffer=PipeBuffer())
+Base.read(http::Stream, buf::Base.GenericIOBuffer=PipeBuffer()) = take!(readall(http, buf))
+
+function readall!(http::Stream, buf::Base.GenericIOBuffer=PipeBuffer())
     if ntoread(http) == unknown_length
         while !eof(http)
             readbytes!(http, buf)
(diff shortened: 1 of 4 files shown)


quinnj:
Notes to myself before I fall asleep:
  * want to probably make sure this works if user passes in a view of a byte array; I think in CloudStore.jl we want to be able to pre-allocate the full object buffer, then pass in views for individual range requests
  * need to test that the buffer isn't grown if it's too small; but what error will be thrown?
  * the # of allocations per request right now varies too much from request to request already, so it's hard to even tell if this really reduces allocations at all, but it's at least obvious that the total _bytes_ allocated is less, so it seems to at least be somewhat working!

src/Streams.jl
```diff
@@ -288,8 +288,7 @@ function Base.readbytes!(http::Stream, buf::IOBuffer, n=bytesavailable(http))
     buf.size += n
 end
 
-function Base.read(http::Stream)
-    buf = PipeBuffer()
+function Base.read(http::Stream, buf::IOBuffer=PipeBuffer())
```
Drvi, line 291: If we make buf a `Base.GenericIOBuffer` then we can pass views as well. Same applies fo `Base.readbytes!(http::Stream, buf::IOBuffer)`

Drvi:
This looks great! As per our discussion, I think it would be great if we also have a non-allocating path for GenericIOBuffers (for https://github.com/JuliaServices/CloudStore.jl/pull/24)

src/Streams.jl
```diff
@@ -282,14 +280,20 @@ function Base.unsafe_read(http::Stream, p::Ptr{UInt8}, n::UInt)
...
-    Base.ensureroom(buf, n)
-    unsafe_read(http, pointer(buf.data, buf.size + 1), n)
+@noinline bufcheck(buf, n) = ((buf.size + n) <= length(buf.data)) || throw(ArgumentError("Unable to grow response stream IOBuffer large enough for response body size"))
+
+function Base.readbytes!(http::Stream, buf::Base.GenericIOBuffer, n=bytesavailable(http))
+    Base.ensureroom(buf, buf.size + n)
+    # check if there's enough room in buf to write n bytes
+    bufcheck(buf, n)
+    data = buf.data
+    GC.@preserve data unsafe_read(http, pointer(data, buf.size + 1), n)
```
Drvi, line 290: _Sleepy drive by review alert_

Should this be `Base.ensureroom(buf, buf.ptr + n - 1)` and `pointer(data, buf.ptr)` and similarly for `bufcheck`?

quinnj: Well, perhaps, but we're updating `buf.size` ourselves (and not `buf.ptr`), so `buf.size` is correct for this case. From what I can tell, `buf.size` tracks the last byte _written_ to the IOBuffer, whereas `buf.ptr` tracks the next by to _read_ from the IOBuffer. So if that's right, then I think tracking `buf.size` is correct here.

Drvi: AFAIK, writing to IOBuffer should advance both `ptr` and `size`:
```julia
julia> io = IOBuffer(UInt8[1, 2, 3, 4, 5, 6], write=true)
IOBuffer(data=UInt8[...], readable=false, writable=true, seekable=true, append=false, size=0, maxsize=Inf, ptr=1, mark=-1)

julia> write(io, "aaaa")
4

julia> io
IOBuffer(data=UInt8[...], readable=false, writable=true, seekable=true, append=false, size=4, maxsize=Inf, ptr=5, mark=-1)

julia> seekstart(io)
IOBuffer(data=UInt8[...], readable=false, writable=true, seekable=true, append=false, size=4, maxsize=Inf, ptr=1, mark=-1)

julia> write(io, "aa")
2

julia> io
IOBuffer(data=UInt8[...], readable=false, writable=true, seekable=true, append=false, size=4, maxsize=Inf, ptr=3, mark=-1)
```
Note that seekstart will allow us reuse the parts of the buffer that were pre-allocated earlier.

src/clientlayers/StreamRequest.jl
```diff
@@ -128,10 +128,36 @@ function readbody(stream::Stream, res::Response, decompress::Union{Nothing, Bool
...
+                    # for HTTP.Stream, there's already an optimized read method
+                    # that just needs an IOBuffer to write into
+                    readall!(buf_or_stream, body)
+                else
+                    error("unreachable")
+                end
+            else
+                res.body = read(buf_or_stream)
+            end
+        elseif (res.body isa IOBuffers || res.body isa Base.GenericIOBuffer) && buf_or_stream isa Stream
```
Drvi, line 158: I think this could be simplified, since:
```julia
julia> IOBuffers <: Base.GenericIOBuffer
true
```

PR #986, opened 2023-01-08
Add content auto decoding for `download()`

### LINKED ISSUE #889: HTTP.download() does not automatically decode Content-Encoding
# Verisons
Julia v1.7.3
HTTP.jl v1.2.0
MbedTLS.jl v1.1.1

# Issue
When calling `HTTP.get()` on a resource with `Content-Encoding: gzip`, the response is automatically decoded for me:

```
julia> using HTTP
       io = IOBuffer()
       r = HTTP.get("https://httpbin.org/gzip", response_stream=io);
       println(String(take!(io)))
{
  "gzipped": true, 
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "gzip", 
    "Content-Length": "0", 
    "Host": "httpbin.org", 
    "User-Agent": "HTTP.jl/1.7.3", 
    "X-Amzn-Trace-Id": "Root=1-62d6f292-06e09add2b03b541490576d4"
  }, 
  "method": "GET", 
  "origin": "75.172.111.107"
}
```

Using `HTTP.download()`, it does not:
```
julia> using HTTP
       HTTP.download("https://httpbin.org/gzip", "/dev/stdout"; update_period=Inf);
���b�=��j�0�w?��*Œ]�:��B�R\�*[gY���|]���x����ݵ(Kb.vY@���<��͠4�5�k#���3��wd
                                                                         eJ�n��:�tj���;����9�*��_��q�c>���3q5%N}��~~��IV��7��G��F�o�|��=>s�
�QZL��u�&C����p$6�g�-��{7������K�8�W����Sj2
```

AtsushiSakai:
This PR adds content auto decoding function for `download()` using [TranscodingStreams\.jl](https://github.com/JuliaIO/TranscodingStreams.jl) to fix #889

### DIFF
--- a/src/download.jl
+++ b/src/download.jl
@@ -1,6 +1,5 @@
 using .Pairs
 using CodecZlib
-using TranscodingStreams
 
 """
     safer_joinpath(basepart, parts...)
@@ -70,12 +69,6 @@ function determine_file(path, resp, hdrs)
                     )
 
         
-        # get the extension, if we are going to save it in encoded form.
-        # unlike a web-browser we don't automatically decompress
-        if header(resp, "Content-Encoding") == "gzip"
-            filename *= ".gz"
-        end
-        
         safer_joinpath(path, filename)
     else
         # We have been given a full filepath
@@ -124,7 +117,7 @@ function download(url::AbstractString, local_path=nothing, headers=Header[]; upd
         prev_time = now()
 
         if header(resp, "Content-Encoding") == "gzip"
-            stream = TranscodingStream(GzipDecompressor(), stream) # auto decoding
+            stream = GzipDecompressorStream(stream) # auto decoding
             total_bytes = NaN # We don't know actual total bytes if the content is zipped.
         end
 
--- a/src/download.jl
+++ b/src/download.jl
@@ -1,4 +1,6 @@
 using .Pairs
+using CodecZlib
+using TranscodingStreams
 
 """
     safer_joinpath(basepart, parts...)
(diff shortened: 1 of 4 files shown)


src/download.jl
```diff
@@ -121,6 +123,11 @@ function download(url::AbstractString, local_path=nothing, headers=Header[]; upd
         start_time = now()
         prev_time = now()
 
+        if header(resp, "Content-Encoding") == "gzip"
+            stream = TranscodingStream(GzipDecompressor(), stream) # auto decoding
```
quinnj, line 127: I think this can just be:
```suggestion
            stream = GzipDecompressorStream(stream) # auto decoding
```
and then we don't need the explicit dependency on TranscodingStreams

AtsushiSakai: Thank you. I fixed.

test/runtests.jl
```diff
@@ -13,6 +13,7 @@ include(joinpath(dir, "resources/TestRequest.jl"))
             "chunking.jl",
             "utils.jl",
             "client.jl",
+            "download.jl",
```
AtsushiSakai, line 16: I found this test file is missing.

src/download.jl
```diff
@@ -68,12 +69,6 @@ function determine_file(path, resp, hdrs)
                     )
 
         
-        # get the extension, if we are going to save it in encoded form.
-        # unlike a web-browser we don't automatically decompress
-        if header(resp, "Content-Encoding") == "gzip"
-            filename *= ".gz"
-        end
-
```
AtsushiSakai, line 76: Auto decoding is added, so I think ".gz" is not needed anymore. And this change is needed to pass "test/download.jl".

quinnj (approved):

quinnj:
Thanks @AtsushiSakai!

Issue #988, opened 2023-01-11
new customizable retry logic introduced unnecessary retries and retry_check is not executed


MarkusSchildhauer:
[This commit](https://github.com/JuliaWeb/HTTP.jl/commit/02ba3454af7db22f8ae2151df066139d768af373#diff-b2c5d0a3fdbbe9ee74c7d758e6e600a881b844302b7dfe83a0d4a944caebfc81R48) aimed to provide a customizable retry logic. It can be seen in line 48, that the new `retry_check` function is only executed if `retryable(req)` returns `false`:
```
retry = retryable(req) || retryablebody(req) && _retry_check(s, ex, req, retry_check)
```
But with default arguments on `retry_delays` and `retries`, `retryable(req)` is typically `true`. 
An ugly workaround to make `retry_check` accessible is this:
```
HTTP.get(<url>, retries=1, retry_delays=ExponentialBackOff(n=100),  retry_check=(args...)->(@info(: retry_check); true))
```
I think those changes are critical since by default there are no further retry checks anymore, which causes a lot of unnecessary time-consuming retries.
Was it actually intended that neither the response status is checked with "retryable" nor all the other possibilities ("isrecoverable") that make a recovery very unlikely?

What is actually the use of the `retryattempt[]` counter, because the ExponentialBackOff iterator is already limiting the number of retries?

 - Julia 1.8.4
 - HTTP.jl 1.6.3
 - MbedTLS.jl 1.1.7

MarkusSchildhauer:
Many thanks for the quick response. But I doubt that this is all about it. If `retryablebody(req)` returns `false` `retryable(req)` returns `false`, too (see [here](https://github.com/JuliaWeb/HTTP.jl/blob/fe9bf996dd8a9e0c113e8c2819c9608013ed1d18/src/Messages.jl#L287)). So your proposed statement reduces to:
```
retry = retryablebody(req) && _retry_check(s, ex, req, retry_check)
```
Which basically means, that by default there are no retires at all, since `retry_check=FALSE`.

Please don't forget about the other subissues which are related and somehow influenced by your proposal:

1. currently unnecessary retries (or no retries at all -> your proposal) if `retry_check` is not configured manually
2. retryattempt[] counter. More precise: currently, when I set `retry_delays` with `n > 4`, I additionally have to set `retries=n`. This seems unexpected and not intentional. The proposal would solve that but the `retryattempt` counter is entirely useless than.

[...]

PR #989, opened 2023-01-11
fix for issue #988 --> seemles highly customizable retry setup

MarkusSchildhauer:
Solves issue https://github.com/JuliaWeb/HTTP.jl/issues/988 to my understanding.

features: 

- `retry_check` is always executed/checked if retryable(req) returns true.
- `retry_check` keyword arguments defaults to function `HTTP.retry_check` respectively `HTTP.isrecoverable` which restores the old behaviour that minimizes unnecessary (little promising) retries.
- `retry_delays` is not restricted to `ExponentialBackOff`, it can be any iterator with length over Reals
- `retry` defaults to `retries>0`
- `retries` is default `n` for `ExponentialBackOff`, but is overwritten by `length(retry_delays)` to avoid inconsistencies caused by `req.context[:retrylimitreached]`
- even though I don't see the use of `req.context[:retrylimitreached]` and `req.context[:retryattempt]` it's consistently integrated into the `Base.retry`-check-function.

### DIFF
--- a/src/clientlayers/RetryRequest.jl
+++ b/src/clientlayers/RetryRequest.jl
@@ -3,7 +3,7 @@ module RetryRequest
 using Sockets, LoggingExtras, MbedTLS, OpenSSL
 using ..IOExtras, ..Messages, ..Strings, ..ExceptionRequest, ..Exceptions
 
-export retrylayer, retry_check, isrecoverable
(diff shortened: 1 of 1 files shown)


quinnj:
Closing as obsolete; functionality already exists in the recent 2.0 release

PR #991, opened 2023-01-11
Cleanup header initialization to avoid excessive copies

quinnj:
Set keepalive to true by default for tcp layer.
Fix bug where both StatusCodes & Messages were exporting `statustext`.

### DIFF
--- a/src/ConnectionPool.jl
+++ b/src/ConnectionPool.jl
@@ -410,6 +410,9 @@ end
 function getconnection(::Type{TCPSocket},
                        host::AbstractString,
                        port::AbstractString;
+                       # set keepalive to true by default since it's cheap and helps keep long-running requests/responses
+                       # alive in the face of heavy workloads where Julia's task scheduler might take a while to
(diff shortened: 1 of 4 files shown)


src/ConnectionPool.jl
```diff
@@ -410,7 +410,7 @@ end
 function getconnection(::Type{TCPSocket},
                        host::AbstractString,
                        port::AbstractString;
-                       keepalive::Bool=false,
+                       keepalive::Bool=true,
```
nickrobinson251, line 413: can you add a comment on why this default to true?

nickrobinson251 (approved):

src/ConnectionPool.jl
quinnj, line 413:
```suggestion
                       # set keepalive to true by default since it's cheap and helps keep long-running requests/responses
                       # alive in the face of heavy workloads where Julia's task scheduler might take a while to
                       # keep up with midflight requests
                       keepalive::Bool=true,
```

PR #995, opened 2023-01-12
Fix mistake in default value of `readtimeout`

simsurace:
The keyword default is set to 60, as stated in the docstring of `request`, but this doc page says otherwise.


simsurace:
@quinnj, it seems that just as I was making this PR, you came out with #994.

quinnj:
Yeah, sorry for the switch and pull; thanks for the PR and helping with docs though!

PR #1000, opened 2023-01-18
Switch tests to use TestItems

nickrobinson251:
Taking TestItems.jl for a run on another repo

No idea if this is right yet, because it relies on the new `@testsetup` stuff and i wasn't sure how to get that in-development stuff locally given that the `TestItem*.jl` packages vendor their dependencies


nickrobinson251:
https://github.com/JuliaWeb/HTTP.jl/pull/1066

PR #1005, opened 2023-01-31
Properly trigger CorsMiddleware on OPTIONS method

ahjulstad:
Update example in documentation to properly trigger CorsMiddleware on OPTIONS method


quinnj (approved):
Thanks @ahjulstad!

PR #1013, opened 2023-02-14
Remove Inifile as a dep in Project.toml

sjkelly:
Seems to have been removed in: https://github.com/JuliaWeb/HTTP.jl/pull/789 and is no longer needed.


quinnj (approved):
Thanks!

PR #1016, opened 2023-02-23
Write cookie separator with space

gustafsson:
When a request includes cookies as a header (as opposed to a keyword argument) HTTP.jl strings them together, but misses a space character, causing the receiving party to parse the cookies incorrectly if it adheres to `; ` as separator.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie)
> Pairs in the list are separated by a semicolon and a space ('; ').

### DIFF
--- a/test/cookies.jl
+++ b/test/cookies.jl
@@ -58,8 +58,11 @@ using Sockets, Test
             @test HTTP.stringify("cookie-0", cookies) == "cookie-0; $expected"
             @test HTTP.stringify("cookie-0=", cookies) == "cookie-0=; $expected"
             @test HTTP.stringify("cookie-0=0", cookies) == "cookie-0=0; $expected"
-            @test HTTP.stringify("cookie-0=0;", cookies) == "cookie-0=0;; $expected"
+            @test HTTP.stringify("cookie-0=0 ", cookies) == "cookie-0=0 ; $expected"
+            @test HTTP.stringify("cookie-0=0  ", cookies) == "cookie-0=0  ; $expected"
+            @test HTTP.stringify("cookie-0=0;", cookies) == "cookie-0=0; $expected"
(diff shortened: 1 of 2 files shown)


test/cookies.jl
```diff
@@ -51,7 +51,16 @@ using Sockets, Test
...
                   ]
-        @test HTTP.stringify("", cookies) == "cookie-1=v\$1; cookie-2=v\$2; cookie-3=v\$3"
+        expected = "cookie-1=v\$1; cookie-2=v\$2; cookie-3=v\$3"
+        @test HTTP.stringify("", cookies) == expected
+
+        @testset "combine cookies with existing header" begin
+            @test HTTP.stringify("cookie-0", cookies) == "cookie-0; $expected"
+            @test HTTP.stringify("cookie-0=", cookies) == "cookie-0=; $expected"
+            @test HTTP.stringify("cookie-0=0", cookies) == "cookie-0=0; $expected"
+            @test HTTP.stringify("cookie-0=0;", cookies) == "cookie-0=0;; $expected"
```
nickrobinson251, line 61: why is `";; "` with the duplicate semi-colon the correct answer here?

gustafsson: My line of thinking was "Because `;` on its own shouldn't be interpreted as a separator, so it shouldn't be stripped from the value".

Looking at it again I realize that a single `;` isn't valid in the value anyways. I guess the most sensible approach would be to assume that it was intended as a separator.

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) says:
> A `<cookie-value>` can optionally be wrapped in double quotes and include any US-ASCII character excluding a control character, [Whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace), double quotes, comma, semicolon, and backslash

quinnj (approved):
SGTM; thanks @gustafsson!

PR #1022, opened 2023-03-14
Document verbose=-1 kwarg for HTTP.serve

fonsp:
Added in https://github.com/JuliaWeb/HTTP.jl/pull/955 . Should we add a compatibility note about HTTP >= 1.5.2?


fonsp:
@quinnj

PR #1029, opened 2023-03-29
Add AbstractDict for request bodies

screwdog:
I've broadened the accepted request body types from `Dict` to `AbstractDict` to allow for custom dictionary types.

I have a number of custom dictionary types representing JSON objects of various schema (a fairly common practice) and it is convenient to not have to explicitly convert them.

All tests pass and I don't think this will break any uses? Happy for any feedback.


quinnj:
Thanks!

PR #1035, opened 2023-04-25
Parse set-cookie expires with trailing GMT

gustafsson:
`Set-Cookie` with the `AlternateRFC1123Format` (as in `expires=Wed, 23-Nov-2011 01:05:03 GMT`) works fine.

But I couldn't see any tests for parsing `Set-Cookie` with `Dates.RFC1123Format`, i.e without the dashes as in `expires=Wed, 23 Nov 2011 01:05:03 GMT`.

`Dates.RFC1123Format` doesn't include the trailing `" G\MT"` so we need to append that to the format string.


quinnj (approved):
Thanks for this @gustafsson!

Issue #1038, opened 2023-04-27
Wrap most of the request process in a `@spawn` 


quinnj:
Our use of `@async` in a few places in the HTTP client request process means that we'll cause the calling task to be pinned to a thread while the request is happening. If we instead do an initial spawn at the start of the request process, we would then only be pinning our own HTTP parent task to a thread, possibly separate from the original calling task, which gives more freedom for that task to migrate as needed.

PR #1039, opened 2023-05-01
Remove uses of at-async in favor of at-spawn

quinnj:
https://github.com/JuliaWeb/HTTP.jl/issues/1038


quinnj:
Tests currently failing until https://github.com/JuliaServices/ConcurrentUtilities.jl/pull/22 is merged and tagged.

src/clientlayers/ExceptionRequest.jl
```diff
@@ -10,14 +10,15 @@ using ..IOExtras, ..Messages, ..Exceptions
 Throw a `StatusError` if the request returns an error response status.
 """
 function exceptionlayer(handler)
-    return function exceptions(stream; status_exception::Bool=true, logerrors::Bool=false, kw...)
-        res = handler(stream; logerrors=logerrors, kw...)
+    return function exceptions(stream; status_exception::Bool=true, timedout=nothing, logerrors::Bool=false, logtag=nothing, kw...)
```
Drvi, line 13: IIUC, `timedout` kwarg is strictly internal thing, the user shouldn't provide their own `TimedOut` struct to a `HTTP.get`? If that is the case, we might want to call it `_timedout` just to make it clearer that this is an internal thing.

src/clientlayers/ConnectionRequest.jl
```diff
@@ -126,8 +127,10 @@ function connectionlayer(handler)
                     break
                 end
             end
-            if logerrors && !(e isa StatusError || e isa TimeoutError)
-                @error "HTTP.ConnectionRequest" exception=(e, catch_backtrace()) method=req.method url=req.url context=req.context
+            root_err = e isa CapturedException ? e.ex : e
+            if logerrors && !(root_err isa StatusError || root_err isa Exceptions.TimeoutError || root_err isa Base.IOError)
```
Drvi, line 131: I'm not really sure how to tell if this list of exceptions is right... Would it make sense to only log the error if the exception is not a `CapturedException` (presumably, if it is a CapturedException, we already logged it before we wrapped it)?

Drvi (approved):
Looks good to me:) I think the `@async -> @spawn` changes are not the most complicated, it's more the switch to OpenSSL and the changes to exception handling which are more difficult to access. But since these are all based on your experiments with large workload, I believe they'll make nice improvements in terms of performance and log deduplication.

src/clientlayers/ConnectionRequest.jl
quinnj: Yeah, I like that better

PR #1043, opened 2023-05-09
Fix qualification of try_with_timeout

quinnj:


src/Connections.jl
```diff
@@ -509,7 +509,7 @@ function getconnection(::Type{TCPSocket},
                 tcp = Sockets.TCPSocket()
                 Sockets.connect!(tcp, addr, p)
                 try
-                    try_with_timeout(connect_timeout) do
+                    Exceptions.try_with_timeout(connect_timeout) do
```
mattBrzezinski, line 512: I can't mark them, but other usages of `try_with_timeout` on the commit `37c6adc` are...

* src/clientlayers/ConnectionRequest.jl: L101
* src/clientlayers/TimeoutRequest.jl: L20

quinnj: I think those are fine though because those modules don't _also_ import `ConcurrentUtilities`, so there's no ambiguity

mattBrzezinski (approved):

PR #1048, opened 2023-05-13
Also lock changes of request.context in readbody

quinnj:


msagarpatel (commented):
@quinnj I think we need to use the lock in `readbody!` 😅

quinnj:
Oh my goodness; not my day! Thanks for helping: https://github.com/JuliaWeb/HTTP.jl/pull/1049

PR #1050, opened 2023-05-15
Proxy support for HTTPS connection pool

gustafsson:
Connection pools are really effective. This PR makes them work through a proxy as well.

When establishing an HTTPS connection through a proxy a connection is first made with the proxy and then a second connection, a tunnel, is established with the target host. This second connection is kept in a pool whereas the connection with the proxy is not kept in the pool.

Later, when a subsequent connection is made with the same target host it could reuse the tunnel. Unfortunately the tunnel isn't found when looking in the pool for available connections. The reason is that the `sslupgrade` function creates a new `Connection` with a different `connectionkey`. This mismatch between keys causes  `newconnection` to not find the reusable tunnel connection, even though we readily keep the tunnel alive sitting there just waiting to be reused.

In this PR, `newtunnelconnection` uses a `connectionkey` that can be reused as expected.

Does this look like the right approach for this kind of functionality?

### DIFF
--- a/src/Connections.jl
+++ b/src/Connections.jl
@@ -615,31 +615,6 @@ function sslconnection(::Type{SSLContext}, tcp::TCPSocket, host::AbstractString;
     return io
 end
 
-function sslupgrade(::Type{IOType}, c::Connection{T},
-                    host::AbstractString;
-                    pool::Union{Nothing, Pool}=nothing,
-                    require_ssl_verification::Bool=NetworkOptions.verify_host(host, "SSL"),
-                    keepalive::Bool=true,
-                    readtimeout::Int=0,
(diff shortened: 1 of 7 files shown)


gustafsson:
(I've rebased since this PR was created a while ago)

This PR currently consists of 3 commits:

1. `test: reuse connection from pool` I was unable to find any prior references to the pool in the test suite.

2. `broken test: https pool with proxy` adds a broken test to illustrate the issue.

3. `Tunnel: newtunnelconnection` is my proposed implementation to resolve the broken test.

quinnj:
Underlying issue/functionality was factored into teh 2.0 release

Issue #1021, opened 2023-03-14
Type piracy with `Base.readuntil`


mortenpi:
https://github.com/JuliaWeb/HTTP.jl/blob/a3a576807a4ecddcefa8d76f9efd07ea2b309f8e/src/IOExtras.jl#L114-L124

This _could be_ causing some issues for us in our use case, although that needs a bit more investigation. But regardless, it might be a good idea to try to get rid of it, just in case?

mortenpi:
I noticed more type piracy in `IOExtras`:

https://github.com/JuliaWeb/HTTP.jl/blob/a3a576807a4ecddcefa8d76f9efd07ea2b309f8e/src/IOExtras.jl#L54-L84

I see it was [discussed quite extensively, but never really resolved](https://github.com/JuliaWeb/HTTP.jl/pull/745/files/bbff1de765ada6a447891f1da7ed3b801b4dfab4#diff-7c32e5796ccf6698454005ccd585bf868ab944625190c07a4d79e675fa66e847).

mortenpi:
So #1083 fixed the `readuntil` case, but the `startread` / `closeread` / `startwrite` / `closewrite` functions are still there.

https://github.com/JuliaWeb/HTTP.jl/blob/bbe1746b9e84fb3153ac951eb10ba90ef7a776d9/src/IOExtras.jl#L46-L84

Not quite sure how to handle those, since those are exported and/or sometimes extend a `Base` function.

It seems like `IOExtras` assumes that they _might_ be defined in Base, although that's only actually true for `closewrite`. And so HTTP sometimes exports `Base.closewrite` and sometimes `HTTP.IOExtras.closewrite` (depending on Julia version). And if it exports `Base.closewrite`, it does type piracy on it by defining a `Base.closewrite(::Any)` method. Changing HTTP to always export `HTTP.IOExtras.closewrite` might be breaking though, since it's also exported from Base.

The other cases are not really a problem though, and we could just remove the `isdefined` checks. I guess if we really want, we could use a `@static if hasmethod()` in `IOExtras` instead, and trust that the trivial fallback methods more or less make sense for other users of the function.

quinnj:
Resolved as of the 2.0 release; HTTP no longer overloads/pirates readuntil

PR #1052, opened 2023-05-19
Don't use `===` to compare `nobytes`

mortenpi:
This is a bit of a weird one, but in a nutshell

```
julia> HTTP.IOExtras.nobytes == view(UInt8[], 1:0)
true

julia> HTTP.IOExtras.nobytes === view(UInt8[], 1:0)
false
```

If you happen to overwrite the `Base.readuntil` method in `IOExtras`[^1], you will end up returning the "wrong" `nobytes`, and so this check yields the wrong result, and HTTP.jl will usually throw a random error (or also just hangs). Running this with HTTP 1.9.4, for example, should throw in the `HTTP.get()` call:

[^1]: Which is kind-of a valid thing to do, since it's doing type piracy. And we happen to do it because we sometimes need to load two different versions of HTTP.jl in the same Julia session.

```julia
import HTTP
module HTTPOverloads
const ByteView = typeof(view(UInt8[], 1:0))
const nobytes = view(UInt8[], 1:0)
"""
Read from an `IO` stream until `find_delimiter(bytes)` returns non-zero.
Return view of bytes up to the delimiter.
"""
function Base.readuntil(buf::IOBuffer,
                    find_delimiter::Function #= Vector{UInt8} -> Int =#
                   )::ByteView
    l = find_delimiter(view(buf.data, buf.ptr:buf.size))
    if l == 0
        return nobytes
    end
    bytes = view(buf.data, buf.ptr:buf.ptr + l - 1)
    buf.ptr += l
    return bytes
end
end
# At the time of writing at least, hitting GitHub with a GET call will consistently make
# HTTP throw HTTP.Parsers.ParseError(:INVALID_STATUS_LINE, ""). Although, hitting different
# servers leads to different errors, and sometimes also works fine.
HTTP.get("https://github.com/", verbose=true, retry=false)
```

Evaluating that `HTTPOverloads` module will lead to a method overwrite:

```
WARNING: Method definition readuntil(Base.GenericIOBuffer{Array{UInt8, 1}}, Function) in module IOExtras at /home/mortenpi/.julia/packages/HTTP/A83Es/src/IOExtras.jl:114 overwritten in module HTTPOverloads at /home/mortenpi/mwe/mwe.jl:9.
```

I would say that the correct fix here would be to get rid of the type piracy altogether (#1021), but this should fix the immediate problem (and therefore allow two different versions of HTTP to be safely loaded in the same Julia session).

I presume `nobytes` is largely just a convenience alias for an empty vector of bytes. So changing the comparison to `==` shouldn't introduce any incorrectness there. I suppose `===` might be there for performance reasons? If yes, them I am not sure if changing it to `==` affects it significantly or not.

### DIFF
--- a/src/Connections.jl
+++ b/src/Connections.jl
@@ -232,7 +232,7 @@ function Base.readuntil(c::Connection, f::F #=Vector{UInt8} -> Int=#,
     if bytesavailable(buf) == 0
         read_to_buffer(c, sizehint)
     end
-    while (bytes = readuntil(buf, f)) === nobytes
+    while (bytes = readuntil(buf, f)) == nobytes
         read_to_buffer(c, sizehint)
     end
     return bytes


mortenpi:
Although, now that I think about it some more, fixing the `===` won't actually make it safe to load a newer HTTP.jl together with an older version. So maybe we should tackle the type piracy head on? The obvious option would be to define our own `readuntil` in HTTP that can have the extra method, fall back to `Base.readuntil` for everything else with

```
readuntil(args...; kwargs...) = Base.readuntil(args...; kwargs...)
```

and then replace all the uses of `readuntil` here with `IOExtras.readuntil(...)`. Possibly using a different function name to avoid confusion and errors?

Another alternative would be to get rid of the need for that pirated method altogether. This seems like would actually be the cleanest option, but I am not sure how ingrained the assumption is that this method is available. In particular, the `.stream` field of `Stream` is parametric and may be `IOBuffer`, and so probably gets triggered here? https://github.com/JuliaWeb/HTTP.jl/blob/d8392a22a87a19c4ebdd3994291c69913ea51389/src/Streams.jl#L326

quinnj (approved):
I think this is fine; thanks.

PR #1053, opened 2023-05-19
Fix set_default_connection_limit not being respected

quinnj:
With all the default global connection pool shuffling that has happened, we missed making sure `set_default_connection_limit!` was still being respected. It was updating the global connection limit ref, but the global pools were still stuck with whatever their values were when first initialized.

This PR reinitializes the global connection pools when the default connection pool limit is changed.

### DIFF
--- a/test/client.jl
+++ b/test/client.jl
@@ -13,6 +13,7 @@ using InteractiveUtils: @which
 
 # test we can adjust default_connection_limit
 HTTP.set_default_connection_limit!(12)
+@test HTTP.Connections.TCP_POOL[].max == 12
 
 @testset "@client macro" begin
     @eval module MyClient
--- a/test/client.jl
+++ b/test/client.jl
@@ -12,8 +12,12 @@ using URIs
 using InteractiveUtils: @which
 
 # test we can adjust default_connection_limit
-HTTP.set_default_connection_limit!(12)
-@test HTTP.Connections.TCP_POOL[].max == 12
+for x in (10, 12)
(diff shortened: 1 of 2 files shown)


test/client.jl
```diff
@@ -13,6 +13,7 @@ using InteractiveUtils: @which
 
 # test we can adjust default_connection_limit
 HTTP.set_default_connection_limit!(12)
+@test HTTP.Connections.TCP_POOL[].max == 12
```
Sacha0, line 16: Possibly paranoid, but might save us at some point: Perhaps test the same expectation of the other pools?

Sacha0 (approved):
Naively LGTM, assuming no cleanup need be done after rebinding the pools? Thanks Jacob!

test/client.jl
nickrobinson251, line 16: i think ideally we'd also reset the limit a second time and check it take effect (who knows, maybe `12` is/will be the default one day), and test all the pools

how about:
```suggestion
for x in (10, 12)
    HTTP.set_default_connection_limit!(x)
    @test HTTP.Connections.TCP_POOL[].max == x
    @test HTTP.Connections.MBEDTLS_POOL[].max == x
    @test HTTP.Connections.OPENSSL_POOL[].max == x
end
```
?

nickrobinson251 (approved):

test/client.jl
quinnj: Great idea

Issue #1041, opened 2023-05-05
req.url is always URI("")


edyu:
When looking for the uri of the request :: HTTP.Request, it seems req.url is always URI(""), if I want the URI, I have to do URI(req.target).path in order to see it. It makes req.url completely useless.

quinnj:
I'm assuming you mean server-side? Yeah, the `.url` field started out as a client-side property, but never got a good treatment on the server-side. If you want to make a PR, that would be welcome! Essentially in the default stream handler (in the Handlers.jl file), right before (or after) we do `request.body = read(stream)`, we should also set the `url` field based on the example you gave above.

edyu:
I just submitted a PR and honestly it's only one line and you pretty much nailed on where to change it so feel bad for taking the credit. You can decide whether to go through the trouble of merging or you can just change it yourself. Thank you again.

quinnj:
Sorry for the slow follow up here; we've canonicalized in the 2.0 release on always having req.target and req.host, both client & server side.

PR #1055, opened 2023-05-24
Setup Request.url on the server side (issue #1041)

edyu:
#1041 describes what I need. The fact request.url exists would make the user of the package want to use it but having it always empty and that you have to call URI(request.target).path is not obvious. 
It's a one line change because I don't want to change the other gethandler functions.

### DIFF
--- a/src/Handlers.jl
+++ b/src/Handlers.jl
@@ -54,6 +54,7 @@ function streamhandler(handler)
     return function(stream::Stream)
         request::Request = stream.message
         request.body = read(stream)
+        request.url = URI(request.target)
         closeread(stream)
         request.response::Response = handler(request)
         request.response.request = request


quinnj:
Hmmm, weird that this seems to be causing downloads tests to fail? I'm not sure why that would be. I can help take a look next week, but feel free to try to dig in in the mean time.

quinnj:
Hmmmm, @oxinabox, do you happen to know if the `"http://test.greenbytes.de/tech/tc2231/inlwithasciifilenamepdf.asis"` url somehow changed what it's returning? I've honestly never looked at the downloads.jl tests in much detail, so it's curious why they're failing all the sudden.

oxinabox:
Hmm it seems that tester service is broken.
It is also not responding correctly for in FireFox.
and 
`curl http://test.greenbytes.de/tech/tc2231/inlwithasciifilenamepdf.asis -O -J` is returning the wrong filename
It is also givng `inlwithasciifilenamepdf.asis` but the whole point of that test case is that  it should give the filename `foo.pdf` because that's what the response header has.

I think their webserver might be broken.

quinnj:
Yeah, that's kind of what I suspected as well. We recently switched to use the JuliaHub-hosted httpbin; I wonder if we could utilize that for these tests as well (they're still _occasionally_ flaky, but much more robust than any other test service we've ever used)

oxinabox:
The only test cases we use are:
 - http://test.greenbytes.de/tech/tc2231/#inlwithasciifilenamepdf
 - http://test.greenbytes.de/tech/tc2231/#attonlyquoted
 - http://test.greenbytes.de/tech/tc2231/#attwithasciifnescapedquote
 
 Those links fully describe them.
 They require the webserver to set some fields in the response
 
Which HTTP-Bin does not have examples of.
We could possibly put together our own little example of this, and run it on a JuliaHub hosted webserver.
Or find something reliable the happens to have these properties (maybe AWS S3?)

edyu:
Yeah, some of these errors are over my head tbh. Thanks for looking into the errors but not sure what I can do here.

schlichtanders:
Any progress here?

Run again into this missing feature (second time at least - it is so intuitive to think that `request.url` simply has the url)

oxinabox:
What needs to happen i think is:

1. This PR needs to be rebased
2. A seperate PR needs to be made to mark those old tests broken and leave a not explaining why. (if they are still broken)
3. This PR needs to be rebased onto of that one after that is merged.

quinnj:
We've canonicalized in the 2.0 release, for both client and server side, on having the request.host and request.target be the official ways to get url info

PR #1066, opened 2023-06-22
Switch tests to use ReTestItems

nickrobinson251:


nickrobinson251:
can't do this til we drop support for v1.6 -- but will leave this here til then

PR #1068, opened 2023-06-26
Attempt parallel connections to all addrs when connecting

quinnj:
Part of the "happy eyeballs" recommendations.

Note I attempted to preserve the exception throwing behavior we had previously. i.e. if I'm the last task running and errors, I'll close the `Channel{TCPSocket}` with my exception which will then propagate up. We could possibly accumulate all the exceptions into a CompositeException, but meh, I'm not sure if that adds much value.

In addition to the new parallel connecting, we're also adjusting how the `connect_timeout` is implemented. Instead of only applying to the tcp connection, we now wrap the entire `newconnection` call, which means any ssl handshake timing will also count towards the timeout. We saw a case at RelationalAI where we had a reasonable connect_timeout (10s), yet then saw long requests (>50s) where all the time was reported in the connection layer. It would seem to suggest that the ssl layer is somehow getting stuck or slow, so the proposal here is that if the ssl operations also count, then we'll more aggressively cancel/restart the connection process in the case of slow ssl handshaking.

### DIFF
--- a/src/Connections.jl
+++ b/src/Connections.jl
@@ -501,40 +501,29 @@ function getconnection(::Type{TCPSocket},
                        # alive in the face of heavy workloads where Julia's task scheduler might take a while to
                        # keep up with midflight requests
                        keepalive::Bool=true,
-                       connect_timeout::Int=10,
                        readtimeout::Int=0,
                        kw...)::TCPSocket
 
     p::UInt = isempty(port) ? UInt(80) : parse(UInt, port)
     @debugv 2 "TCP connect: $host:$p..."
     addrs = Sockets.getalladdrinfo(host)
-    connect_timeout = connect_timeout == 0 && readtimeout > 0 ? readtimeout : connect_timeout
-    lasterr = ErrorException("unknown connection error")
+    ch = Channel{TCPSocket}(1)
+    n = Ref(length(addrs))
     for addr in addrs
-        try
-            if connect_timeout > 0
-                tcp = Sockets.TCPSocket()
-                Sockets.connect!(tcp, addr, p)
-                try
-                    try_with_timeout(connect_timeout) do _
-                        Sockets.wait_connected(tcp)
-                        keepalive && keepalive!(tcp)
-                    end
-                catch
-                    close(tcp)
-                    rethrow()
-                end
-            else
-                tcp = Sockets.connect(addr, p)
+        Threads.@spawn begin
+            _addr = $addr
+            try
+                tcp = Sockets.connect($addr, p)
                 keepalive && keepalive!(tcp)
+                put!(ch, tcp)
+            catch e
+                @lock ch begin
+                    (n[] -= 1) == 0 && close(ch, e)
+                end
             end
-            return tcp
-        catch e
-            lasterr = e isa ConcurrentUtilities.TimeoutException ? ConnectTimeout(host, port) : e
         end
     end
-    # If no connetion could be set up, to any address, throw last error
-    throw(lasterr)
+    return take!(ch)
 end
 
 const nosslconfig = SSLConfig()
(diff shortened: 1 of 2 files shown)


src/Connections.jl
```diff
@@ -501,40 +501,29 @@ function getconnection(::Type{TCPSocket},
...
-                        keepalive && keepalive!(tcp)
-                    end
-                catch
-                    close(tcp)
-                    rethrow()
-                end
-            else
-                tcp = Sockets.connect(addr, p)
+        Threads.@spawn begin
+            _addr = $addr
```
Drvi, line 514: It seems `_addr` is not used here, is that intended?

src/Connections.jl
```diff
@@ -501,40 +501,29 @@ function getconnection(::Type{TCPSocket},
...
+            try
+                tcp = Sockets.connect($addr, p)
                 keepalive && keepalive!(tcp)
+                put!(ch, tcp)
+            catch e
+                @lock ch begin
+                    (n[] -= 1) == 0 && close(ch, e)
```
Drvi, line 521: Ah, so if all connection attempts fail, this would make the `take!` throw. Maybe this could be a comment?

src/Connections.jl
```diff
@@ -501,40 +501,29 @@ function getconnection(::Type{TCPSocket},
...
+                end
             end
-            return tcp
-        catch e
-            lasterr = e isa ConcurrentUtilities.TimeoutException ? ConnectTimeout(host, port) : e
         end
     end
-    # If no connetion could be set up, to any address, throw last error
-    throw(lasterr)
+    return take!(ch)
```
Drvi, line 526: Question: Once we leave the scope, there might be a bunch of tasks waiting to `put!` on a Channel -- would those be GC'd, or do we need to close Channel so they fail?

src/Connections.jl
```diff
@@ -501,40 +501,29 @@ function getconnection(::Type{TCPSocket},
...
-                        Sockets.wait_connected(tcp)
-                        keepalive && keepalive!(tcp)
-                    end
-                catch
-                    close(tcp)
-                    rethrow()
-                end
-            else
-                tcp = Sockets.connect(addr, p)
+        Threads.@spawn begin
```
Drvi, line 513: I think ideally we'd stop spawning tasks once we successfully establish a connection.

src/clientlayers/ConnectionRequest.jl
```diff
@@ -73,10 +73,17 @@ function connectionlayer(handler)
             url = target_url = req.url
         end
 
+        connect_timeout = connect_timeout == 0 && readtimeout > 0 ? readtimeout : connect_timeout
         IOType = sockettype(url, socket_type, socket_type_tls)
         start_time = time()
         try
-            io = newconnection(IOType, url.host, url.port; readtimeout=readtimeout, kw...)
+            io = if connect_timeout > 0
+                try_with_timeout(connect_timeout) do _
```
Drvi, line 81: Are we potentially using 2 Timers on a single `newconnection`? I think `newconnection` can call `getconnection` which also does `try_with_timeout` and has a default value `connect_timeout::Int=10`.

Drvi (commented):
Left a couple of notes, but I think the overall idea is great:+1:. There are also some CI failures we might need to look into.

src/Connections.jl
quinnj: Good catch

src/Connections.jl
quinnj: Pushing a fix for this

src/Connections.jl
quinnj: Pushed a fix for this (we now do multiple checks and if the channel is "done" (i.e. already put!), then we stop any further operations)

src/Connections.jl
quinnj: Good call; I'll add a comment

quinnj: pushed

Issue #1067, opened 2023-06-25
WebSocket client hiding errors thrown from within do block


bryaan:
The websocket client is catching errors thrown from within the do block.  The outer try catch never gets triggered and the server restarts unexpectedly.  The only hack is to use a condition variable.  This adds even more boilerplate.  Can we fix this?

```
while true
    try
        WebSockets.open("wss://ws-feed.exchange.coinbase.com") do ws
            try
                error("test")

                for msg in ws
                end
            catch e
                @info "Inner try catch"
                rethrow(e)
            end
        end
    catch e
        @info "Outer try catch"
        rethrow(e)
    end
    @info "Restarting"
    sleep(2)
end
```

quinnj:
This is fixed in the 2.0 release

PR #1069, opened 2023-06-29
Rethrow errors so they may be handled outside of WebSocket.open() 

bryaan:
This is a fix for a bug where any errors throw from within WebSocket.open() is hidden from an outer try/catch.

https://github.com/JuliaWeb/HTTP.jl/issues/1067

### DIFF
--- a/src/WebSockets.jl
+++ b/src/WebSockets.jl
@@ -384,6 +384,7 @@ function open(f::Function, url; suppress_close_error::Bool=false, verbose=false,
                     close(ws, CloseFrameBody(1008, "Unexpected client websocket error"))
                 end
             end
+            rethrow(e)
         finally
             if !isclosed(ws)
                 close(ws, CloseFrameBody(1000, ""))


src/WebSockets.jl
```diff
@@ -384,6 +384,7 @@ function open(f::Function, url; suppress_close_error::Bool=false, verbose=false,
                     close(ws, CloseFrameBody(1008, "Unexpected client websocket error"))
                 end
             end
+            rethrow(e)
```
nickrobinson251, line 387: is `rethrow()` fine here? 
```suggestion
            rethrow()
```
I think `rethrow(e)` is not usually the right choice, per the docs:

```julia
help?> rethrow
search: rethrow

  rethrow()

  Rethrow the current exception from within a catch block. The rethrown exception will continue propagation as if it had not been caught.

  │ Note
  │
  │  The alternative form rethrow(e) allows you to associate an alternative exception object e with the current backtrace. However this
  │  misrepresents the program state at the time of the error so you're encouraged to instead throw a new exception using throw(e). In Julia
  │  1.1 and above, using throw(e) will preserve the root cause exception on the stack, as described in current_exceptions.

```

quinnj: I think we should actually do the rethrowing from inside the `!isok` block above, since we're currently "throwing" a "closed normally" exception when we get a close notification from the peer socket.

quinnj:
This was fixed/incorporated in the websocket re-write in the 2.0 release

PR #1072, opened 2023-07-06
Updated: Rethrow errors so they may be handled outside of WebSocket.open()

Zentrik:
Addresses comments in #1069


Zentrik:
bump

gustafsson:
This change makes sense to me. If I would like to silently ignore exceptions I could explicitly wrap my calls in `try ... catch end`.

quinnj (approved):

PR #1075, opened 2023-07-11
Ensure connect_timeout doesn't count acquire time on the connection pool

quinnj:


Drvi (approved):
LGTM, but we should add a test in a followup

PR #1083, opened 2023-07-28
Remove `readuntil` type piracy

mortenpi:
This makes sure that the `readuntil` function used by HTTP is owned by the `IOExtras` module, and so that it would be safe to add the currently type-pirated method taking an `IOBuffer`. Towards #1021.

I am not entirely sure what the best approach here is in terms of code maintainability though. Currently, I opted to keep the name the same (`readuntil`) and export it from `IOExtras`, so any module doing `using IOExtras` and using an unqualified `readuntil` will error (as the same name is exported from `Base`).

### DIFF
--- a/src/Connections.jl
+++ b/src/Connections.jl
@@ -233,13 +233,13 @@ end
 Read until `find_delimiter(bytes)` returns non-zero.
 Return view of bytes up to the delimiter.
 """
-function Base.readuntil(c::Connection, f::F #=Vector{UInt8} -> Int=#,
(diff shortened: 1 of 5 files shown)


quinnj (approved):
LGTM; I retried the failing tests, not sure if the failures are due to changes here or not.

mortenpi:
It looks like the invalidations test failure is unrelated?

quinnj:
All looks good; thanks @mortenpi!

Issue #1085, opened 2023-08-01
`isrecoverable` check for DNSErrors seems ineffective and needing correction


tanmaykm:
There are some `isrecoverable` checks in the HTTP client retry layer [here](https://github.com/JuliaWeb/HTTP.jl/blob/10182c0380e3c370a788cb48efafdf3fff4bbd43/src/clientlayers/RetryRequest.jl#L79-L83). The check that makes `Sockets.DNSError` due to `EAI_AGAIN` non-recoverable seems surprising, and is probably broken.

It checks the error code to match the declared `const EAI_AGAIN = 2`. However the codes can only be one of [these](https://github.com/JuliaLang/julia/blob/ec8df3da3597d0acd503ff85ac84a5f8f73f625b/stdlib/Sockets/src/addrinfo.jl#L108-L112), and the closest matching seems to be `UV_EAI_AGAIN` which has a value of `-3001`

```julia
julia> Base.UV_EAI_AGAIN
-3001
```

Which implies that the `isrecoverable` check for `EAI_AGAIN` DNS errors are in-effective now, and HTTP client considers all kinds of DNS errors recoverable. Unless there are any complains regarding this behavior, we could probably remove the specific `EAI_AGAIN` check then?

Or should HTTP client rather not be retrying on certain DNS errors? From the [list of possible DNS errors](https://github.com/JuliaLang/julia/blob/ec8df3da3597d0acd503ff85ac84a5f8f73f625b/stdlib/Sockets/src/addrinfo.jl#L108-L112) actually only `EAI_AGAIN` seems recoverable. So should the check be corrected to check against `UV_EAI_AGAIN` and make that recoverable, and all others non-recoverable?

PR #1088, opened 2023-08-02
mark recoverability of DNS erros correctly

tanmaykm:
This changes HTTP client to not retry on certain DNS errors. From the [list of possible DNS errors](https://github.com/JuliaLang/julia/blob/ec8df3da3597d0acd503ff85ac84a5f8f73f625b/stdlib/Sockets/src/addrinfo.jl#L108-L112) only EAI_AGAIN may be recoverable. This also changes the error code comparison to use `UV_EAI_AGAIN`, because that is what a [DNSError instance would contain](https://github.com/JuliaLang/julia/blob/ec8df3da3597d0acd503ff85ac84a5f8f73f625b/stdlib/Sockets/src/addrinfo.jl#L108-L112).

fixes: #1085

### DIFF
--- a/src/clientlayers/RetryRequest.jl
+++ b/src/clientlayers/RetryRequest.jl
@@ -17,7 +17,7 @@ increasing delay is introduced between attempts to avoid exacerbating network
 congestion.
 
 By default, requests that have a retryable body, where the request wasn't written
-or is idempotent will be retries. If the request is made and a response is received
+or is idempotent will be retried. If the request is made and a response is received
 with a status code of 403, 408, 409, 429, or 5xx, the request will be retried.
 
 `retries` controls the # of total retries that will be attempted.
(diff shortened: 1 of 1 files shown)


quinnj (approved):

nickrobinson251:
can we add a test please?

tanmaykm:
We had faced this in a likely stressed condition in our application. I was testing this with a modified julia binary to artificially throw the DNS error. I found it hard to replicate the error otherwise.

I could add a unit test for `isrecoverable`, maybe that would be useful?

tanmaykm:
@nickrobinson251  Actually I think I can add some relevant tests. And I also missed out some more changes to `isrecoverable`. PR in a bit.

nickrobinson251:
> I think I can add some relevant tests

Thanks!

> a unit test for isrecoverable, maybe that would be useful?

yeah, i think unit tests would be helpful

PR #1091, opened 2023-08-16
Test that `connect_timeout` does not include the time needed to acqui…

Drvi:
…re a connection from the pool


quinnj (approved):

NHDaly (commented):
👏 awesome, thanks!

PR #1108, opened 2023-09-07
tests: Use ARGS to choose which tests to run

cmcaine:
ARGS can be specified like this:

	Pkg.test("HTTP"; test_args=`ascii.jl parser.jl`)
	# or
	Pkg.test("HTTP"; test_args=["ascii.jl", "parser.jl"])

This is useful because the whole test suite takes ages to run.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,12 +1,11 @@
 using Test, HTTP, JSON
 
-const dir = joinpath(dirname(pathof(HTTP)), "..", "test")
-
+# See https://httpbingo.julialang.org/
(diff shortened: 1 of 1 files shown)


cmcaine:
Bump.

test/runtests.jl
```diff
@@ -1,40 +1,40 @@
 using Test, HTTP, JSON
 
-const dir = joinpath(dirname(pathof(HTTP)), "..", "test")
```
quinnj, line 3: can we keep this please? I don't mind moving it down to the for loop, but it's really handy because I often have to step through the test files one by one and `@__DIR__` doesnt' work when run at the REPL.

cmcaine: Sure!

quinnj (approved):

PR #1110, opened 2023-09-14
Don't retry on internal exceptions

Drvi:
Previously, we'd retry on exceptions happening inside layers, e.g. when cloudsign layer in CloudStore.jl decodes base64 encoded SAS tokens and if those are malformed we could get an `ArgumentException`

### DIFF
--- a/src/clientlayers/RetryRequest.jl
+++ b/src/clientlayers/RetryRequest.jl
@@ -1,6 +1,6 @@
 module RetryRequest
 
-using Sockets, LoggingExtras, MbedTLS, OpenSSL
+using Sockets, LoggingExtras, MbedTLS, OpenSSL, ExceptionUnwrapping
 using ..IOExtras, ..Messages, ..Strings, ..ExceptionRequest, ..Exceptions
 
 export retrylayer
@@ -67,7 +67,7 @@ function retrylayer(handler)
                         mark(req.body)
                     end
                 else
-                    @debugv 1 "🚷  No Retry: $(no_retry_reason(ex, req))"
+                    @debugv 1 "🚷  No Retry: $(no_retry_reason(unwrapped_ex, req))"
                 end
                 return s, retry
             end
@@ -76,12 +76,10 @@ function retrylayer(handler)
     end
 end
 
-isrecoverable(ex) = true
-isrecoverable(ex::ArgumentError) = false
-isrecoverable(ex::ErrorException) = false
-isrecoverable(ex::RequestError) = isrecoverable(ex.error)
-isrecoverable(ex::CapturedException) = isrecoverable(ex.ex)
-isrecoverable(ex::ConnectError) = isrecoverable(ex.error)
+isrecoverable(ex) = is_wrapped_exception(ex) ? isrecoverable(unwrap_exception(ex)) : false
+isrecoverable(::Union{Base.EOFError, Base.IOError, MbedTLS.MbedException, OpenSSL.OpenSSLError}) = true
+isrecoverable(ex::ArgumentError) = ex.msg == "stream is closed or unusable"
+isrecoverable(ex::CompositeException) = all(isrecoverable, ex.exceptions)
 # Treat all DNS errors except `EAI_AGAIN`` as non-recoverable
 # Ref: https://github.com/JuliaLang/julia/blob/ec8df3da3597d0acd503ff85ac84a5f8f73f625b/stdlib/Sockets/src/addrinfo.jl#L108-L112
 isrecoverable(ex::Sockets.DNSError) = (ex.code == Base.UV_EAI_AGAIN)
@@ -95,9 +93,11 @@ end
 
 function no_retry_reason(ex, req)
     buf = IOBuffer()
+    unwrapped_ex = unwrap_exception(ex)
     show(IOContext(buf, :compact => true), req)
     print(buf, ", ",
-        ex isa StatusError ? "HTTP $(ex.status): " :
+        unwrapped_ex isa StatusError ? "HTTP $(ex.status): " :
+        !isrecoverable(unwrapped_ex) ? "unrecoverable exception: " :
         !isbytes(req.body) ? "request streamed, " : "",
         !isbytes(req.response.body) ? "response streamed, " : "",
         !isidempotent(req) ? "$(req.method) non-idempotent" : "")
--- a/src/clientlayers/RetryRequest.jl
+++ b/src/clientlayers/RetryRequest.jl
@@ -77,6 +77,9 @@ function retrylayer(handler)
 end
 
 isrecoverable(ex) = true
+isrecoverable(ex::ArgumentError) = false
+isrecoverable(ex::ErrorException) = false
+isrecoverable(ex::RequestError) = isrecoverable(ex.error)
 isrecoverable(ex::CapturedException) = isrecoverable(ex.ex)
 isrecoverable(ex::ConnectError) = isrecoverable(ex.error)
 # Treat all DNS errors except `EAI_AGAIN`` as non-recoverable
--- a/src/clientlayers/RetryRequest.jl
+++ b/src/clientlayers/RetryRequest.jl
@@ -67,7 +67,7 @@ function retrylayer(handler)
                         mark(req.body)
                     end
                 else
-                    @debugv 1 "🚷  No Retry: $(no_retry_reason(unwrapped_ex, req))"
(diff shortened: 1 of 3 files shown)


src/clientlayers/RetryRequest.jl
```diff
@@ -77,6 +77,9 @@ function retrylayer(handler)
 end
 
 isrecoverable(ex) = true
+isrecoverable(ex::ArgumentError) = false
```
nickrobinson251, line 80: is `isrecoverable(ex) = true` just the wrong default?

e.g. what if CloudBase stops throwing a `ArgumentError` and decides to throw a `MalformedTokenError`?

nickrobinson251 (commented):
can you explain why this is the right solution?

src/clientlayers/RetryRequest.jl
Drvi: >can you explain why this is the right solution?

This seemed like the least disruptive fix for the problem we've seen. Maybe the right solution is to formalize the contract a `layer` needs to fulfill wrt throwing exceptions.

>is isrecoverable(ex) = true just the wrong default?

I think Jacob planned a bigger refactor for retries, switching exception retries from opt-out to opt-in would make sense to me.

> what if CloudBase stops throwing a ArgumentError and decides to throw a MalformedTokenError?

I understand, but I'm less worried about this possibility since this is the first instance of this issue we've seen and I don't expect the underlying packages to make such changes without us noticing.

nickrobinson251: i think this bug (i.e. we'd retry on exceptions happening inside layers) shows that `isrecoverable(ex) = true` is the wrong default. What would be the disruption of flipping that to `isrecoverable(ex) = false`? 

> I don't expect the underlying packages to make such changes without us noticing

packages like CloudBase.jl are _users_ on HTTP.jl (not dependencies), and we don't have much visibility on users, so i am worried about other users running into this same issue (just instead of via an ArgumentError via a MethodError or a AssertionError or a BoundError or ...)

nickrobinson251: If @quinnj is not around, perhaps we can get a third view from @NHDaly (and/or whoever was running into this issue)?

NHDaly: Looks like Blame shows this as the PR that set that as the default: https://github.com/JuliaWeb/HTTP.jl/pull/990

Looks like maybe the PR was trying to change to _always_ retry if the request is retryable (has a retryable body or is idempotent)? But yeah... it does seem like a strange default...

Do you have any memory of why that PR was made, @nickrobinson251?

nickrobinson251: i think that PR was a fix for some oversight in the preceding refactor: https://github.com/JuliaWeb/HTTP.jl/pull/974/files

I don't believe that PR intended to change any behaviour, only increase flexibility. The description (from the previous draft of the same thing https://github.com/JuliaWeb/HTTP.jl/pull/877) was
> this PR proposes **more configurability for retry logic** in the hopes that users won't opt to roll their own incorrect retry logic.

and i notice that before that refactor [we had
](https://github.com/JuliaWeb/HTTP.jl/pull/974/files#diff-b2c5d0a3fdbbe9ee74c7d758e6e600a881b844302b7dfe83a0d4a944caebfc81L66)
```julia
isrecoverable(e) = false 
isrecoverable(e::Union{Base.EOFError, Base.IOError, MbedTLS.MbedException, OpenSSL.OpenSSLError}) = true
isrecoverable(e::ArgumentError) = e.msg == "stream is closed or unusable"
isrecoverable(e::Sockets.DNSError) = true
isrecoverable(e::ConnectError) = true
isrecoverable(e::RequestError) = isrecoverable(e.error)
isrecoverable(e::StatusError) = retryable(e.status)
```
which makes me think the default of `true` was unintended

Drvi: Amusingly, there is now a failing test:

```julia
@test HTTP.RetryRequest.isrecoverable(ErrorException(""))
```

so my attempt to make an improvement with as little behavior change as possible was doomed to fail from the beginning.

Drvi: >What would be the disruption of flipping that to isrecoverable(ex) = false?

The fact that we wouldn't retry because we relied on certain exceptions being considered recoverable implicitly. Even this current PR seem to have missed `isrecoverable(e::ArgumentError) = e.msg == "stream is closed or unusable"` which is unfortunate.

Drvi:
Here are mentiones of isrecoverable `git log -pS isrecoverable > isrecoverable.txt`
[isrecoverable.txt](https://github.com/JuliaWeb/HTTP.jl/files/12610969/isrecoverable.txt)

Features an amazing docstring ascii art for `stack`, where did it go? Would be great for extended help.

Drvi:
Here is a short history of `isrecoverable`:
```julia
isrecoverable(e) = false
isrecoverable(e::Union{Base.EOFError, Base.IOError, MbedTLS.MbedException, OpenSSL.OpenSSLError}) = true
isrecoverable(e::ArgumentError) = e.msg == "stream is closed or unusable"
isrecoverable(e::Sockets.DNSError) = true
isrecoverable(e::ConnectError) = true
isrecoverable(e::RequestError) = isrecoverable(e.error)
isrecoverable(e::StatusError) = retryable(e.status)
```
->
```julia
# yes, nothing -- this effectively meant `isrecoverable(ex) = true`
```
->
```julia
isrecoverable(ex) = true
isrecoverable(ex::StatusError) = retryable(ex.status)
```
->
```julia
isrecoverable(ex) = true
isrecoverable(ex::ConnectError) = ex.error isa Sockets.DNSError && ex.error.code == EAI_AGAIN ? false : true
isrecoverable(ex::StatusError) = retryable(ex.status)
```
->
```julia
isrecoverable(ex) = true
isrecoverable(ex::CapturedException) = isrecoverable(ex.ex)
isrecoverable(ex::ConnectError) = ex.error isa Sockets.DNSError && ex.error.code == EAI_AGAIN ? false : true
isrecoverable(ex::StatusError) = retryable(ex.status)
```
->
```julia
isrecoverable(ex) = true
isrecoverable(ex::CapturedException) = isrecoverable(ex.ex)
isrecoverable(ex::ConnectError) = isrecoverable(ex.error)
isrecoverable(ex::Sockets.DNSError) = (ex.code == Base.UV_EAI_AGAIN)
isrecoverable(ex::StatusError) = retryable(ex.status)
```

From that it seems to me, that default to true _was_ deliberate -- without it no common retriable exceptions would've been retried.

quinnj:
Just a quick chime-in that the discussion all sounds great and correct. We used to have default `false` w/ a bunch of opt-ins, then in a large refactor where we had a bunch of cases that _weren't_ being correctly retried, we changed the default to `true`. I think it's fine to go back to default `false`, but I would just want to make sure we're handling cases like `CompositeException`, `CapturedException`, etc. i.e. we probably want to hook into ExceptionUnwrapping _first_ and then pass the root error to a dispatchable check for whether it's recoverable. We may also want to consider making `isrecoverable` part of the HTTP public API so that if, for example, CloudBase _does_ have a certain kind of error that _should_ be retried, it can _opt-in_ a custom exception type to being retryable.

Drvi:
Thanks Jacob! Good points, I totally forgot about composite exceptions and ExceptionUnwrapping.

>We may also want to consider making isrecoverable part of the HTTP public

I agree we want to do that eventually. Earlier I suggested that we might want to formalize what the layer can expect to be retried. Maybe via making it the layers job to wrap retryable exception in an `RetryableException` or something. This way the layer can signal that some `ArgumentErrors` are okay to retry and some are not. An alternative is to force the layers to always use custom types for exceptions and overload isrecoverable for them.

src/clientlayers/RetryRequest.jl
```diff
@@ -92,9 +93,11 @@ end
 
 function no_retry_reason(ex, req)
     buf = IOBuffer()
+    unwrapped_ex = unwrap_exception(ex)
```
quinnj, line 96: I don't think we need this `unwrap_exception` if we're doing it via dispatch in `isrecoverable`; so I think we remove it here and keep it above.

quinnj: Oh, but this is just for the `no_retry_reson`, so maybe we just need this for printing, which is fine.

Drvi: Exactly yeah, this is for printing. Thanks!

PR #1114, opened 2023-09-27
Retry on `TimeoutException`, as we did in HTTP v1.9

Drvi:
In https://github.com/JuliaWeb/HTTP.jl/pull/1110 we switched `isrecoverable(::Exception)` to be opt-in instead of opt-out, and we tried to include the exceptions we historically retried in older versions of HTTP. We didn't include `TimeoutExceptions` in that PR, but given some of the connection timeouts we've encountered in production, I think we want to retry those as well. 

What do others think? 

In the future, ideally, we'd have more granular control over retries to make this a simple runtime switch.

### DIFF
--- a/test/client.jl
+++ b/test/client.jl
@@ -707,7 +707,7 @@ end
     try
         HTTP.pushlayer!(test_context_layer)
         # 10.0.0.0 is non-routeable and will result in a connection timeout
-        HTTP.get("http://10.0.0.0", connect_timeout=1, retries=3, retry_delays=[0.1, 0.1, 0.1], test_context=test_context)
+        HTTP.get("http://example.com:81", connect_timeout=1, retries=3, retry_delays=[0.1, 0.1, 0.1], test_context=test_context)
     catch e
         @assert e isa HTTP.ConnectError
         @test e.error isa ConcurrentUtilities.TimeoutException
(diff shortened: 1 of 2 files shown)


test/client.jl
```diff
@@ -692,6 +692,35 @@ end
...
+            merge!(test_context, req.context)
+            return handler(req; kw...)
+        end
+    end
+
+    test_context = Dict{Symbol,Any}()
+    try
+        HTTP.pushlayer!(test_context_layer)
+        # 10.0.0.0 is non-routeable and will result in a connection timeout
+        HTTP.get("http://10.0.0.0", connect_timeout=1, retries=3, retry_delays=[0.1, 0.1, 0.1], test_context=test_context)
```
Drvi, line 710: Hmm, this doesn't quite work in CI

gustafsson:
I've often seen retried requests succeed after initial timeouts. Including it by default as `isrecoverable` makes sense to me.

Making isrecoverable part of the public API and mentioning it in the docs around [Request retry arguments](https://github.com/JuliaWeb/HTTP.jl/blob/8f9eb57cd9a04fd0ff24634c4504c61fa2f0e511/src/HTTP.jl#L172) would also make sense with or without this change.

Could `retry_check` provide such granular control when isrecoverable is false? What would a more simple runtime switch look like?

quinnj:
Closing as stale

PR #1116, opened 2023-10-03
Add function to report metrics for a HTTP.Connection.Pool

nickrobinson251:
- requires https://github.com/JuliaServices/ConcurrentUtilities.jl/pull/29


Manifest.toml
```diff
@@ -0,0 +1,158 @@
+# This file is machine-generated - editing it directly is not advised
```
NHDaly, line 1: Did you mean to commit the manifest? That doesn't seem right. Drop it?

nickrobinson251: yes, just temporarily, so we can run the tests with https://github.com/JuliaServices/ConcurrentUtilities.jl/pull/29 (before that's released)

will drop before merge

quinnj:
Closing as obsolete as of the 2.0 release, but give me a shout if these kinds of metrics would still be useful; it would be pretty easy to add publicly documented accessors for this in the 2.0 machinery

PR #1119, opened 2023-10-10
Fix Handlers.getcookies erroring

nguiard:
Hi!

Presently if you do HTTP.getcookies(req), you get:
`ERROR: MethodError: no method matching get(::Pair{Tuple{}, Vector{HTTP.Cookies.Cookie}}, ::Dict{Symbol, Any}, ::Symbol)`

This is because the definition of getcookies contained `() => Cookie[]` instead of `() -> Cookie[]`.

This PR fixes the error and proposes a basic test to verify that we can call `getcookies` on a request without an error.


nickrobinson251 (approved):
thanks!

Issue #1125, opened 2023-11-20
Mask Authorization header when showing RequestError


nkottary:
- Julia 1.9.3
- HTTP.jl 1.10.0
- MbedTLS.jl 1.1.7

*The problem:* The value of the Authorization header is printed as is when displaying a RequestError. This leaks the token in the key to the logs.

Steps to reproduce:
1) The following HTTP server has an error in it, so it will result in an eof error for clients:

```server.jl
using HTTP

# start a blocking server
HTTP.listen() do http::HTTP.Stream
    @show http.message
    @show HTTP.header(http, "Content-Type")
    while !eof(http)
        println("body data: ", String(readavailable(http)))
    end
    HTTP.setstatus(http, 404)
    HTTP.setheader(http, "Foo-Header" => "bar")
    HTTP.startwrite(http)
    HTTP.close!(http)    # Error
    return
    write(http, "response body")
    write(http, "more response body")
end
```

Run this as a server: `julia server.jl`

2) Send a request to this server with an Authorization header:

```
using HTTP
using JSON

function run_get()
    headers = [
        "Content-Type" => "application/json",
        "Authorization" => "Basic LeakedOhNo",
        "Accept" => "*/*",
    ]

    body = JSON.json(
        Dict(
            "values" => ["a", "b"]
        )
    )
    resp = HTTP.post(
        "http://localhost:8081/";
        headers=headers,
        body=body,
    )
end
```

Example:
```
$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.3 (2023-08-24)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> include("client.jl")
run_get (generic function with 1 method)

julia> run_get()
```

3) This results in error with Authorization header revealed in stacktrace:

```
ERROR: HTTP.RequestError:
HTTP.Request:
HTTP.Messages.Request:
"""
POST / HTTP/1.1
Content-Type: application/json
Authorization: Basic LeakedOhNo
Accept: */*
Host: localhost:8081
User-Agent: HTTP.jl/1.9.3
Content-Length: 20
Accept-Encoding: gzip

{"values":["a","b"]}"""Underlying error:
[...]

PR #1127, opened 2023-11-23
Hide Authorization header value

nkottary:
Edits the `show` and `writeheaders` methods to hide the value of Authorization header. This hides the Authorization header value when `Request` or `RequestError` objects are `show`n.

Fix for https://github.com/JuliaWeb/HTTP.jl/issues/1125


src/Messages.jl
```diff
@@ -609,7 +620,7 @@ function Base.show(io::IO, m::Message)
     end
     println(io, typeof(m), ":")
     println(io, "\"\"\"")
-    writeheaders(io, m)
+    writeheaders(io, m; mask_authorization_header=true)
```
fredrikekre, line 623: Perhaps it is better to not modify the code of `writeheaders` and instead handle everything locally in this show method? Something like
```julia
iob = IOBuffer()
writeheaders(iob, m)
header_str = String(take!(iob))
header_str = replace(header_str, r"(*CRLF)^(Authorization: ).*$"m => s"\1******")
write(io, header_str)
```

Are there any other headers that may contain sensitive information that we should consider scrubbing as well?

pfitzseb: Agreed. I don't think we should touch `writeheaders` for a change to the `show` method.

> Are there any other headers that may contain sensitive information that we should consider scrubbing as well?

`Cookie`/`Set-Cookie`.

nkottary: Fixed in https://github.com/JuliaWeb/HTTP.jl/pull/1127/commits/1dd214ba07c93095a8d147c87a12680974c36e34

src/Messages.jl
```diff
@@ -609,7 +609,23 @@ function Base.show(io::IO, m::Message)
...
+    # Mask sensitive header values
+    # The following headers values contain sensitive information that
+    # we don't want to show
+    # - Authorization
+    # - Cookie
+    # - Set-Cookie (in response)
+    # We will show "**********" instead
+    ioh = IOBuffer()
+    writeheaders(ioh, m)
+    header_str = String(take!(ioh))
```
pfitzseb, line 622:
```suggestion
    header_str = sprint(writeheaders, m)
```
should be equivalent and a bit shorter.

src/Messages.jl
```diff
@@ -609,7 +609,23 @@ function Base.show(io::IO, m::Message)
...
+    mask_headers = ["Authorization", "Cookie", "Set-Cookie"]
+    for mh in mask_headers
+        header_str = replace(header_str, Regex("($mh: ).*\$", "m") => s"\1******")
```
pfitzseb, line 625:
```suggestion
        header_str = replace(header_str, Regex("($mh: ).*\$", "mi") => s"\1******")
```
not sure if we're likely to get differently cased headers, but making the regex case insensitive seems like a good idea regardless.

fredrikekre: Should also match the start of a line, and only match `\r\n` line breaks, I think:
```suggestion
        header_str = replace(header_str, Regex("(*CRLF)^($mh: ).*\$", "mi") => s"\1******")
```

Perhaps also show `<REDACTED>` or something instead of `*****` to make it more obvious that it is not the actual header value that is shown? On the other hand, `****` is widely used to mask things like this.

fredrikekre (approved):
I added some tests for case insensitivity, combined into one regex, and added masking also for `Proxy-Authorization`.

PR #1131, opened 2023-11-27
Rename from `connection_timeout` -> `connect_timeout`

staticfloat:
I think this may be a typo; I'm a little surprised there's not a warning or something that you're using a useless keyword argument.


fredrikekre:
`kwargs...` are just forwarded down the chain at all stages I think. Perhaps at the end there should be a check that all kwargs have been picked up by some layer below.

PR #1148, opened 2024-02-14
Update readbytes! to avoid using internal API of IOBuffer less

vtjnash:
Note that this claims to be supported for all GenericIOBuffer, but Base only defines this function on a normal IOBuffer for which `pointer(data)` is contiguous space


vtjnash:
Alright, I think I fixed the remaining test issues so this is passing on all versions now

oscardssmith:
bumping this.

quinnj (approved):

PR #1159, opened 2024-03-11
Do not migrate tasks to a different thread pool

### LINKED ISSUE #1153: HTTP.request negates interactive thread usage.
I observe that using `HTTP.request` within an interactive thread makes the request run, in part, on a default thread.  The result is that when the default thread(s) are busy, the request is delayed (i.e. not "interactive").   Below is an example where we keep a thread in the default thread-pool busy by computing an approximation to pi (via the `compute` method), and we use a thread in the interactive thread-pool to make an http request once every two seconds (via the `interact` method).  We can exercise this pattern using HTTP.jl or LibCURL.jl for the http request.  In the case of HTTP.jl the `interact` method is blocked until the `compute` method finishes where-as in the case of LibCurl.jl, the `interact` and `compute` methods run simultaneously, as expected.

```julia
using HTTP,LibCURL

function compute(tic)
    @info "compute" Threads.threadpool() Threads.threadid()
    x = 1
    c = -1
    d = 3
    while true
        if time() - tic > 30
            break
        end
        x += c/d
        c *= -1
        d += 2
    end
    @info "approximation to pi is $(4*x)"
end

function interact(tic, ishttp)
    @info "interact" Threads.threadpool() Threads.threadid()
    while true
        if ishttp
            r = HTTP.request("GET", "https://example.com")
            @info "status=$(r.status)"
        else
            curl = curl_easy_init()
            curl_easy_setopt(curl, CURLOPT_URL, "https://example.com")
            curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0)
            redirect_stdout(devnull) do
                curl_easy_perform(curl)
            end
            status = Clong[300]
            curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, status)
            @info "status=$(status[1])"
            curl_easy_cleanup(curl)
        end
        sleep(2)
        if time() - tic > 32
            break
        end
    end
    nothing
end

ishttp = parse(Bool, ARGS[1])

tic = time()
t1 = Threads.@spawn :interactive interact(tic, ishttp)
sleep(.001) # not sure why this is needed, but without it, the interactive thread sometimes does not run.
t2 = Threads.@spawn :default compute(tic)

wait.((t1,t2))
```

Here is the result using `ishttp=true`:
```bash
$ julia -t 1,1 --project=@http httptrouble.jl true
┌ Info: interact
│   Threads.threadpool() = :interactive
└   Threads.threadid() = 1
┌ Info: compute
│   Threads.threadpool() = :default
└   Threads.threadid() = 2
[ Info: approximation to pi is 3.1415926553757925
[ Info: status=200
```

Here is the result using `ishttp=false`:
```bash
$ julia -t 1,1 --project=@http httptrouble.jl false
┌ Info: interact
│   Threads.threadpool() = :interactive
└   Threads.threadid() = 1
┌ Info: compute
│   Threads.threadpool() = :default
└   Threads.threadid() = 2
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: status=200
[ Info: approximation to pi is 3.141592655058851
[ Info: status=200
```

For what it is worth, the real-world [use case here](https://github.com/ChevronETC/AzManagers.jl/blob/50b241cbaeac6b7f9f3e68939b299b416cd44fff/src/AzManagers.jl#L1009-L1135) is monitoring for spot eviction events when using Azure cloud virtual machines.

I suppose the solution is to somehow propagate the interactive thread-pool usage through the HTTP.jl layers so that the appropriate thread pool can be used when `Threads.@spawn` is called, but I'm not super familiar with the structure of the layers in HTTP.jl, and I'm not sure what the best mechanism for propagating the interactive thread usage is.  But, perhaps the simplest solution is to use something like `Threads.@spawn Threads.threadpool() ...` in places like [this](https://github.com/JuliaWeb/HTTP.jl/blob/040cf0ac773670bbd1ea66e91715583a02537213/src/clientlayers/StreamRequest.jl#L39).

 - Julia 1.10.0
 - HTTP.jl 1.10.2
 - MbedTLS.jl 1.1.9

samtkaplan:
This depends on [ConcurrentUtilities#32](https://github.com/JuliaServices/ConcurrentUtilities.jl/pull/32).  Therefore, I don't expect unit tests to pass until [ConcurrentUtilities#32](https://github.com/JuliaServices/ConcurrentUtilities.jl/pull/32) is merged, and a new version of ConcurrentUtilities is released.

Together with [ConcurrentUtilities#32](https://github.com/JuliaServices/ConcurrentUtilities.jl/pull/32), this fixes #1153.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "HTTP"
 uuid = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 authors = ["Jacob Quinn", "contributors: https://github.com/JuliaWeb/HTTP.jl/graphs/contributors"]
-version = "1.10.3"
+version = "1.10.4"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
@@ -22,7 +22,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 CodecZlib = "0.7"
-ConcurrentUtilities = "2.2"
+ConcurrentUtilities = "2.4"
 ExceptionUnwrapping = "0.1"
 LoggingExtras = "0.4.9,1"
 MbedTLS = "0.6.8, 0.7, 1"
--- a/src/clientlayers/StreamRequest.jl
+++ b/src/clientlayers/StreamRequest.jl
@@ -3,6 +3,7 @@ module StreamRequest
 using ..IOExtras, ..Messages, ..Streams, ..Connections, ..Strings, ..RedirectRequest, ..Exceptions
 using LoggingExtras, CodecZlib, URIs
 using SimpleBufferStream: BufferStream
+using ConcurrentUtilities: @samethreadpool_spawn
 
 export streamlayer
 
@@ -36,7 +37,7 @@ function streamlayer(stream::Stream; iofunction=nothing, decompress::Union{Nothi
                 # use a lock here for request.context changes (this is currently the only places
                 # where multiple threads may modify/change context at the same time)
                 lock = ReentrantLock()
-                Threads.@spawn try
+                @samethreadpool_spawn try
                     writebody(stream, req, lock)
                 finally
                     Base.@lock lock begin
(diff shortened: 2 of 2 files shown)


samtkaplan:
@quinnj Let me know if there are any more changes you would like here before it can be merged?  Thanks :)

quinnj (approved):

PR #1164, opened 2024-04-05
Make multipart parsing more forgiving: start-of-line and casing

NHDaly:
I had a multipart response i wanted to parse, and julia failed to parse it. These changes let it parse:
1. The content-disposition was the _second_ line in each part, with the content-type coming first, so the `^` was failing to parse.
2. The `content-type:` key was lower-cased, not Title-Cased as expected.

Dunno if these are generally correct, but they worked in my case.

### DIFF
--- a/src/parsemultipart.jl
+++ b/src/parsemultipart.jl
@@ -122,7 +122,7 @@ end
 
 const content_disposition_regex = Parsers.RegexAndMatchData[]
 function content_disposition_regex_f()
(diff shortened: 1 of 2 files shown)


quinnj (approved):
Seems fine to me; could we add a test case that covers what was previously failing?

quinnj (approved):

quinnj:
Let me know if this is needed in a release (for 0.9 or 0.10), otherwise, I'll let is sit on master until a new release is needed.

NHDaly:
not needed; i was just trying to be a good citizen here and fix a bug I found. 😊 thanks! 👋

PR #1168, opened 2024-04-13
Don't eagerly write response body in case of redirect

quinnj:


src/clientlayers/StreamRequest.jl
```diff
@@ -144,7 +144,7 @@ end
 
 function readbody!(stream::Stream, res::Response, buf_or_stream, lock)
     n = 0
-    if !iserror(res)
+    if !iserror(res) && !isredirect(res)
```
fredrikekre, line 147: I don't think this is sufficient, you also need to check if this redirect is to be followed or not (`redirect::Bool` and `redirect_limit::Int` arguments).

This case doesn't seem to be tested either since `HTTP.get("http://localhost:8123"; redirect=false, response_stream = stderr);` from the issue description gives no body at all.

PR #1170, opened 2024-04-20
Fix lost response body data. Fixes #1166

### LINKED ISSUE #1166: Character lost in HTTP post response body.
Versions:
 - Julia 1.11.0-beta1
 - HTTP.jl 1.10.5
 - MbedTLS.jl 1.1.9

I have a HTTP server running locally, but querying the server using HTTP.post give me this:
```
In [9]: HTTP.post("http://127.0.0.1:5700/get_login_info", ["Content-Type" => "application/json"], body = "{}")
HTTP.Messages.Response:
"""
HTTP/1.1 200 OK
Content-Type: application/json
Server: Microsoft-NetCore/2.0
Date: Thu, 11 Apr 2024 11:54:50 GMT
Content-Length: 113

"status":"ok","retcode":0,"data":{"user_id":3785797225,"nickname":"\u609F\u7406\u80DA\u80CE\u7403"},"echo":null}"""
```
The response body is missing the first character, in my case, the beginning `{` in json string is lost.
Correct result should look like this curl result:
````
❯ curl http://127.0.0.1:5700/get_login_info
{"status":"ok","retcode":0,"data":{"user_id":3785797225,"nickname":"\u609F\u7406\u80DA\u80CE\u7403"},"echo":null}
````
The issue happens today when I updated julia version.

quinnj:
This creates one extra copy of the bytes instead of the view, but the result of `readuntil` is always being materialized as a `String` anyway, which was making a copy of the bytes before anyway (i.e. the behavior that changed was String(::View) copied the view bytes then made the string, and now we're explicitly making the copy and then the string is made).

### DIFF
--- a/src/IOExtras.jl
+++ b/src/IOExtras.jl
@@ -119,7 +119,7 @@ function readuntil(buf::IOBuffer,
     if l == 0
         return nobytes
     end
-    bytes = view(buf.data, buf.ptr:buf.ptr + l - 1)
+    bytes = buf.data[buf.ptr:buf.ptr + l - 1]
     buf.ptr += l
     return bytes
 end


KristofferC:
Would be good with a new release with this fix since quite a few packages error on PkgEval with issues that are likely originating from this.

fredrikekre:
https://github.com/JuliaRegistries/General/pull/105361

quinnj:
Thanks @fredrikekre

PR #1174, opened 2024-05-08
Fix 1172

cmcaine:
The IOType was not being set correctly when using a proxy.


quinnj (approved):
Thanks for the fix @cmcaine

PR #1177, opened 2024-05-10
Tune websocket performance

### LINKED ISSUE #1124: Make it easy to set TCP_NODELAY and TCP_QUICKACK on client sockets
Similar to [this PR](https://github.com/JuliaWeb/HTTP.jl/issues/629). I compared latency times in Python WebSockets vs this package and I found that while Python latency essentially equaled my ping to the server, Julia took nearly twice as long.

I did some digging and I found that I can match the times by doing the following

```julia
HTTP.WebSockets.open("wss://localhost") do ws
  Sockets.nagle(ws.io.io.io, false)
  Sockets.quickack(ws.io.io.io, true)
end
```
I propose 4 options
1. Make this behaviour more explicit in the documentation
2. I find ws.io.io.io path quite funky, and maybe it will be possible to make it more easily accessible from ws object.
3. Disable Nagle algorithm and enable TCP_QUICKACK by default
4. Create an option to pass these options into the open method

I am using
 - Julia 1.9.0
 - HTTP.jl 1.9.16

I didn't submit my code I used for benchmarking, but I am happy to do so if requested. I can implement the changes, but I want to know which of the 4 options would best align with the vision of this project

### LINKED ISSUE #1140: Websocket performance optimization
By default, `HTTP.WebSockets` uses a TCP socket configuration that is rather slow:

-  In the default setting, `HTTP.WebSockets` takes about 0.04s to respond to a request from a python [websockets](https://websockets.readthedocs.io/en/stable/) client.
- An equivalent python server takes only 0.0002s to respond.

However, if one disables Nagel's algorithm on the tcp socket, the Julia implementation is on par.
In fact, with the tuned socket below I get Julia to 1e-5s response time.

```julia
function server()
    server = WebSockets.listen!("127.0.0.1", 8081) do ws
       # without these options below, the socket is way too slow.
        Sockets.nagle(ws.io.io, false)
        Sockets.quickack(ws.io.io, true)

        for msg in ws
            WebSockets.send(ws, msg)
        end
    end
end
```

**Proposal**: It would be nice to make this tuning the default and expose some relevent settings via the keywords of `HTTP.WebSockets.listen!` and `HTTP.WebSockets.open`.

lassepe:
This PR tunes the performance of websockets by defaulting to TCP_NODELAY and TCP_QUICKACK.

Resolves #1140
Resolves #1124

### DIFF
--- a/src/Connections.jl
+++ b/src/Connections.jl
@@ -96,6 +96,9 @@ mutable struct Connection{IO_t <: IO} <: IO
     state::Any # populated & used by Servers code
 end
 
+has_tcpsocket(c::Connection) = applicable(tcpsocket, c.io)
+IOExtras.tcpsocket(c::Connection) = tcpsocket(c.io)
+
 """
     connectionkey
 
@@ -142,7 +145,7 @@ function Base.flush(c::Connection)
     # Flushing the TCP buffer requires support for `Sockets.nagle()`
     # which was only added in Julia v1.3
     @static if VERSION >= v"1.3"
-        sock = tcpsocket(c.io)
+        sock = tcpsocket(c)
         # I don't understand why uninitializd sockets can get here, but they can
         if sock.status ∉ (Base.StatusInit, Base.StatusUninit) && isopen(sock)
             Sockets.nagle(sock, false)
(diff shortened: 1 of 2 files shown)


lassepe:
@quinnj I am not sure i there is a better place to apply this fix. It feels a bit arbitrary to do this in the `upgrade` function. Alternatively, we could move this into the constructor of the websocket or maybe you even have suggestions for a "higher-up" place?

quinnj (approved):

quinnj:
Thanks @lassepe! LGTM

PR #1182, opened 2024-06-14
Add case for empty set-cookie header list

JBlaschke:
I am encountering a `ERROR: LoadError: BoundsError: attempt to access 0-element Vector{SubString{String}} at index [1]` error when making calls to the acuitischeduling.com API.

It seems that `headers(h, "Set-Cookie")` can return an list of empty lists. This is a dirty workaround -- a HTTP.jl developer might find a better solution.


quinnj (approved):
Seems like a decent fix to me

PR #1188, opened 2024-10-24
document ssl cert env var

mrufsvold:
After much hair pulling, I discovered that I needed to set two environment variables to get HTTP to pick up on a custom SSL cert path. This PR just documents that fact, so others may not need to read the source code to find out :)


mrufsvold:
I'd love to have HTTP.jl also check the "JULIA_SSL_CA_ROOTS_PATH" env variable for cert paths too so that this wouldn't be necessary, but I don't know the code base well enough to figure out the repercussions of changing anything.

quinnj:
Closing as obsolete; HTTP now relies on Reseau for TLS instead of mbetls/openssl

PR #1197, opened 2024-11-22
Fix WebSocket concurrents send

### LINKED ISSUE #1196: WebSockets does not support concurrent sends
First of all, thanks for the great HTTP.jl package, I really appreciate this project!

I've noticed an issue with the WebSocket implementation: when multiple async tasks send WebSocket messages, a race condition may occur, causing the frames to become corrupted.
 
Below a minimal working example that exposes the issue (HTTP.jl v1.10.10).

```julia
using HTTP.WebSockets

function write_message(ws, msg)
    send(ws, msg)
end

function client_twin(ws)
    for count in 1:10
        @async write_message(ws, count)
    end
end

function serve()
    server = WebSockets.listen!("127.0.0.1", 8081) do ws
        client_twin(ws)
        # just for waiting
        receive(ws)
    end

    wait(server)
end

srvtask = @async serve()

WebSockets.open("ws://127.0.0.1:8081") do ws
    try
        while true
            s = receive(ws)
            println(s)
        end
    catch e
        @error "Ws client: $e"
    end
end;

wait(srvtask)
```
```text
[ Info: Listening on: 127.0.0.1:8081, thread id: 3
┌ Error: Ws client: HTTP.WebSockets.WebSocketError(HTTP.WebSockets.CloseFrameBody(1007, "Invalid UTF-8"))
```

I've a forked HTTP.jl with a fix: https://github.com/cardo-org/HTTP.jl/commit/d60bb5eab707422c1ad3e62ca9398dd6e0bb4ef8

Would you be open to reviewing the fix and eventually accept a pull request?

greetings,
Attilio

attdona:
Fixes #1196

### DIFF
--- a/src/WebSockets.jl
+++ b/src/WebSockets.jl
@@ -182,22 +182,24 @@ end
 
 # writing a single frame
 function writeframe(io::IO, x::Frame)
-    n = write(io.io, hton(uint16(x.flags)))
+    buff = IOBuffer()
+    n = write(buff, hton(uint16(x.flags)))
     if x.extendedlen !== nothing
-        n += write(io.io, hton(x.extendedlen))
+        n += write(buff, hton(x.extendedlen))
     end
     if x.mask != EMPTY_MASK
-        n += write(io.io, UInt32(x.mask))
+        n += write(buff, UInt32(x.mask))
     end
     pl = x.payload
     # manually unroll a few known type cases to help the compiler
     if pl isa Vector{UInt8}
-        n += write(io.io, pl)
-    elseif pl isa Base.CodeUnits{UInt8, String}
-        n += write(io.io, pl)
+        n += write(buff, pl)
+    elseif pl isa Base.CodeUnits{UInt8,String}
+        n += write(buff, pl)
     else
-        n += write(io.io, pl)
+        n += write(buff, pl)
     end
+    write(io.io, take!(buff))
     return n
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -31,6 +31,7 @@ isok(r) = r.status == 200
         "mwe.jl",
         "httpversion.jl",
         "websockets/autobahn.jl",
+        "websockets/multiple_writers.jl",
     ]
     # ARGS can be most easily passed like this:
     # import Pkg; Pkg.test("HTTP"; test_args=`ascii.jl parser.jl`)
--- a/test/websockets/multiple_writers.jl
+++ b/test/websockets/multiple_writers.jl
@@ -1,0 +1,43 @@
+using Test
+using HTTP.WebSockets
+
+function write_message(ws, msg)
+    send(ws, msg)
+end
+
+function client_twin(ws)
+    for count in 1:10
+        @async write_message(ws, count)
+    end
+end
+
+function serve(ch)
+    WebSockets.listen!("127.0.0.1", 8081) do ws
+        client_twin(ws)
+        response = receive(ws)
+        put!(ch, response)
+    end
+end
+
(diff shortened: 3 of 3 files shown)


attdona:
Ok, before merging I can try to do the comparison using a lock instead, I think it should work, let's see the benchmarks ....

quinnj:
Yeah, it makes sense that total throughput of concurrent writes would go down w/ the lock. I guess that's a shift in the paradigm of how I've traditionally used websockets where it's always a single send-and-receive. So all in all, let's go w/ the IOBuffer per writeframe approach. Is this PR ready to merge then?

(thanks for doing the ReentrantLock approach; I appreciate it!)

fredrikekre:
The buffer could possibly be a task local variable and reused between calls, but let's merge this to get the fix in first.

fredrikekre (approved):

attdona:
I tried the task-local storage but results are not what was expected, it seems that memory allocation gets worst, and I find
this hard to understand.

Below what I tried.

The modified `writeframe()` method:

```julia
const frameio = TaskLocalValue{IOBuffer}(() -> IOBuffer())

function writeframe(io::IO, x::Frame)
    n = write(frameio[], hton(uint16(x.flags)))
    if x.extendedlen !== nothing
        n += write(frameio[], hton(x.extendedlen))
    end
    if x.mask != EMPTY_MASK
        n += write(frameio[], UInt32(x.mask))
    end
    pl = x.payload
    # manually unroll a few known type cases to help the compiler
    if pl isa Vector{UInt8}
        n += write(frameio[], pl)
    elseif pl isa Base.CodeUnits{UInt8,String}
        n += write(frameio[], pl)
    else
        n += write(frameio[], pl)
    end
    write(io.io, take!(frameio[]))
    return n
end
```
  
`bench.jl` send 100 packets using 100 async tasks:

```julia
using HTTP.WebSockets

const COUNT = 100

function write_message(ws, msg)
    for i in 1:100
        send(ws, msg)
    end
end

function client_twin(ws)
    for count in 0:COUNT
        @async write_message(ws, count)
    end
end

function serve()
    server = WebSockets.listen!("127.0.0.1", 8081) do ws
        client_twin(ws)
        receive(ws)
    end
    wait(server)
end

function write()
    WebSockets.open("ws://127.0.0.1:8081") do ws
        try
            s = receive(ws)
            t = time()
            count = 0
            while count < COUNT * 100
                s = receive(ws)
                count += 1
            end
            delta = time() - t
            println("delta time: $delta")
            close(ws)
        catch e
            @error "ws client: $e"
        end
    end
end

srvtask = @async serve()
```

```shell
julia -i bench.jl
julia> write()
julia> @time write()

# writeframe current version with additional buffer
delta time: 0.13498997688293457
  0.139230 seconds (142.53 k allocations: 11.041 MiB)

# writeframe with TaskLocalValue
delta time: 0.14266705513000488
  0.146890 seconds (183.03 k allocations: 11.225 MiB)
```

fredrikekre:
`IOBuffer` is not a concrete type so every access to `frameio` is unstable. With
```
const frameio = TaskLocalValue(() -> IOBuffer())
```
and using `buff = frameio[]` once at the top of the function and replacing `write(io.io, take!(frameio[]))` with `write(io.io, seekstart(buff))` you might see some differences. I get
```
julia> @time write();
delta time: 0.13005280494689941
  0.133447 seconds (202.61 k allocations: 11.956 MiB)
```
for the master branch and
```
julia> @time write();
delta time: 0.022337913513183594
  0.031560 seconds (70.95 k allocations: 18.650 MiB, 5.93% gc time)
```
with those changes to your bechmark.

PR #1198, opened 2024-11-22
Update/Fix CI setup

IanButterworth:
New nightly failures are some unexpected allocations
```
HTTPVersion: Test Failed at /home/runner/work/HTTP.jl/HTTP.jl/test/httpversion.jl:19
  Expression: #= /home/runner/work/HTTP.jl/HTTP.jl/test/httpversion.jl:19 =# @allocated(HTTPVersion("1.1")) == 0
   Evaluated: 16 == 0

  Expression: #= /home/runner/work/HTTP.jl/HTTP.jl/test/httpversion.jl:20 =# @allocated(VersionNumber(HTTPVersion("1.1"))) == 0
   Evaluated: 64 == 0
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,9 @@ jobs:
           - os: windows-latest
             version: '1'
             arch: x86
+        exclude:
+          - os: macOS-latest
+            version: 'min' # no apple silicon support for Julia 1.6
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - 'min'
           - 'lts'
           - '1' # automatically expands to the latest stable 1.x release of Julia
           - 'nightly'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,54 +21,43 @@ jobs:
...
+          - 'lts'
           - '1' # automatically expands to the latest stable 1.x release of Julia
           - 'nightly'
         os:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
         arch:
-          - x64
+          - default
```
fredrikekre, line 32: Can be dropped?

.github/workflows/ci.yml
```diff
@@ -21,54 +21,43 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - 'lts'
```
fredrikekre, line 24: Should either keep 1.6 or bump julia compat in Project.toml I guess.

fredrikekre (approved):

.github/workflows/ci.yml
IanButterworth: It can't because we use arch for an included x86 so it needs to be populated with the default

.github/workflows/ci.yml
IanButterworth, line 24:
```suggestion
          - 'min'
          - 'lts'
```

IanButterworth: Good point, I added min

PR #1195, opened 2024-11-20
add an offline precompile workload

IanButterworth:
Fixes https://github.com/JuliaWeb/HTTP.jl/issues/1194
Or at least makes decent progress.

Before
```
julia> using HTTP
   1732.9 ms  ✓ HTTP (precompile time)

julia> @time HTTP.get("http://google.com");
  4.109167 seconds (13.00 M allocations: 658.440 MiB, 2.90% gc time, 96.96% compilation time: 2% of which was recompilation)

julia> @time HTTP.get("http://google.com");
  0.155600 seconds (11.27 k allocations: 746.781 KiB, 12.46% compilation time)

julia> @time HTTP.get("http://google.com");
  0.122175 seconds (574 allocations: 196.016 KiB)
```

This PR
```
julia> using HTTP
   6780.5 ms  ✓ HTTP (precompile time)

julia> @time HTTP.get("http://google.com");
  0.678762 seconds (1.17 M allocations: 64.638 MiB, 1.65% gc time, 77.77% compilation time: 86% of which was recompilation)

julia> @time HTTP.get("http://google.com");
  0.174314 seconds (11.28 k allocations: 746.500 KiB, 10.45% compilation time)

julia> @time HTTP.get("http://google.com");
  0.137826 seconds (596 allocations: 196.641 KiB)
```

All the big precompiles are baked in and just one big recompile left.
Screenshot 2024-11-19 at 10 38 17 PM
```
[...]

PR #1199, opened 2024-11-25
Don't run the precompile workload on Julia 1.9

DilumAluthge:
I think the precompile workload is broken on Julia 1.9.

I think the easiest approach will just be to skip the precompile workload on Julia 1.9. We'll continue to run the precompile workload on Julia 1.10+.

Cross-reference: https://github.com/JuliaWeb/HTTP.jl/pull/1195#issuecomment-2496480235

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
         version:
           - 'min'
           - 'lts'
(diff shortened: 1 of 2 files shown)


IanButterworth:
Superseded by https://github.com/JuliaWeb/HTTP.jl/pull/1201

PR #1201, opened 2024-11-25
Fix trailing task precompilation issue on 1.9. Run CI on 1.9

IanButterworth:
Fixes https://github.com/JuliaWeb/HTTP.jl/issues/1202
See https://github.com/JuliaWeb/HTTP.jl/pull/1195#issuecomment-2496480235

https://github.com/JuliaWeb/HTTP.jl/pull/1200 shows that without this, things go bad on 1.9

The print that it was causing is weird.. just seems to show a task object.. 

But it seems to be something like 1.9 not yielding for tasks to exit before serialization, so do it explicitly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
         version:
           - 'min'
           - 'lts'
+          - '1.9' # test because we know there's a precompilation issue that should be fixed
           - '1' # automatically expands to the latest stable 1.x release of Julia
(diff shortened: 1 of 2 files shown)


DilumAluthge (approved):
Thanks!

DilumAluthge:
@IanButterworth or @quinnj Can we get a new release?

IanButterworth:
Way ahead of you https://github.com/JuliaRegistries/General/pull/120120

Issue #1204, opened 2024-12-11
Cannot precompile (connection refused)


fmarotta:
- Julia 1.10.3
 - HTTP.jl 1.10.11 to 1.10.13
 - MbedTLS.jl 1.1.9
- OS: Rocky Linux 8.10

Hi, the package fails to precompile with the following `ERROR: LoadError: HTTP.ConnectError for url = https://localhost:57813: IOError: connect: connection refused (ECONNREFUSED)`. Precompilation works fine for 1.10.10. I think v1.10.11 started to use `PrecompileTools`, but it may be a red herring. Also, everything works in my arch linux laptop with Julia 1.9.4.

IanButterworth:
We should make the workload soft error irrespective of if we can fix this specific case.

PR #1205, opened 2024-12-11
Soft error the precompile workload

IanButterworth:
Protect against oddities like https://github.com/JuliaWeb/HTTP.jl/issues/1204

Just a `try-catch-@info`. Recommend reviewing with hide whitespace changes on.

There may be a specific fix for that case, but this seems worth doing irrespective of that.

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -1,40 +1,44 @@
 using PrecompileTools: @setup_workload, @compile_workload
 
-@setup_workload begin
-    # These need to be safe to call here and bake into the pkgimage, i.e. called twice.
-    Connections.__init__()
-    MultiPartParsing.__init__()
-    Parsers.__init__()
+try
+    @setup_workload begin
+        # These need to be safe to call here and bake into the pkgimage, i.e. called twice.
+        Connections.__init__()
+        MultiPartParsing.__init__()
+        Parsers.__init__()
 
-    # Doesn't seem to be needed here, and might not be safe to call twice (here and during runtime)
-    # ConnectionRequest.__init__()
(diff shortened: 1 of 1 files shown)


src/precompile.jl
```diff
@@ -1,40 +1,44 @@
...
 
-    HTTP.forceclose(server)
-    yield() # needed on 1.9 to avoid some issue where it seems a task doesn't stop before serialization
-    server = nothing
+        HTTP.forceclose(server)
+        yield() # needed on 1.9 to avoid some issue where it seems a task doesn't stop before serialization
+        server = nothing
+    end
+catch e
+    @warn "Error during precompilation workload:" exception=(e, catch_backtrace())
```
IanButterworth, line 43:
```suggestion
    @warn "Ignoring an error that occurred during the precompilation workload" exception=(e, catch_backtrace())
```

src/precompile.jl
```diff
@@ -1,40 +1,44 @@
...
 
-    HTTP.forceclose(server)
-    yield() # needed on 1.9 to avoid some issue where it seems a task doesn't stop before serialization
-    server = nothing
+        HTTP.forceclose(server)
+        yield() # needed on 1.9 to avoid some issue where it seems a task doesn't stop before serialization
+        server = nothing
+    end
+catch e
+    @warn "Ignoring an error that occurred during the precompilation workload" exception=(e, catch_backtrace())
```
IanButterworth, line 43:
```suggestion
    @info "Ignoring an error that occurred during the precompilation workload" exception=(e, catch_backtrace())
```

quinnj (approved):

PR #1206, opened 2024-12-12
Do not migrate tasks to a different thread pool (backport of #1159)

Drvi:


IanButterworth:
This seems like functionality Base should have. Is there a feature request for it?

Drvi:
I think the tests are broken on 1.11+ because of we're touching internals that changed with the Memory PR.

IanButterworth:
```
Threads.@spawn Threads.threadpool()
```
That looks good to me. Any reason not to use that?

NHDaly (approved):
makes sense to me 👍

Issue #1208, opened 2024-12-18
Compilation Hang v1.10.14 


eohne:
installed through
```julia
] add HTTP
```

Installing version 1.9.19 worked for me without precompilation hang. 

GitHub suggests latest version is 1.9.19 but add HTTP installs version 1.10.14

IanButterworth:
Can you share `versioninfo()`.

Also, some things are being backported to 1.9 hence why the _latest_ release is 1.9.19, whilst the proper latest release is 1.10.14

eohne:
```julia
julia> versioninfo()
Julia Version 1.11.2
Commit 5e9a32e7af2 (2024-12-01 20:02 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (aarch64-linux-gnu)
```

Full transparency this is running on my Galaxy S24 Ultra.

IanButterworth:
With 1.19.19 can you try each of this line by line in the repl and share the output, and note where it hangs.

```
  gzip_data(data::String) = read(GzipCompressorStream(IOBuffer(data)))
  _port = 57813
  cert, key = joinpath.(@__DIR__, "../test", "resources", ("cert.pem", "key.pem"))
  sslconfig = MbedTLS.SSLConfig(cert, key)

  server = HTTP.serve!("0.0.0.0", _port; verbose = -1, listenany=true, sslconfig=sslconfig) do req
      HTTP.Response(200,  ["Content-Encoding" => "gzip"], gzip_data("dummy response"))
  end

  _port = HTTP.port(server)
  url = "https://localhost:$_port"

  env = ["JULIA_NO_VERIFY_HOSTS" => "localhost", "JULIA_SSL_NO_VERIFY_HOSTS" => nothing, "JULIA_ALWAYS_VERIFY_HOSTS" => nothing]
  withenv(env...) do
      HTTP.get(url);
  end

  HTTP.forceclose(server)
```

eohne:
1.9.19 works fine it's. 1.10 that hangs for me. 
I would just mention that somewhere in the readme since 1.10.14 is installed by default

IanButterworth:
Please do use 1.19.19 for these tests.

The path was wrong. Updated:
```
gzip_data(data::String) = read(GzipCompressorStream(IOBuffer(data)))
  _port = 57813
  cert, key = joinpath.(pkgdir(HTTP), "/test", "resources", ("cert.pem", "key.pem"))
[...]

PR #1210, opened 2024-12-23
skip precompile workload if localhost cannot be resolved

IanButterworth:
Fixes https://github.com/JuliaWeb/HTTP.jl/issues/1208

Or is there a way to work around it, not just skip it?


quinnj (approved):
Yeah, this seems fine enough to me.

PR #1213, opened 2025-01-05
[BREAKING] HTTP.jl 2.0: Overhaul internals by switching to use aws-crt libraries

quinnj:
This isn't quite ready to merge yet, but starting the PR to get CI running and document what there is left to do:

[ ] Incorporate @IanButterworth's recent precompile work
[ ] Update documentation: manual and all inline docs; also write up a 1.X -> 2.0 migration guide
[ ] Update upstream AWS CRT libraries and fix anything required in our wrappers


quinnj:
Made some progress here, but blocked on 2 issues I need to dig into:
* There's an occasional segfault from an unhandled C callback BoundsError; it's when we try to write to the temporary Base.BufferStream and it seems like the code should correctly resize the underlying IOBuffer of the BufferStream, but it doesn't for some reason and then we get a BoundsError; I'm actively investigating this now to try to find out if it's an actual bug in Base.BufferStream/IOBuffer, or how we're using BufferStream (pretty vanilla, but maybe?)
* There's an sporadic test failure where providing an IOBuffer as a response body seems to sometimes duplicate part of the body writing. Possibly related to the first issue.

quinnj:
Ok, turns out the segfault is due to Base.BufferStream `readavailable` not being threadsafe, there's a PR to fix here: https://github.com/JuliaLang/julia/pull/57211. I'll add a compat shim here until that's fully fixed/released. I can confirm that the same root cause was causing both issues mentioned above.

src/precompile.jl
```diff
@@ -2,44 +2,39 @@ using PrecompileTools: @setup_workload, @compile_workload
...
+            @info "Shutting down HTTP server"
+            close(server)
+            @info "HTTP server shut down"
             yield() # needed on 1.9 to avoid some issue where it seems a task doesn't stop before serialization
             server = nothing
+            close_all_clients!()
+            close_default_aws_server_bootstrap!()
+            close_default_aws_client_bootstrap!()
+            close_default_aws_host_resolver!()
+            close_default_aws_event_loop_group!()
```
Octogonapus, line 37: Could a new event loop group, etc. be created instead of using the default? I wouldn't want this package to conflict with others because they both use default resources.

quinnj: In my opinion, it's better that they _do_ share the same event loop group. The point is that the default group should be able to saturate os resources, so it's better that packages/libraries _coordinate_ tasks in the same group instead of having competing groups then possibly scheduling tasks in a non-optimal configuration.

Octogonapus: That's a fair point, but if an event loop group is shared, who is responsible for closing it? It could be closed by one user when another user still needs it.

quinnj: > That's a fair point, but if an event loop group is shared, who is responsible for closing it? It could be closed by one user when another user still needs it.

In my mind, (and it would probably be helpful to document this perhaps in the LibAwsIO README or something), there are 2 different scenarios for LibAwsIO usage:
1. Casual, REPL usage, scripts, etc. In this case, you may load libraries (Postgres.jl, HTTP.jl, etc.) where you want things to "just work" and do sensible default things. That's currently the case w/ the machinery here. Sensible default resources are set up, they can be used, we're not too worried about cleaning things up because everything will be taken care of when the process exits.
2. Private application. In this case, again, you load libraries or perhaps use LibAwsIO directly, but you may have unique, specific settings you want to set for these LibAwsIO resources. You're free to create your own (not too hard, but takes some manual setup/effort), and then you can set them in LibAwsIO so that any package that relies on them will use the same, manually set up resources you'd like. This is as opposed to needing to pass the resources manually, for example, to every HTTP call you may have in a huge, nested application. Those kinds of one-off custom resources _should_ still be supported, so if you had a specific HTTP request that you wanted to use slightly different resource group, you could pass them to that call specifically.

So to circle back and answer the original question: if you're relying on the default shared resources, then you _shouldn't_ be closing them or worrying about closing them. They're global shared. If you're in a situation where you _need_ to close them, then you should probably be initializing them yourself and passing them or setting them as the defaults yourself, and then you're free to close them.

quinnj: > Perhaps the names of these functions is confusing because AFAIU the aws calls are referencing/de-referencing them, so it closes when the counter reaches zero. So confusing to say `close`?

Hmmmm.....perhaps. But we _are_ specifically calling this `close_*default*_...` which we are specifically interested in closing the _default_ resources that have been set here. So yes, if you happen to have your own references to those resources, then calling these will prematurely finalize then and you wouldn't want to use them afterwards.

We could perhaps compose these a bit better with individual `close` methods on, e.g. the `EventLoopGroup` and then these close defaults could use that.

IanButterworth: Doesn't the crt orchestrate this with reference counting? https://github.com/awslabs/aws-c-io/blob/3041dabfc13fe9bc9a0467e15aa1d5a09c7fc06f/include/aws/io/event_loop.h#L169-L174

PR #1215, opened 2025-02-18
Add public names on supported Julia versions

fredrikekre:


quinnj:
Sorry I was slow on the responding/review here; I think this is a good idea; I've tried to apply it to the refactored 2.0 code in #1312

PR #1065, opened 2023-06-20
Integrate with ExceptionUnwrapping

quinnj:
cc: @NHDaly

PR #1222, opened 2025-05-08
Accept `URI` in `download` function

mfsch:
The `request`-based functions (e.g. `HTTP.get`) accept the `URI` type for the `url` argument, but the `download` function currently only accepts strings. This (tiny) change allows `URI`s there too.


fredrikekre:
Looks good, can you add a test?

mfsch:
I’ve added a test – however, it looks like the tests for `download` were commented out in #1065. Maybe @quinnj remembers why? Commenting them in an running `Pkg.test("HTTP", test_args = ["download.jl"])` runs without error for me.

quinnj:
Closing as `download` has now been removed from HTTP

PR #1229, opened 2025-07-15
Try a real URL first during precompile to avoid firewall warnings

IanButterworth:
Apparently Windows firewall can warn about our local server being open, so it might be worth using the internet if available?

Developed with Claude.


BioTurboNick:
Just noticed this because it wasn't linked to my issue, but maybe nothing so complicated is needed? #1230

IanButterworth:
Great. Sorry, missed the issue. FYI @davidanthoff

PR #1232, opened 2025-08-25
Fix redirect response bodies being written to response_stream

### LINKED ISSUE #1165: Response body for redirects not ignored in streaming mode
HTTP.jl do not ignore the response body in redirects as discovered here: https://discourse.julialang.org/t/bad-cert-when-accessing-the-general-registry/112738/10. The is introduced in HTTP.jl 1.6 (1.0 - 1.5 is OK) so the bug is in this range: https://github.com/JuliaWeb/HTTP.jl/compare/v1.5.5...v1.6.3 which includes a number of patches related to redirects.

---

Sample server:
```julia
server = HTTP.listen!(8123) do http
    if http.message.target == "/200"
        HTTP.setstatus(http, 200)
        HTTP.startwrite(http)
        HTTP.write(http, "This is /200\n")
    else
        HTTP.setstatus(http, 301)
        HTTP.setheader(http, "Location" => "/200")
        HTTP.startwrite(http)
        HTTP.write(http, "Go to /200\n")
    end
    return
end
```
HTTP.jl:
```
julia> HTTP.get("http://localhost:8123"; redirect=false, response_stream = stderr);
Go to /200                                                           # <------

julia> HTTP.get("http://localhost:8123"; response_stream = stderr);
Go to /200                                                           # <------
This is /200                                                         # <------

```

curl:
```
$ curl -v http://localhost:8123
*   Trying 127.0.0.1:8123...
* Connected to localhost (127.0.0.1) port 8123 (#0)
> GET / HTTP/1.1
> Host: localhost:8123
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Location: /200
< Transfer-Encoding: chunked
<
Go to /200                                                           # <------
* Connection #0 to host localhost left intact

$ curl -Lv http://localhost:8123
*   Trying 127.0.0.1:8123...
* Connected to localhost (127.0.0.1) port 8123 (#0)
> GET / HTTP/1.1
> Host: localhost:8123
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Location: /200
< Transfer-Encoding: chunked
<
* Ignoring the response-body                                         # <------
* Connection #0 to host localhost left intact
* Issue another request to this URL: 'http://localhost:8123/200'
* Found bundle for host localhost: 0x55d6fea933c0 [serially]
* Can not multiplex, even if we wanted to!
* Re-using existing connection! (#0) with host localhost
* Connected to localhost (127.0.0.1) port 8123 (#0)
> GET /200 HTTP/1.1
> Host: localhost:8123
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
<
This is /200                                                         # <------
* Connection #0 to host localhost left intact
```

IanButterworth:
Note, written by Claude, hence the draft.

----

Fixes #1165

## Problem

When using `response_stream` with HTTP requests that involve redirects, both the redirect response body AND the final response body were being written to the stream. This is incorrect behavior - only the final response body should be written to the stream, and redirect response bodies should be ignored (similar to how curl handles redirects with "Ignoring the response-body").

## Root Cause

The issue was introduced in the version range v1.5.5...v1.6.3, specifically related to changes in PR #975 that fixed StatusError exception handling. While that fix was correct and necessary, it highlighted an existing issue in StreamRequest.jl:

- Redirect responses (3xx status codes) return `false` for `iserror(res)` but `true` for `isredirect(res)`
- In `readbody!()`, they were going through the normal "success" path and writing their bodies to `response_stream`
- When redirects are followed, these intermediate response bodies should be discarded, not written to the stream

## Solution

Modified StreamRequest.jl to:

1. **Added `willredirect()` function** that determines if a redirect response will actually be followed (using the same logic as `RedirectRequest` layer)
2. **Updated the condition** in `readbody!()` from `!iserror(res)` to `!iserror(res) && !willredirect(res)`
3. **Redirect responses that will be followed** are now treated like error responses - their bodies are read but stored in request context instead of written to `response_stream`

## Behavior

- ✅ **Redirect followed**: Only final response body written to `response_stream`
- ✅ **Multiple redirects**: Only final response body written to `response_stream`
- ✅ **`redirect=false`**: Redirect response body written to `response_stream` (correct)
- ✅ **Redirect limit reached**: Final redirect response body written to `response_stream` (correct)
- ✅ **No redirects**: Normal response body written to `response_stream` (unchanged)

## Testing

Added comprehensive test case in client.jl that covers:
- Basic redirect with `response_stream`
- Multiple redirects with `response_stream`
- Redirect disabled (`redirect=false`) with `response_stream`

The fix ensures HTTP.jl now matches the expected behavior described in the issue and aligns with how other HTTP clients (like curl) handle redirect response bodies in streaming mode.

### DIFF
--- a/src/clientlayers/StreamRequest.jl
+++ b/src/clientlayers/StreamRequest.jl
@@ -1,6 +1,7 @@
 module StreamRequest
 
 using ..IOExtras, ..Messages, ..Streams, ..Connections, ..Strings, ..RedirectRequest, ..Exceptions
+using ..Messages: isredirect, allow_redirects, redirectlimitreached, header
 using CodecZlib, URIs
 using SimpleBufferStream: BufferStream
 using ConcurrentUtilities: @samethreadpool_spawn
@@ -143,9 +144,21 @@ function readbody(stream::Stream, res::Response, decompress::Union{Nothing, Bool
     end
 end
 
+# Check if this redirect response will be followed
+# Based on the logic in RedirectRequest.redirectlayer
+function willredirect(res::Response)
+    req = res.request
+    return (
+        isredirect(res) &&
+        allow_redirects(req) &&
+        !redirectlimitreached(req) &&
+        header(res, "Location") != ""
+    )
+end
+
 function readbody!(stream::Stream, res::Response, buf_or_stream, lock)
     n = 0
-    if !iserror(res)
+    if !iserror(res) && !willredirect(res)
         if isbytes(res.body)
             if length(res.body) > 0
                 # user-provided buffer to read response body into
@@ -179,6 +192,7 @@ function readbody!(stream::Stream, res::Response, buf_or_stream, lock)
         # read the response body into the request context so that it can be
         # read by the user if they want to or set later if
         # we end up not retrying/redirecting/etc.
+        # This handles both error responses and redirect responses that will be followed
         Base.@lock lock begin
             res.request.context[:response_body] = read(buf_or_stream)
         end
--- a/test/client.jl
+++ b/test/client.jl
@@ -287,6 +287,73 @@ end
         @test isok(HTTP.request(read_method, "https://$httpbin/redirect-to?url=http%3A%2F%2Fgoogle.com", socket_type_tls=tls))
     end
 
+    @testset "Client Redirect Response Stream" begin
+        # Test that redirect response bodies are not written to response_stream 
+        # (Issue #1165: https://github.com/JuliaWeb/HTTP.jl/issues/1165)
+        server = nothing
+        try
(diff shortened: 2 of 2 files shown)


quinnj:
Sorry for ghosting this; it must have fallen off my radar. I've been churning through internal rewrites for a while as well, so definitely fell behind on review. The original issue though shoudl be resolved in the latest 2.0 release; verified locally.

PR #1241, opened 2025-12-22
Update example and docs

thirtysixbananas:
This example failed. Looking at the code there's never any "excess", as the channel type is `socketready = Channel{Tuple{IO, Response}}(0)`.

At first I thought this was just an outdated example, what confuses me is that from git blame I can't tell why someone thought there was an excess. Maybe it's just a mistake, as opposed to being out of date, or maybe I'm missing something.

Please review.


quinnj:
Sorry for the slow review/response here; closing as obsolete since `openraw` was actually removed in the 2.0 release (we could look into adding the functionality back if people have strong opinions)

PR #1250, opened 2026-04-12
docs: migrate examples from JSON3.jl to JSON.jl v1

### LINKED ISSUE #1249: is JSON3 deprecated?
This package has been deprecated. Please migrate to [JSON.jl v1](https://github.com/JuliaIO/JSON.jl) (and see the [migration guide](https://juliaio.github.io/JSON.jl/stable/migrate/#Migration-guide-for-JSON3.jl))

JSON.jl  should be used in Examples?

s-celles:
Closes #1249.

JSON3.jl is deprecated. Migrate the two bundled REST-server examples to JSON.jl v1 per the official migration guide:

  - JSON3.read(body, T) -> JSON.parse(body, T)
  - JSON3.write(x)      -> JSON.json(x)
  - StructTypes.StructType(::Type{T}) = StructTypes.Mutable() -> removed
    (JSON.jl v1 auto-detects struct fields)

Add an explicit positional inner constructor `Animal(id, userId, type, name)` to both examples so JSON.jl v1's StructUtils can construct the struct from its parsed fields. The existing zero-arg `Animal()` constructor is preserved because the in-script client section still uses `x = Animal(); x.type = ...`.

Add a regression test (test/examples_no_json3.jl) that walks docs/examples/ and asserts no .jl file references the JSON3 or StructTypes modules. This test was written first, observed to fail on the unmigrated tree (the TDD red step required by the project constitution), and turned green by the migration.

Two pre-existing example bugs were fixed along the way because they blocked end-to-end validation of the migration on Julia >= 1.12:

  - The leading `"""..."""` description in both files used to be parsed as a docstring attached to the following `using` statement, which Julia 1.12 rejects with "cannot document the following expression". Convert both to `#=...=#` block comments.
  - cors_server.jl's intentional bad-path probe (`HTTP.get(".../badpath")`) raised on the 404 because HTTP.jl raises on error statuses by default. Pass `status_exception=false` so the example exercises the cors404 handler as intended.

Both fixes are documented under `### Fixed` in CHANGELOG.md alongside the main `### Changed` migration entry.

No public API of HTTP.jl changes. Project.toml [deps] is unchanged; JSON.jl remains a test/extras dependency only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Documentation examples migrated from the deprecated JSON3.jl to JSON.jl v1.
+  Affects `docs/examples/simple_server.jl` and `docs/examples/cors_server.jl`.
+  No public API of HTTP.jl changes. ([#1249])
+### Fixed
+- `docs/examples/simple_server.jl` and `docs/examples/cors_server.jl` no longer
+  fail to parse on Julia ≥ 1.12: the leading description is now a `#=...=#`
+  block comment instead of a top-of-file string literal that the parser tried
+  to attach as a docstring to the following `using` statement. ([#1249])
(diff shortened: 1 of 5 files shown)


quinnj:
sorry to be slow on this; this has been updated separately in the HTTP 2.0 release

PR #1251, opened 2026-05-17
Add support for passing IO objects to Websockets.open()

JamesWrigley:
This is handy if the caller already has the required stream/socket open.

Written with help from Claude :robot:


JamesWrigley:
One failure on 1.6, and it looks like the 32bit windows CI is hanging.

JamesWrigley:
Now rewritten to work on v2 :crossed_fingers: I think the design is ok (went through a couple rounds with Claude) though I'm certainly no expert.

src/http_websockets.jl
```diff
@@ -1044,6 +1075,105 @@ function open(
...
+redirect, cookie, and timeout options do not apply.
+
+`io` is **not** closed by `open` — the caller retains ownership of the underlying
+stream's lifetime. As with the URL form, calling `open` with a function closes
+the WebSocket with status code `1000` when `f` returns (the close frame is sent
+over `io`, but `io` itself stays open).
+"""
+function open(
+    io::IO;
+    suppress_close_error::Bool=false,
```
quinnj: this keyword isn't getting passed along to the `_open_client_websocket_io` call; I think I'd prefer a one-line definition here to simplify

src/http_core.jl
```diff
@@ -1056,6 +1056,34 @@ function defaultheader!(headers::Headers, item::Pair)
...
+    eof(c.io) && return 0
+    n = min(bytesavailable(c.io), Int(nb))
+    return readbytes!(c.io, b, n)
+end
+
+Base.read(c::_RawIOConn, ::Type{UInt8}) = read(c.io, UInt8)
+Base.write(c::_RawIOConn, x::UInt8) = write(c.io, x)
+Base.unsafe_write(c::_RawIOConn, p::Ptr{UInt8}, n::UInt) = unsafe_write(c.io, p, n)
+Base.eof(c::_RawIOConn) = eof(c.io)
+
```
quinnj: I think this could all be a bit simpler than having a dedicated struct. I think you only really need 2 methods of a new HTTP-specific function called `_blocking_readybtes!` or something. For Reseau conn types, you'd pass through to `readbytes!(; all=false)` and for all other IO, you'd have the readbytes! implementation here.

I don't think I realized this subtle different in how Reseau.Conns work, but it seems simpler to have ~6 lines for `_blocking_readybtes!` vs. this whole struct + passthrough.

src/http_websockets.jl
JamesWrigley: Fixed in a21f95ed.

src/http_core.jl
JamesWrigley: Ah yeah, that is much simpler. Refactored in a21f95ed.

quinnj (approved):

PR #1257, opened 2026-05-29
test(server): fix intermittent Windows CI hang + add suite hang-watchdog

quinnj:
## What this fixes

The intermittent **Windows CI hang** that struck a different timeout/network testset each run (and previously ran to the 6-hour cap before #1256 added a 25-minute timeout).

## Root cause (captured, not guessed)

A test-suite **hang watchdog** (added here in `runtests.jl`) dumped task backtraces when a Windows run exceeded its budget, pinpointing the stuck task:

```
readavailable      → Reseau 3_tcp.jl:883   (parked in waitread/pollwait!)
_read_until_quiet  → test/http_server_http1_tests.jl
@test              → idle-timeout testset
```

`_read_until_quiet` loops `set_read_deadline!` + `readavailable`; on Windows the **re-armed read deadline intermittently fails to wake the parked `readavailable`**, so it blocks forever. The idle-timeout testset called `_read_until_quiet` **directly** (unguarded), unlike `_raw_http_request` which #1256 already watchdog-wrapped — so nothing bounded it.

The underlying transport bug is filed as **JuliaServices/Reseau.jl#107** (it is *not* reproducible in isolation — four targeted Reseau repros pass cleanly on Windows; it needs the full integrated poller load, which is why it was so elusive).

## Changes

- **`runtests.jl`** — hang watchdog that dumps all task backtraces if the suite exceeds `HTTP_HANG_WATCHDOG_S` (default 1200s, under the 25-min job cap) and hard-exits, so any future Windows hang is self-diagnosing. *(This is how the root cause was found; worth keeping permanently.)*
- **`_read_until_quiet`** — bounded with the existing `_run_with_timeout` task watchdog, so a Reseau read-deadline strand fails fast instead of hanging. Covers direct callers and `_raw_http_request` callers.
- **write-timeout test** — guard the `take!` so a missed write-timeout can't block on an empty channel (a latent second hang).
- A looped, watchdog-bounded write-timeout repro testset (diagnostic; harmless, can stay as a regression guard).

## Result

Windows CI no longer hangs on this class of strand — a strand now surfaces as a fast, legible test failure with a backtrace, while the common case stays green. Verified locally on macOS (server suite passes).

### DIFF
--- a/test/http_server_http1_tests.jl
+++ b/test/http_server_http1_tests.jl
@@ -111,34 +111,43 @@ function _read_until_deadline(conn::NC.Conn; timeout_s::Float64 = 1.0)::String
 end
 
 function _read_until_quiet(conn::NC.Conn; timeout_s::Float64 = 1.0, quiet_timeout_s::Float64 = 0.1)::String
-    buf = Vector{UInt8}(undef, 1024)
-    out = UInt8[]
-    deadline_ns = Int64(time_ns()) + round(Int64, timeout_s * 1.0e9)
-    saw_bytes = false
-    while true
-        remaining_ns = deadline_ns - Int64(time_ns())
-        remaining_ns <= 0 && break
-        read_timeout_s = saw_bytes ? min(quiet_timeout_s, remaining_ns / 1.0e9) : (remaining_ns / 1.0e9)
-        NC.set_read_deadline!(conn, Int64(time_ns()) + round(Int64, read_timeout_s * 1.0e9))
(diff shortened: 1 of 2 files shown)


quinnj:
Superseded by #1255, which now carries the watchdog + _read_until_quiet fix (folded in alongside the WebSocket feature so they land and run together). Root cause tracked in JuliaServices/Reseau.jl#107.

PR #1259, opened 2026-05-29
Percent-decode userinfo before building Basic auth header (RFC 3986)

### LINKED ISSUE #1258: v2.0.0: userinfo not percent-decoded before Basic auth (regression vs 1.x)
**HTTP.jl v2.0.0** base64-encodes URL `userinfo` **without percent-decoding** it before building the `Basic` credential, so percent-encoded credentials produce a wrong `Authorization` / `Proxy-Authorization` header ([RFC 3986 §3.2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1)). This is a regression from 1.x (which decoded). It hits request URLs **and** authenticated proxies — e.g. an `HTTPS_PROXY` whose password contains `%` fails `CONNECT` with `407`.

```julia
# HTTP.jl v2.0.0
using HTTP
HTTP._parse_proxy_target("http://user:p%40ss@proxy.example:8080").authorization
# got:      "Basic " * base64encode("user:p%40ss")   # literal %40
# expected: "Basic " * base64encode("user:p@ss")
```

Root cause: [`_userinfo_basic_authorization`](https://github.com/JuliaWeb/HTTP.jl/blob/v2.0.0/src/http_client_url.jl#L185-L190) base64-encodes the raw userinfo substrings. The same function feeds request-URL auth ([L155](https://github.com/JuliaWeb/HTTP.jl/blob/v2.0.0/src/http_client_url.jl#L155), [L370](https://github.com/JuliaWeb/HTTP.jl/blob/v2.0.0/src/http_client_url.jl#L370)) and proxy auth. A PR follows.

mathieu17g:
Fixes #1258

Percent-decode each `userinfo` component via `URIs.unescapeuri` before building the `Basic` credential. No-op for non-encoded userinfo, so existing behavior is preserved. Test added in `test/http_client_proxy_tests.jl` (encoded password, encoded username, request-URL userinfo).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Percent-decode `userinfo` before building the `Basic` auth header (RFC 3986); fixes wrong credentials for request URLs and proxies containing percent-encoded characters.
+
 ## [v2.0.0] - 2026-04-27
 HTTP.jl 2.0 is a major rewrite of the package internals and public API. The
 release keeps the familiar `HTTP.request`, verb helpers, `HTTP.serve`, routing,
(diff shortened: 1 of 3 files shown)


quinnj (approved):
Great catch; thanks for the PR!

Issue #1181, opened 2024-06-06
Websockets max payload limit


bryaan:
There needs to be an option to limit the maximum payload allowed by a websocket server.  This prevents misuse/attacks that could lead to DOS.

quinnj:
As of HTTP.jl 2.0, the WebSocket server supports this via the `maxframesize` keyword:

```julia
HTTP.WebSockets.listen!("0.0.0.0", 8080; maxframesize = 1024 * 1024) do ws  # 1 MiB cap
    for msg in ws
        # ...
    end
end
```

A frame whose payload exceeds `maxframesize` is rejected and the connection is closed with status `1009` ("frame too large"), so an oversized/abusive frame can't be buffered. (`maxfragmentation` similarly bounds reassembled fragmented messages.)

One note: the default is currently unbounded (`typemax(Int)`), so you do need to set `maxframesize` explicitly to get DoS protection. Closing as resolved by 2.0 — whether a safer default is worth adopting can be a separate discussion.

> 🤖 This reply was drafted by an AI assistant and reviewed by @quinnj (maintainer) before posting.

PR #1262, opened 2026-05-30
feat(client): max_decompressed_size guard against decompression bombs

### LINKED ISSUE #1178: gzip / zip bomb
I think it is a good idea to have a similar check implemented by Python's [WebSocket](https://github.com/python-websockets/websockets/pull/407) library, as it is a very easy [attack](https://www.mimecast.com/content/what-is-a-zip-bomb/). Mainly, check that decompressed size does not exceed some kind of limit when executing `HTTP.decode` .

A simple example.

First, generate a gzip file. I lifted code from [this repo](https://github.com/khast3x/flaskbomb/tree/master)
```shell
time dd if=/dev/zero bs=1M count=$((20*1024)) | gzip > ./cake.gzip
```

When I execute the following I observe a jump in the resource usage, eventually leading to a crash of the julia process.
```julia
using HTTP

data = read("cake.gzip")

server = HTTP.serve!() do request::HTTP.Request
   @show request
   @show request.method
   @show HTTP.header(request, "Content-Type")
   @show request.body
   try
       return HTTP.Response(data)
   catch e
       return HTTP.Response(400, "Error: $e")
   end
end

r = HTTP.get("http://127.0.0.1:8081/"; decompress=false)

HTTP.decode(r, "gzip")
```

Happy to provide further details :)

quinnj:
Closes #1178.

Auto-decompression (gzip/deflate) currently has no size limit, so a small compressed payload can inflate to exhaust memory (a decompression/zip bomb). This adds an opt-in `max_decompressed_size` request keyword:

```julia
# reject any response whose decompressed body exceeds 10 MiB
HTTP.get(url; max_decompressed_size = 10 * 1024 * 1024)
# -> throws HTTP.DecompressionLimitError before the body inflates
```

- `0` (default) keeps the current unbounded behavior — fully backward compatible.
- The cap wraps the decompressor stream and is checked on each chunked read, so it fires after at most one extra chunk; memory stays bounded near the limit rather than materializing the whole bomb first.
- Applies to both the default materialized `resp.body` and caller-owned `response_stream` sinks.
- New `HTTP.DecompressionLimitError` (with the offending `limit`) for callers to catch.

Tested with a ~4 MB-of-zeros gzip bomb (a few KB on the wire): rejected under a small limit, succeeds with no/large limit, and the limit is enforced on an `IOBuffer` sink too.

Note: scoped to HTTP response-body decompression (the documented OOM vector). WebSocket frame limits are separate (`maxframesize`, see #1181).

### DIFF
--- a/src/http_client.jl
+++ b/src/http_client.jl
@@ -1076,12 +1076,15 @@ function _status_throws(resp::Response)::Bool
     return resp.status >= 300 && !_is_redirect_status(resp.status)
 end
 
-function _read_all_response_bytes(io::IO)::Vector{UInt8}
+function _read_all_response_bytes(io::IO, limit::Int=0)::Vector{UInt8}
     out = UInt8[]
     buf = Vector{UInt8}(undef, 8192)
+    total = 0
     while true
         n = readbytes!(io, buf, length(buf))
         n == 0 && return out
+        total += n
+        limit > 0 && total > limit && throw(DecompressionLimitError(limit))
         append!(out, @view(buf[1:n]))
     end
 end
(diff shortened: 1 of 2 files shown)


fredrikekre:
Bit of a driveby comment, but it looks like the error may trigger even if the body is not compressed? If so, the name of the kwarg and error is a bit misleading?

quinnj:
> Bit of a driveby comment, but it looks like the error may trigger even if the body is not compressed? If so, the name of the kwarg and error is a bit misleading?

It looks a bit generic, but in the response body consume function, we first do a check if the body doesn't need to be decompressed and then follow a couple different paths that don't involve the new keyword. It's only if we _do_ need to dcompress that we fall through to the with_response_ready path where we pass the keyword along.

Issue #1260, opened 2026-05-30
Make HTTP/2 flow-control window sizes configurable


csvance:
HTTP.jl currently advertises the HTTP/2 protocol-default flow-control window of 65535 bytes for both the per-stream window (`SETTINGS_INITIAL_WINDOW_SIZE`) and the connection-level window, and these are hardcoded at connection setup. Because flow control caps the in-flight unacknowledged bytes, single-stream throughput is bounded by roughly `window / RTT`. This is fine for small requests but throttles any significant upload or download on a link with non-trivial latency.

I hit this building a gRPC server on HTTP.jl 2.0.0, where large protobuf messages are the common case. Raising the advertised window from 64 KiB to 256 KiB nearly doubled client upload throughput in my setup.

It would be useful to configure this when setting up a server or client, since uploads depend on the server's receive window and downloads on the client's. I have an implementation in a fork and am doing more testing and review before opening a PR if this is something the project would be open to.

quinnj:
Yeah, that sounds great; would love a PR.

quinnj:
Implemented in #1264

PR #1264, opened 2026-05-31
HTTP/2 window size args

csvance:
Implements https://github.com/JuliaWeb/HTTP.jl/issues/1260

I first went with a kwargs type approach with each window argument being passed in as `h2_$suffix`, but I think the current approach with a dedicated HTTP2Settings is much cleaner:

```julia
struct HTTP2Settings
    initial_window_size::Int
    connection_window_size::Int
end
```

In the future additional HTTP2 options/tunning parameters can be bundled inside this. This is also mirrors how ProxyConfig works.

I'm still making sure I have all of the window size parameter validation / documentation is correct and testing the new changes end-to-end as part of the server I'm working on. But it would be good to get feedback on the shape of the approach itself.


src/http2.jl
```diff
@@ -17,6 +17,66 @@ const FLAG_PADDED = UInt8(0x8)
...
+const _H2_DEFAULT_MAX_BUFFERED_BYTES = 256 * 1024
+
+# Maximum legal flow-control window (RFC 7540 §6.9.1): 2^31 - 1.
+const _H2_FLOW_CONTROL_MAX_WINDOW = Int64(0x7fff_ffff)
+
+"""
+    HTTP2Settings(; initial_window_size=65535, connection_window_size=65535)
+
+HTTP/2 flow-control configuration shared by [`Server`](@ref), [`Client`](@ref),
+and [`connect_h2!`](@ref).
```
quinnj: This `@ref` breaks the docs because `connect_h2!` is not documented/resolvable. Either add `HTTP.connect_h2!` to the API docs or avoid the `@ref`.

src/http2.jl
```diff
@@ -17,6 +17,66 @@ const FLAG_PADDED = UInt8(0x8)
...
+non-trivial latency, where the default 64 KiB window would otherwise cap a
+transfer at roughly `window / RTT`. The per-stream receive buffer cap is derived
+from `initial_window_size`, so it does not need to be configured separately.
+
+`initial_window_size` may be set below the default to apply tighter per-stream
+backpressure. `connection_window_size` must be at least the protocol default of
+65535: the connection-level window starts at that value and can only be enlarged
+with a `WINDOW_UPDATE`, so a smaller value cannot be advertised.
+"""
+struct HTTP2Settings
```
quinnj, line 54: The generated positional constructor bypasses the validation below (`HTTP2Settings(0, 0)` succeeds). Please use an inner constructor and have the keyword constructor delegate to it.

csvance:
One thing I'm debating is `HTTP2Settings` vs `HTTP2Config`. The later feels more consistent with `ProxyConfig` which is used in the same way.

quinnj:
I think I prefer HTTP2Settings because that's what the protocol tends to refer to them as, right?

csvance:
> I think I prefer HTTP2Settings because that's what the protocol tends to refer to them as, right?

Yes, and window size is configured in the `SETTINGS` frame: https://httpwg.org/specs/rfc7540.html#SETTINGS

Looking through the rest of the fields it seems like there wouldn't be any dissonance with calling it settings. All of these could eventually be user configurable through the same struct. My concern was that maybe there wouldn't be a 1:1 mapping of what the user wants to customize with the underlying `SETTINGS` frame and there was already the existing "Config" convention.

quinnj:
Yeah, that's fair. I don't really feel strongly either way, so up to you and I'm fine with it.

csvance:
I'm not sure about the two failing tests here (windows). In my experience, Github Windows runners cause intermittent failures due to strange hangs where it seems like the runner is starved of resources. Reading the failures I'm getting the impression its not something specific to my changes, especially because the failure has to do with HTTP/1 timeout and all I changed functionally was moving HTTP2Settings validation to the inner constructor.

quinnj:
I rebased your branch here because I put in some windows test hardening fixes the other day; looks like that helped here. Is this ready for a final review?

csvance:
> I rebased your branch here because I put in some windows test hardening fixes the other day; looks like that helped here. Is this ready for a final review?

Going to do a final once over + packet analysis with Wireshark to make sure there isn't anything strange going on.

csvance:
I added a few more tests specifically targeting any pre-existing issues that non default window sizes could have uncovered as well as one that exercises flow control more generally. 

As for my actual changes: without specifying anything non default in HTTP2Settings, behavior is unchanged. Per-stream window is specified in `SETTINGS` when `initial_window_size` is not default. Per connection we get an initial `WINDOW_UPDATE` when `connection_window_size` is not default. Beyond that, my changes build ontop of the existing HTTP/2 window handling. I confirmed in Wireshark what the tests claim to check.

@quinnj Based on this I'm confident my changes are correct and I haven't seen any evidence that they trigger pre-existing bugs or introduce new ones.

csvance:
Rebased onto the latest master. 

EDIT: Looks like the windows issue again, doesn't appear to be related to my changes.

quinnj (approved):
Looks great! Thanks!

Issue #1252, opened 2026-05-29
Precompile never stops with v 2.0.0?


colbec:
I see that there has been a major upgrade to 2.0.0. I'm on Julia 1.12.6 and HTTP takes an inordinately long time to precompile. No error messages are generated. I need to get an application up and running so I went back to julia 1.12.5 and HTTP 1.11 and everything is working so no problems.

Since the precompile process is quite silent,  I wonder if there is a way to increase the verbosity of the process to verify that progress is actually under way? Admittedly this is a big jump and it might be that my application demands are getting too loud.

colbec:
I discovered that pressing 'v' while the precompile is in progtess should provide feedback on the machine state as the precompile proceeds. However when I press 'v' no output is generated.

For further info I am running Fedora 44 fully updated.

quinnj:
Thanks for the report! Precompile workload in 2.0 is much more extensive that previous versions. A couple of things to try/report on:

You can disable precompilation entirely by doing:
```julia
using Preferences
set_preferences!("HTTP", "precompile_workload" => false; force=true)
```
that can at least be an escape hatch for you and more precisely help us know that it is indeed the precompilation workload that is hanging.

Also, how long did you let it run? As I mentioned, it's definitely more extensive than previous versions, and possible it's hitting some extreme compiler case or something where it's technically working, but takes forever.

Can you share the result of `versioninfo()` and `JULIA_NUM_THREADS` please?

also if you know of any firewall, SELinux, or ipv6 restrictions or things that may affect localhost serving/callbacks?

colbec:
Here is version info:

```
julia>` versioninfo()
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 4 × Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, haswell)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 4 virtual cores)

```
JULIA_NUM_THREADS was not defined for my first attempts, so default values.

The longest I let it run was about 20 minutes.

```
julia --threads=4
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
[...]

PR #1274, opened 2026-06-03
Avoid Linux precompile hangs in live workload

quinnj:
## Summary

- use `Base.timedwait` for the HTTP precompile workload watchdog instead of `Reseau.IOPoll.timedwait`
- make the live server/request precompile workload opt-in on Linux with Julia 1.11+
- add `Preferences` as a direct dependency since HTTP now reads its own `precompile_workload` preference

## Context

Issue #1252 is still reproducible for affected Linux users even with `Reseau@1.1.4`. The issue history points at the live HTTP precompile workload: `precompile_workload=false` avoids the hang, and the reports are on Linux Julia 1.11/1.12 while Julia 1.10 eventually completes.

This keeps the workload enabled by default for macOS/Windows and Linux Julia 1.10, but avoids running live local servers/requests during precompile for the Linux Julia 1.11+ lane where users are hanging. Users can still explicitly opt in with `precompile_workload=true`.

## Testing

- `julia --project=. -e 'using HTTP; HTTP._run_precompile_workload!(); println("precompile workload ok")'`
- `julia +1.10.11 --startup-file=no -e 'using Pkg; tmp = mktempdir(); Pkg.activate(tmp); Pkg.develop(path="/tmp/http-1252-fix"); Pkg.precompile(); using HTTP; @show VERSION; @show HTTP._precompile_workload_enabled(); @show Base.timedwait(() -> true, 1.0; pollint=0.01)'`
- `docker run --rm --platform linux/arm64 -v /tmp/http-1252-fix:/work -e JULIA_DEPOT_PATH=/tmp/julia-depot-arm julia:1.12.6 --startup-file=no -e 'using Pkg; Pkg.activate(temp=true); Pkg.develop(path="/work"); Pkg.precompile(); using HTTP; @show Sys.islinux(); @show VERSION; @show HTTP._precompile_workload_enabled()'`
- `julia --project=. -e 'using Pkg; Pkg.test()'`

Co-authored by Codex


quinnj:
Closing this. The workaround shape here is not the right answer for issue #1252; we need to continue root-cause investigation instead of disabling the workload.

PR #1282, opened 2026-06-04
HTTP/2 server: write response body in place, drop per-response body-sized buffer (fixes #1271)

### LINKED ISSUE #1271: HTTP/2 server leaks memory proportional to response body size (HTTP/1.1 does not)
## Summary

Under HTTP/2, the server retains a small amount of memory **on every request, proportional to the response body size**, and never releases it — the growth survives both connection close and a forced `GC.gc()`, so it is live, reachable memory, not heap retention. The identical workload over HTTP/1.1 is completely flat. This makes any long-running HTTP/2 server that returns non-trivial response bodies grow without bound.

We hit this in a production service that returns ~90 KB JSON responses at a few requests/second over a persistent HTTP/2 connection; RSS climbed continuously until restart. Downgrading the same server code to HTTP.jl 1.x eliminated it.

## Environment

- HTTP.jl **2.0.0**
- Reseau **1.1.x**
- Julia **1.11.3** / 1.11.5
- Linux (x86_64)

## Reproduction

Two processes so we can measure **the server's own `gc_live_bytes`** (via a `/mem` endpoint that GCs and reports the server process's live bytes) — this rules out the client as the source.

`server.jl`:

```julia
using HTTP, JSON3
const BIG = Vector{UInt8}(JSON3.write(Dict("col$(c)" => collect(1:200) .+ 0.5 for c in 1:60))) # ~90 KB
handler = function (req::HTTP.Request)
    if req.target == "/mem"
        GC.gc(); GC.gc()
        return HTTP.Response(200; body = string(Base.gc_live_bytes()))
    end
    return HTTP.Response(200, ["Content-Type" => "application/json"]; body = BIG)
end
server = HTTP.serve!(handler, "127.0.0.1", 18080)
println("ready"); wait(server)
```

`client.jl`:

```julia
using HTTP
const PROTO = Symbol(get(ENV, "PROTO", "h2"))    # h2 or h1
client = HTTP.Client()
server_live() = parse(Int, String(HTTP.get(client, "http://127.0.0.1:18080/mem"; protocol=PROTO, retry=false).body))
fire(n) = for _ in 1:n; HTTP.get(client, "http://127.0.0.1:18080/"; protocol=PROTO, retry=false); end
fire(100); base = server_live()
println("protocol=$PROTO baseline=$(round(base/1048576,digits=2))MB")
for r in 1:8
    fire(3000); live = server_live()
    println("  after $(r*3000) reqs: server live=$(round(live/1048576,digits=2))MB  Δ$(round((live-base)/1024))KB")
end
```

Run: `julia server.jl` in one terminal, then `PROTO=h2 julia client.jl` and `PROTO=h1 julia client.jl` in another.

## Results (server process live bytes, after forced GC)

| requests | **h2** server live Δ | **h1** server live Δ |
|---|---|---|
| 3,000  | +992 KB   | 0 KB |
| 6,000  | +1,981 KB | 0 KB |
| 9,000  | +2,970 KB | 0 KB |
| 12,000 | +3,961 KB | 0 KB |
| 15,000 | +4,952 KB | 0 KB |
| 18,000 | +5,940 KB | 0 KB |
| 21,000 | +6,928 KB | 0 KB |
| 24,000 | +7,917 KB | 0 KB |

h2: dead-linear ~330 bytes/request, never recovered. h1 (same server, same body): flat.

## Additional facts from bisecting

- **Scales with response body size.** A tiny (~50 byte) response body leaks negligibly; the 90 KB body leaks ~330 B/request. Suggests per-DATA-frame or per-stream retention rather than full-body retention.
- **Independent of connection lifecycle.** Churning a fresh connection per request leaks at the *same* rate, and the server's `active_conns` set stays bounded (1–2). So it is **not** connection accumulation, and closing/recycling connections does not free it.
- **Survives `GC.gc()`** — it is live memory, measured via `Base.gc_live_bytes()` in the server process.
- **Affects both `serve!` (buffered) and `listen!` (streaming) handlers** identically — so it's in the shared HTTP/2 server/transport path, not the handler layer.
- HTTP/1.1 is unaffected under all of the above.

## Impact

Any long-lived HTTP/2 server returning non-trivial bodies grows unbounded. The only mitigations we found are forcing HTTP/1.1, shrinking responses (proportional slowdown), or periodic restart.

Happy to test patches or provide more diagnostics.

canemon-markov:
## Problem

Fixes #1271. The HTTP/2 server allocates a fresh `Vector{UInt8}` sized to the
(windowed) response body on **every** response and copies the body into it
before writing to the socket. Two paths do this:

- `_write_h2_batch_via_single_buffer!` — the buffered-response fast path
- `_write_data_frames_h2_server!` — streaming / window-overflow batches

On the h2 server's per-stream `Threads.@spawn`, these body-sized per-response
allocations drive sustained memory growth proportional to traffic × body size —
in both retained live heap and process RSS. The HTTP/1 path does **not** show
this: it writes the body in place (`write(io, body)`), with no intermediate copy.

## Fix

Write each DATA frame as a reused 9-byte header followed by the payload slice
taken **directly from the body** via a view. A unit-range view of a
`Vector{UInt8}` or `Base.CodeUnits` (both `DenseVector`) is a stride-1
`StridedVector`, so `write(conn, ::view)` takes Reseau's zero-copy `pointer`
path straight to the socket — matching the HTTP/1 behavior. Per-response
allocation drops from **O(body size)** to **O(1)** (the 9-byte header plus the
small HEADERS-frame bytes).

Accepted server connections enable `TCP_NODELAY`, so splitting the frame header
and payload into separate writes does not introduce Nagle latency.

## Results

Two-process repro: a standalone HTTP/2 server, driven with 100,000 measured
requests of the same response body. Server-process memory sampled via a `/mem`
endpoint after a settled forced GC.

| Scenario                 | Protocol | Requests | Live heap Δ | RSS Δ    | Server proto counts      | Client failures |
|--------------------------|----------|----------|-------------|----------|--------------------------|-----------------|
| Before (HTTP.jl 2.1.0)   | h2       | 100,000  | +20,508 KB  | +28.0 MB | h1=0 h2=100,111 other=0  | 0               |
| After (this PR)          | h2       | 100,000  | −0.0 KB     | −2.9 MB  | h1=0 h2=100,111 other=0  | 0               |

Before, the server retained ~20 MiB of live heap (after settled forced GC) and
+28 MiB RSS over the run. After, both stay flat on the identical HTTP/2-only
workload, with zero client failures.

## Testing

- Full `test/http2_server_tests.jl` passes, including the flow-control / window
  / large-body DATA-frame-splitting testsets that exercise this code.
- Note: the timeout-middleware testset has 2 failures **present on `master`
  independently of this change** (verified against the pristine base) — a
  separate, pre-existing issue, not introduced here.

### DIFF
--- a/src/http2_server.jl
+++ b/src/http2_server.jl
@@ -730,55 +730,68 @@ function _write_data_frames_h2_server!(
     end
     max_frame_size = min(Int(_H2_SERVER_MAX_DATA_FRAME_SIZE), peer_max)
     write_deadline_ns > 0 && _set_write_deadline!(conn, write_deadline_ns)
-    # Each "batch" drains as much of the body as the peer's current send
-    # window allows, builds it into a single contiguous byte buffer
-    # (interleaved 9-byte DATA-frame headers + payload slices), and writes
-    # the whole batch to the socket under one lock acquisition. The previous
-    # implementation paid one IOBuffer allocation, one frame-header
-    # allocation, and one socket write per frame; this version pays one
-    # allocation and one socket write per *batch*.
+    # Each "batch" drains as much of the body as the peer's current send window
+    # allows and emits it as DATA frames. Each frame is written as a 9-byte
+    # header (from a single reused scratch buffer) followed by the payload slice
+    # taken DIRECTLY from `data` (zero-copy: a unit-range view of a Vector{UInt8}
+    # or Base.CodeUnits is a stride-1 StridedVector, so `write(conn, ::view)`
+    # passes the underlying pointer straight to the socket — exactly like the
+    # HTTP/1 body path).
+    #
+    # The previous implementation allocated a fresh Vector sized to the whole
+    # windowed body on EVERY response and copied the body into it. Under HTTP/2's
+    # per-stream `@spawn` those large, short-lived allocations are retained by the
+    # glibc malloc per-thread arenas and never returned to the OS, so process RSS
+    # climbs in proportion to response body size even though the Julia heap stays
+    # flat. Writing the body in place removes that per-response allocation; the
+    # only buffer here is the reused 9-byte frame header.
+    header = Vector{UInt8}(undef, 9)
     while offset <= total_len
         # Probe the current allowed window with the size remaining.
         # Reservation is bounded by stream/conn windows only; per-frame
         # max-frame-size is applied below when slicing the reservation
-        # into DATA frames.
+        # into DATA frames. Reservation may block on a peer WINDOW_UPDATE,
+        # so it must happen OUTSIDE write_lock.
         first_chunk = _reserve_h2_send_window!(send_state, stream_id, total_len - offset + 1, write_deadline_ns)
         first_chunk <= 0 && continue
-        # Pre-size an output Vector that fits this batch exactly.
-        n_frames_first = cld(first_chunk, max_frame_size)
-        batch_len = first_chunk + 9 * n_frames_first
-        out = Vector{UInt8}(undef, batch_len)
-        out_pos = 1
-        # Emit frames covering this reservation.
-        remaining_in_res = first_chunk
-        while remaining_in_res > 0
-            payload = min(remaining_in_res, max_frame_size, total_len - offset + 1)
-            final_chunk = (offset + payload - 1) == total_len
-            _stamp_h2_data_header!(out, out_pos, payload, end_stream && final_chunk, stream_id)
-            out_pos += 9
-            _copy_h2_payload!(out, out_pos, data, offset, payload)
-            out_pos += payload
-            offset += payload
-            remaining_in_res -= payload
-        end
-        # IMPORTANT: write the bytes we have in `out` BEFORE attempting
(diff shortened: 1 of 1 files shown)


quinnj:
Thanks for digging into this. One tradeoff I want to think through before merging: this extends the v2.1.0 `BytesBody` fix down into the lower DATA-frame writers, which is good for avoiding the body-sized copy, but it also changes the write shape from one contiguous buffered write per DATA batch into separate writes for the frame header and payload.

For TLS, I think that may mean more TLS application writes/records; for plain TCP, it likely means more syscalls. A benchmark that would help: compare `master`/v2.1.0 vs this branch for a large fixed `Vector{UInt8}` response over both h2c and h2/TLS, tracking RSS/live heap plus req/s/latency, and ideally write/syscall counts. That would tell us whether the RSS win comes with a meaningful throughput/latency cost.

quinnj:
Closing as stale; feel free to push work/re-comment if able and I can open it back up

PR #1284, opened 2026-06-05
docs v2.1: note that a manually-set Cookie header is no longer merged with the cookie jar

### LINKED ISSUE #1283: Docs v2.1: manually-set Cookie header no longer merged with the cookie jar
In 1.x (verified through v1.11.0, the last 1.x release), a manually-set `Cookie` request header was merged with the cookie jar. In v2.1.0 the `Cookie` header is built only from the jar (plus the `cookies` keyword) and overwrites any `Cookie` entry passed in `headers` once the jar holds a cookie. This behavior change can break 1.x code and isn't mentioned in the [migration guide](https://github.com/JuliaWeb/HTTP.jl/blob/master/docs/src/guides/migration-1x.md).

1.x — `src/clientlayers/CookieRequest.jl` (v1.11.0) merged with the existing header:

```julia
existingcookie = header(req.headers, "Cookie")
setheader(req.headers, "Cookie" => stringify(existingcookie, cookiestosend))
```

v2.1.0 — `src/http_client.jl` builds from the jar and replaces it:

```julia
cookie_value = _cookie_header(cookiejar, cookies, current_secure, host, path)
cookie_value === nothing || setheader(send_request.headers, "Cookie", cookie_value)
```

With the default `cookies=true`, a hand-built `Cookie` header is sent on the first request (empty jar), but once the server returns a `Set-Cookie` the jar takes over and the manual header is dropped on later requests. Against a server that rotates its session cookie (a `Set-Cookie` on each authenticated response), a static `Cookie` header then works for only one request and the next is rejected; using an `HTTP.CookieJar` instead behaves as in 1.x.

Could a short note be added to `migration-1x.md` (recommended approaches: a `CookieJar`, `cookies = Dict(...)`, or `cookies = false` for a raw header)? Happy to follow up with a PR.

mathieu17g:
Documents the cookie header behavior change reported in #1283: in 1.x a manually-set `Cookie` header was merged with the jar, whereas now it is built solely from the jar and overwritten once the jar holds a cookie. Adds a short "Cookies" section to `migration-1x.md` with the recommended approaches.

Closes #1283.

---

**Optional follow-up (your call — not in this PR): surface the silent overwrite with a `@warn`.**

A user-supplied `Cookie` header is currently dropped without notice once the jar is non-empty. A `@warn` would make it visible. Rough sketch in `src/http_client.jl` (the cookie layer that sets the header):

```julia
# capture whether the *caller* supplied a Cookie header, before the redirect/retry loop
user_cookie = HTTP.hasheader(request.headers, "Cookie")   # exact var/scope may differ
...
cookie_value = _cookie_header(cookiejar, cookies, current_secure, host, path)
if cookie_value !== nothing
    user_cookie && @warn "Manually-set `Cookie` header is overridden by the cookie jar; \
        pass cookies via a `CookieJar`/the `cookies` keyword, or use `cookies=false` to send a raw header." maxlog=1
    setheader(send_request.headers, "Cookie", cookie_value)
end
```

Caveat: it must distinguish a caller-supplied header from one carried across redirects/retries (hence capturing `user_cookie` once before the loop). Happy to turn this into a proper change — or add a `cookies=:manual`-style opt-out — if you think it's worth it.

### DIFF
--- a/docs/src/guides/migration-1x.md
+++ b/docs/src/guides/migration-1x.md
@@ -187,6 +187,24 @@ Useful helpers include `HTTP.header`, `HTTP.headers`, `HTTP.hasheader`,
 `HTTP.headercontains`, `HTTP.setheader`, `HTTP.appendheader`, and
 `HTTP.removeheader`.
 
+## Cookies
+
+In 1.x, a manually-set `Cookie` request header was merged with the cookie jar, so hand-set
+cookies were preserved and augmented with jar / `Set-Cookie` state.
+
+Now, the `Cookie` header is built solely from the jar (plus the `cookies` keyword); a
+`Cookie` entry placed in `headers` is overwritten once the jar holds any cookie (i.e. after
+the first `Set-Cookie`). To send specific cookies:
+
+- use a `CookieJar` (e.g. via a `Client`) — recommended for multi-request sessions,
+  especially when the server **rotates** session cookies (a `Set-Cookie` on each response);
+- pass `cookies = Dict("name" => "value", ...)`;
+- or set `cookies = false` to send a raw `Cookie` header without jar management
+  (no `Set-Cookie` tracking).
+
+A static, hand-built `Cookie` header against a session-rotating server therefore succeeds
+once and is then superseded by the jar; prefer a `CookieJar` so rotation is tracked.
+
 ## Request Context
 
 In 1.x, middleware often treated request context as a plain dictionary. In 2.0,


quinnj:
I think I'd actually prefer to restore hte 1.x behavior here and merge the two; would you prefer that?

mathieu17g:
Ok, I'm looking into it, weighing if I can propose something in between v1.x (iso curl) and v2.1 (iso Python requests) behaviours

mathieu17g:
I have proposed a restoration of the merge between manual cookies and cookies from a cookie jar in another PR.

I close this PR

Issue #1070, opened 2023-06-29
Setting any readtimeout on WebSocket client causes TimeoutError


bryaan:
- HTTP.jl 1.9.7

Setting any readtimeout other than 0 on a websocket client causes a TimeoutError where it doesn't consider whether any messages have been received.

My understanding is it should fail when it can't read on the IO object for N seconds.

gryumov:
Faced with such a problem

```julia
julia> using HTTP

julia> HTTP.WebSockets.open(
           "wss://stream.binance.com:9443/stream?streams=adausdt@depth5@100ms", 
           readtimeout = 10
       ) do ws
           while true
               HTTP.WebSockets.receive(ws) |> String |> println
           end
       end
{"stream":"adausdt@depth5@100ms","data":{"lastUpdateId":7512315159,"bids":[["0.30070000","70165.90000000"],["0.30060000","82422.60000000"],["0.30050000","117418.90000000"],["0.30040000","283804.50000000"],["0.30030000","160890.30000000"]],"asks":[["0.30080000","18922.50000000"],["0.30090000","166428.60000000"],["0.30100000","179508.70000000"],["0.30110000","72360.00000000"],["0.30120000","323729.90000000"]]}}
{"stream":"adausdt@depth5@100ms","data":{"lastUpdateId":7512315161,"bids":[["0.30070000","70165.90000000"],["0.30060000","82422.60000000"],["0.30050000","117418.90000000"],["0.30040000","284112.50000000"],["0.30030000","160890.30000000"]],"asks":[["0.30080000","19077.50000000"],["0.30090000","166428.60000000"],["0.30100000","179508.70000000"],["0.30110000","72360.00000000"],["0.30120000","323729.90000000"]]}}
...
{"stream":"adausdt@depth5@100ms","data":{"lastUpdateId":7512316126,"bids":[["0.30070000","79480.40000000"],["0.30060000","89375.40000000"],["0.30050000","123721.60000000"],["0.30040000","266929.20000000"],["0.30030000","177536.10000000"]],"asks":[["0.30080000","16040.00000000"],["0.30090000","167574.90000000"],["0.30100000","228828.70000000"],["0.30110000","200246.10000000"],["0.30120000","324346.60000000"]]}}
[...]

PR #1286, opened 2026-06-05
Wire WebSocket read_idle_timeout into the client read loop (#1062)

### LINKED ISSUE #1062: Make Websockets readtimeout apply to any message received
It seems that at present when specifying readtimeout on a websocket client open() that it only works when having sent a message first.  But sometimes the websocket server doesn't handle any messages, it only sends.

My Proposal: 
The timeout timer should be reset when any message is received on the websocket.  This works when the websocket server doesn't expect any messages.

quinnj:
Fixes #1062.

## Problem
`HTTP.WebSockets.open` accepts a `read_idle_timeout` keyword, but for a live session it was a **silent no-op**: the connection's read deadline is cleared right after the handshake, so the read loop never enforced it. A WebSocket that goes silent (server only pushes, then stalls) would block `receive` forever — and #1062 (and #1070's discussion) specifically asked for a timeout that **resets on each received message**.

## Fix
The read loop re-arms the underlying Reseau stream's read deadline before each read when `read_idle_timeout > 0` (mirroring how the HTTP response-body path implements `read_idle_timeout`). So the timeout:
- **resets on every frame** → never interrupts an active stream, and
- **fires after `read_idle_timeout` s of silence** → surfaces a stalled connection as `WebSocketError(1006, "websocket read idle timeout")`.

Default `read_idle_timeout = 0` is unchanged (no timeout). Server- and upgrade-side WebSockets are unaffected (they pass the `0` default).

## Tests
Added an integration test: a client with `read_idle_timeout = 1.0` receives a message during activity (no spurious timeout) and then surfaces the idle timeout after silence — verified reliable across repeated runs. Full WS suite green locally (client 18, server 31, codec 88, integration 42).

### DIFF
--- a/src/http_websockets.jl
+++ b/src/http_websockets.jl
@@ -56,8 +56,10 @@ import .._request_deadline_ns
 import .._request_connect_host_resolver
 import .._request_connect_phase_deadline_ns
 import .._request_connect_phase_timeout_ns
+import .._request_read_idle_timeout_ns
 import .._request_response_header_deadline_ns
 import .._request_write_deadline_ns
+import .._phase_deadline_ns
 import .._resolve_request_timeout_settings
 import .._apply_request_timeout_settings!
 import ..get_request_context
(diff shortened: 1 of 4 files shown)


src/http_websockets.jl
```diff
@@ -383,11 +386,21 @@ function _process_incoming_frame!(ws::WebSocket, frame::WsFrame)::Nothing
     return nothing
 end
 
+# Set the read deadline directly on the underlying Reseau stream so an opt-in
+# WebSocket read idle timeout can be re-armed before each read.
+_ws_arm_read_deadline!(stream::TLS.Conn, deadline_ns::Int64) = TLS.set_read_deadline!(stream, deadline_ns)
+_ws_arm_read_deadline!(stream::TCP.Conn, deadline_ns::Int64) = TCP.set_read_deadline!(stream, deadline_ns)
```
quinnj: collapse into one method

```suggestion
_ws_arm_read_deadline!(stream, deadline_ns::Int64) = TLS.set_read_deadline!(stream, deadline_ns)
```

PR #1287, opened 2026-06-07
Restore merging a manual Cookie header with the jar (de-duped by name)

### LINKED ISSUE #1283: Docs v2.1: manually-set Cookie header no longer merged with the cookie jar

mathieu17g:
Following @quinnj     https://github.com/JuliaWeb/HTTP.jl/pull/1284#issuecomment-4633592674, this PR restores the merge — but de-duplicated by name, which lands between 1.x and 2.1:

| | Behaviour | ≈ | Problem |
|---|---|---|---|
| **1.x** | merge the manual `Cookie` header with the jar | curl | can emit a name twice — [curl warns](https://curl.se/libcurl/c/CURLOPT_COOKIE.html) the doubled header may get the request rejected, and RFC 6265 §5.4 leaves the server's pick among same-name cookies undefined |
| **2.1** | build the header from the jar only, overwrite the manual one | Python `requests` | the manual header is silently dropped whenever the jar / `cookies=` dict is non-empty |
| **this PR** | merge, **de-duped by name** | — | neither: not dropped, never doubled |

`_cookie_header` currently does `stringify("", merged)`, dropping the request's existing `Cookie`. The patch re-reads it and appends only the names the jar / `cookies=` dict don't already cover — managed cookies win on a clash. Besides keeping a rotated jar cookie from being shadowed by a stale manual one, managed-wins preserves path-scoped same-name cookies held in the jar (e.g. a server that sets `JSESSIONID` on both `/` and `/app` keeps both, instead of a manual `JSESSIONID` collapsing them). `cookies=false` still sends a manual header verbatim.

Covers the HTTP request path and the websocket handshake, and documents the `cookies` keyword; tests in `test/http_cookie_tests.jl`. Closes #1283.

Open call: managed-wins vs manual-wins on a name clash — I went managed-wins for rotation safety; trivial to flip if you'd rather an explicit header win.

### DIFF
--- a/src/http_client.jl
+++ b/src/http_client.jl
@@ -587,6 +587,7 @@ function _cookie_header(
     secure::Bool,
     host::String,
     path::String,
+    manual::Vector{Cookie}=Cookie[],
 )::Union{Nothing,String}
     cookies === false && return nothing
     merged = Cookie[]
@@ -596,6 +597,22 @@ function _cookie_header(
     end
     if cookies !== true
         append!(merged, cookies::Vector{Cookie})
+    end
+    # Merge a caller-set `Cookie` request header with the managed cookies (this restores
+    # the pre-2.0 behaviour, where a manual header was not silently dropped), but
+    # de-duplicate by name so the wire header never carries the same name twice. Managed
+    # cookies (the jar plus any `cookies=` dict) win; the manual header only contributes
+    # names they do not already cover. This keeps a rotated jar cookie from being
+    # shadowed by a stale manual one and preserves path-scoped same-name cookies held in
+    # the jar. Use `cookies=false` to send a manual `Cookie` header verbatim.
+    if !isempty(manual)
+        managed = Set(c.name for c in merged)
+        for c in manual
+            if !(c.name in managed)
+                push!(merged, c)
+                push!(managed, c.name)
+            end
+        end
     end
     isempty(merged) && return nothing
     return stringify("", merged)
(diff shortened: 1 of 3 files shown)


mathieu17g:
Failing CI for Julia 1 - windows-latest, probably an unrelated flaky Windows server test (http_server_http1_tests.jl:653), re-running via an amend + force-push to see

quinnj (approved):

Issue #1288, opened 2026-06-08
add methods to deal with `Header`s


hhaensel:
I'm currently adapting Genie/Stipple to v2 of HTTP and found it would be very helpful to define
```julia
function Base.setindex!(headers::Headers, value, key::Union{Symbol, AbstractString})
    item = _header_pair(key, value)
    k = first(item)
    res = findfirst(x -> first(x) == k, headers.entries)
    if res !== nothing
        headers[res] = item
    else
        push!(headers.entries, item)
    end
end

function Base.merge!(headers::Headers, items)
    for (k,v) in items
        headers[k] = v
    end
    return headers
end
```
Moreover, I'd appreciate

```julia
    elseif headers_input isa Pair
        _append_header_values!(headers, headers_input.first, headers_input.second)
        return headers
    end
```
in `mkheaders()`
and maybe
```julia
Headers(item::Pair) = mkheaders((item,))
Headers(items...) = mkheaders(items)
```
in order to support the more natural
```julia
Headers(:a => :b)
Headers(:a => :b, :c => "d")
```
What do you think?

hhaensel:
After some more thinking I'd propose to simply define
```julia
mkheaders(items...; kwargs...) = mkheaders(Base.Iterators.flatten((items, kwargs)))
Headers(items...; kwargs...) = mkheaders(items...; kwargs...)
```
let me open a PR with this...

quinnj:
Implemented in #1290

PR #1290, opened 2026-06-09
add methods for Headers construction and treatment

hhaensel:
This PR refers to #1288 
Happy to receive comments

### DIFF
--- a/src/http_core.jl
+++ b/src/http_core.jl
@@ -614,6 +614,13 @@ Headers(items::AbstractDict) = mkheaders(items)
 Headers(items::AbstractVector) = mkheaders(items)
 Headers(items::Tuple) = mkheaders(items)
 
+"""
+    Headers(items...; kwargs...) -> Headers
+
+Construct a canonicalized Headers from items and/or kwargs
+"""
+Headers(items...; kwargs...) = mkheaders(items...; kwargs...)
+
 Base.copy(headers::Headers) = Headers(headers)
 
 Base.IndexStyle(::Type{Headers}) = IndexLinear()
@@ -643,6 +650,15 @@ function Base.setindex!(headers::Headers, item, i::Int)
     return headers.entries[i]
 end
 
+function Base.setindex!(headers::Headers, value, key::Union{Symbol, AbstractString})
+    setheader(headers, String(key) => String(value))
+end
+
+function Base.append!(headers::Headers, items)
+    union!(append!(headers.entries, Headers(items)))
+    return headers
+end
+
 function Base.push!(headers::Headers, item)
     push!(headers.entries, _header_pair(first(item), last(item)))
     return headers
@@ -664,7 +680,7 @@ end
 
 Normalize a header-like input into a mutable `Headers` collection.
 
-`headers_input` may be `nothing`, an existing `Headers`, a dictionary, or an
+`headers_input` may be `nothing`, a pair or 2-tuple, an existing `Headers`, a dictionary, or an
 iterable of `Pair`s/2-tuples. Vector-valued entries are expanded into repeated
 header values using the same merge rules as `appendheader`.
 """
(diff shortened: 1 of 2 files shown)


src/http_core.jl
```diff
@@ -643,6 +650,15 @@ function Base.setindex!(headers::Headers, item, i::Int)
     return headers.entries[i]
 end
 
+function Base.setindex!(headers::Headers, value, key::Union{Symbol, AbstractString})
+    setheader(headers, String(key) => String(value))
+end
+
+function Base.append!(headers::Headers, items)
+    union!(append!(headers.entries, Headers(items)))
```
quinnj: This single-argument `union!` is a no-op — drop it, or if the intent was de-duplication: as written, `append!` duplicates keys that already exist (legal for repeated headers, but #1288 asked for `merge!`/replace semantics, which is probably what you want for e.g. `Content-Type`). Worth deciding which semantic this method should have and adding an overlapping-key test.

Unrelated: the Windows CI failure looks like our known flaky server test — a re-run should clear it.

src/http_core.jl
```diff
@@ -614,6 +614,13 @@ Headers(items::AbstractDict) = mkheaders(items)
 Headers(items::AbstractVector) = mkheaders(items)
 Headers(items::Tuple) = mkheaders(items)
 
+"""
+    Headers(items...; kwargs...) -> Headers
+
+Construct a canonicalized Headers from items and/or kwargs
+"""
+Headers(items...; kwargs...) = mkheaders(items...; kwargs...)
```
quinnj: An untyped splat makes `Headers(<anything>...)` dispatchable, so misuse (e.g. `Headers("foo", 1)`) surfaces as confusing iterator errors downstream. Suggest constraining the items — e.g. `Headers(items::Union{Pair,Tuple}...; kwargs...)` — so bad input fails with a clear MethodError instead.

src/http_core.jl
```diff
@@ -643,6 +650,15 @@ function Base.setindex!(headers::Headers, item, i::Int)
     return headers.entries[i]
 end
 
+function Base.setindex!(headers::Headers, value, key::Union{Symbol, AbstractString})
+    setheader(headers, String(key) => String(value))
```
quinnj: `String(value)` throws for non-string values (`headers["X-Count"] = 5` → MethodError). `string(value)` would match how dict-style assignment is usually expected to behave.

src/http_core.jl
hhaensel: When I filed the issue, I thought merge! would be a good idea, but then I got aware that headers support multiple entries with the same key, so I decided to go for an append! instead.
Maybe this would be better:
```julia
function Base.merge!(headers::Headers, items)
    for (key, value) in Headers(items)
        if key == "Set-Cookie"
            push!(headers.entries, key => value)
        else
            setheader(headers, key => value)
        end
    end
    return headers
end

function Base.append!(headers::Headers, items)
    for (key, value) in Headers(items)
        appendheader(headers, key => value)
    end
    return headers
end
```

src/http_core.jl
hhaensel: that's what I might have done, but I found
```julia
 function _append_header_values!(headers::Headers, key, value)
    if value isa AbstractVector && !(value isa AbstractString)
        for item in value
            appendheader(headers, String(key) => String(item))
        end
        return headers
    end
    appendheader(headers, String(key) => String(value))
    return headers
end
```
some lines below. I could change all for consistency

src/http_core.jl
hhaensel: Sounds reasonable, I will do that

src/http_core.jl
hhaensel: Re-reading I think I understood that only `String(value)` should be changed to `string(value)`.
Agree!

quinnj (approved):
LGTM, thanks!

PR #1291, opened 2026-06-09
Add support for multipart/mixed content type

### LINKED ISSUE #1101: Feature Request: Add handling of multipart/mixed messages
Currently, only multipart/form-data is supported by multipart handling.

While this is probably the most important use case, also multipart/mixed is quite common, e.g. in REST APIs.
I recently struggled with batch requesting a SharepointOnline list and finally found my way through it. I thought, I could share my experience with the community and propose this PR #1100

hhaensel:
This PR is a port of the older #1100 with the help of Claude 4.5
It fixes #1101.

This commit adds comprehensive support for multipart/mixed content, commonly used for batch API requests (e.g., SharePoint, GraphQL).

Core functionality:
- Added `type` field to Form struct to support :formdata and :mixed
- Form constructor accepts type parameter (defaults to :formdata)
- Form(Vector{Multipart}) constructor for multipart/mixed
- Updated content_type() to return correct MIME type based on form type

Parsing enhancements:
- Modified parsing to optionally skip Content-Disposition headers (not required for multipart/mixed per RFC2046)
- Added parse_multipart() as generic parser for any multipart type
- Added parse_multipart_mixed() convenience function
- Support for type filtering in parsing functions

API additions:
- Batch() convenience function for semantic batch request creation
- Exported: parse_multipart, parse_multipart_mixed, Batch

Testing:
- Comprehensive test suite in test/http_multipart_mixed.jl
- 41 tests covering creation, parsing, round-trip, and edge cases

The implementation is adapted from the original work in commit 573fb6238574fa58108517e0dc8b46ec32f70511 but updated to work with HTTP v2 architecture where multipart functionality is consolidated in http_forms.jl.


hhaensel:
@quinnj Just found some possible improvements, please don't merge yet

hhaensel:
@quinnj just review as is! The idea I had was not so good...

src/http_forms.jl
```diff
@@ -110,16 +115,16 @@ function Form(d; boundary=_default_form_boundary())
     write(io, "\r\n--" * boundary * "--" * "\r\n")
     seekstart(io)
     push!(data, io)
-    return Form(data, 1, -1, boundary)
+    return Form(data, 1, -1, boundary, type)
 end
 
-function writemultipartheader(io::IOBuffer, stream::IOStream)
+function writemultipartheader(io::IOBuffer, stream::IOStream, type::Symbol=:formdata)
```
quinnj, line 125: For `type = :mixed` this still writes `; filename="..."` — but the `Content-Disposition:` prefix it continues was skipped above, so a `:mixed` form holding an `IOStream` part serializes a malformed header line (and the generic `IO` method below injects an extra blank line into the part body). The `Batch`/`Multipart` path is correct — these two raw-IO paths should either honor `type == :mixed` like the `Multipart` method does, or reject raw `IO` parts under `:mixed`.

src/http_forms.jl
```diff
@@ -86,7 +90,8 @@ function _default_form_boundary()::String
     return replace(string(UUIDs.uuid4()), '-' => "")
 end
 
-function Form(d; boundary=_default_form_boundary())
+function Form(d; boundary=_default_form_boundary(), type::Symbol=:formdata)
+    @assert type in (:formdata, :mixed) "type must be either :formdata or :mixed"
```
quinnj: Argument validation shouldn't rely on `@assert` (asserts may be disabled); prefer `type in (:formdata, :mixed) || throw(ArgumentError("type must be :formdata or :mixed"))`.

src/http_forms.jl
```diff
@@ -1,7 +1,10 @@
 export Form
 export Multipart
+export Batch
```
quinnj: I'd rather not export `Batch` — it's a very generic name to claim from every `using HTTP`. Qualified `HTTP.Batch(...)` (or just `Form(parts; type=:mixed)`) reads fine.

src/http_forms.jl
hhaensel: I modified the function accordingly, it's been updated in the last commit

src/http_forms.jl
hhaensel: I had fixed that and realised only after pushing and testing that I had also fixed two other instances of `@assert` some lines below. That's why currently the tests fail. They check for AssertionError and now they throw ArgumentError. Should I adapt the tests?

hhaensel:
@quinnj just need feedback on my two open conversations.
I tend to revert the `@assert` change, it seems that `@assert` is quite common in HTTP

src/http_forms.jl
quinnj: I think I'd prefer ArgumentErrors for these.

hhaensel: Done

hhaensel:
Don't merge yet, I'm still adding an example ...

hhaensel:
I've added a multibatch example including server. It turned out that the parsing was needed for both Requests and Responses, so I decided to unify http_forms.jl accordingly. Here's what I did in detail

- Refactor to use Union{Request,Response} pattern (matches http_core.jl style)
- Rename helper `_request_body_bytes()` to `_message_body_bytes()
- Add comprehensive Response parsing tests
- Include batch server example demonstrating multipart/mixed usage
- Update docstrings with jldoctest examples

This is all fully backward compatible with existing Request API

hhaensel:
@quinnj ready for final review :-)

hhaensel:
Found a bug in HTTP 2.3.0, it references VERSION in HTTP's scope for the public declarations. I fixed it here and also filed #1313

quinnj (approved):

PR #1292, opened 2026-06-10
fix missing characters in normal_url_char

hhaensel:
Just received an error during cookie processing. It seems that a line of characters was missing.
I completed the array with the help of Claude

### DIFF
--- a/src/http_cookies.jl
+++ b/src/http_cookies.jl
@@ -462,7 +462,7 @@ const normal_url_char = Bool[
     true, true, true, true, true, true, true, false,         # 120-127: x y z { | } ~ DEL
 ]
 
-@inline isurlchar(c) = c > '\u80' ? true : normal_url_char[Int(c)+1]
+@inline isurlchar(c) = c ≥ '\u80' ? true : normal_url_char[Int(c)+1]
 
 """
     CookieJar()
--- a/src/http_cookies.jl
+++ b/src/http_cookies.jl
@@ -458,8 +458,8 @@ const normal_url_char = Bool[
     true, true, true, true, true, true, true, true,          # 88-95: X Y Z [ \ ] ^ _
     true, true, true, true, true, true, true, true,          # 96-103: ` a b c d e f g
     true, true, true, true, true, true, true, true,          # 104-111: h i j k l m n o
-    true, true, true, true, true, true, true, false,         # 112-119: p q r s t u v w
-    true, true, true, false, false, false, true, false,      # 120-127: x y z { | } ~ DEL
+    true, true, true, true, true, true, true, true,          # 112-119: p q r s t u v w
+    true, true, true, true, true, true, true, false,         # 120-127: x y z { | } ~ DEL
 ]
 
 @inline isurlchar(c) = c > '\u80' ? true : normal_url_char[Int(c)+1]
--- a/src/http_cookies.jl
+++ b/src/http_cookies.jl
@@ -444,21 +444,22 @@ end
 sanitizeCookiePath(v) = filter(validcookiepathbyte, v)
 
 const normal_url_char = Bool[
-    false, false, false, false, false, false, false, false,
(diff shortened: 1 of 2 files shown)


src/http_cookies.jl
```diff
@@ -444,21 +444,22 @@ end
...
+    true, true, true, true, true, true, true, true,          # 48-55: 0-7
+    true, true, true, true, true, true, true, false,         # 56-63: 8 9 : ; < = > ?
+    true, true, true, true, true, true, true, true,          # 64-71: @ A B C D E F G
+    true, true, true, true, true, true, true, true,          # 72-79: H I J K L M N O
+    true, true, true, true, true, true, true, true,          # 80-87: P Q R S T U V W
+    true, true, true, true, true, true, true, true,          # 88-95: X Y Z [ \ ] ^ _
+    true, true, true, true, true, true, true, true,          # 96-103: ` a b c d e f g
+    true, true, true, true, true, true, true, true,          # 104-111: h i j k l m n o
+    true, true, true, true, true, true, true, false,         # 112-119: p q r s t u v w
+    true, true, true, false, false, false, true, false,      # 120-127: x y z { | } ~ DEL
```
quinnj: Confirmed the underlying bug — on master, any `Set-Cookie` whose name starts with `x`…`~` throws `BoundsError` (the table has 120 entries; `isurlchar` indexes up to 128) — so thanks for catching this.

But these last two rows don't match the 1.x table this file was ported from (v1.10.19 `src/cookies.jl`): the port dropped 1.x's all-`true` 112–119 row, shifting the old final row up. Per 1.x, 112–119 (`p`–`w`) should be **all `true`**, and 120–127 should be **`true ×7, false`** — only DEL is `false`. As written here, `w`, `{`, `|`, `}` are `false`, so a cookie named e.g. `w` is silently dropped by `iscookienamevalid`.

Could you sync both rows to the 1.x values and add a regression test? e.g. `setcookies!` with `Set-Cookie: xsrf_token=1` round-trips, plus `isurlchar('w') && isurlchar('{') && isurlchar('|') && isurlchar('}') && !isurlchar('\x7f')`.

hhaensel: Added respective fix

hhaensel:
merge conflicts solved, ready to merge

hhaensel:
Sorry for the messy commits, I can either prepare a new PR or you do a squash merge

src/http_cookies.jl
```diff
@@ -444,21 +444,22 @@ end
...
+    true, true, true, true, true, true, true, true,          # 72-79: H I J K L M N O
+    true, true, true, true, true, true, true, true,          # 80-87: P Q R S T U V W
+    true, true, true, true, true, true, true, true,          # 88-95: X Y Z [ \ ] ^ _
+    true, true, true, true, true, true, true, true,          # 96-103: ` a b c d e f g
+    true, true, true, true, true, true, true, true,          # 104-111: h i j k l m n o
+    true, true, true, true, true, true, true, true,          # 112-119: p q r s t u v w
+    true, true, true, true, true, true, true, false,         # 120-127: x y z { | } ~ DEL
 ]
 
 @inline isurlchar(c) = c > '\u80' ? true : normal_url_char[Int(c)+1]
```
quinnj: [findings by codex]

This still leaves one BoundsError boundary: `normal_url_char` now covers `0x00:0x7f`, but `isurlchar` only bypasses the table for `c > '\u80'`. Exactly `Char(0x80)` still indexes `normal_url_char[129]` and throws; e.g. `HTTP.Cookies.iscookienamevalid(string(Char(0x80)))` errors while `Char(0x81)` returns `true`. Since this PR is closing a BoundsError in this helper, can we make the guard `c >= '\u80'` and add that boundary to the regression?

hhaensel: sure, will do

PR #1296, opened 2026-06-10
Avoid write(::IO, ::String) invalidations from Stream and IO

SimonDanisch:
Extending Base.write for String/AbstractString on a new IO type invalidates every abstractly-inferred write(::IO, ::String) call site in the ecosystem (~1380 instances each at using time, measured with @snoop_invalidations). Hook Base.unsafe_write instead: Base's generic string write funnels through it, so behavior should be unchanged.
ps: uncovered in claude code compile time investigation


quinnj:
Yeah, this is fine/good; but we need to define `Base.write(stream::Stream, b::UInt8)` as well so other `AbstractString`s don't throw. If we can add that, this should be good to merge.

quinnj (approved):

PR #1298, opened 2026-06-10
HTTP/2: retry streams reset with REFUSED_STREAM or rejected by GOAWAY (RFC 9113 §8.7)

### LINKED ISSUE #1294: HTTP/2: streams reset with REFUSED_STREAM or rejected by GOAWAY are never retried (RFC 9113 §8.7)
With `retry=true`, a stream reset by the server surfaces to the caller as `ProtocolError("HTTP/2 stream reset by peer")` (v2.1.1):

- the `RSTStreamFrame` handler does not read the frame's error code ([`src/http2_client.jl:761-766`](https://github.com/JuliaWeb/HTTP.jl/blob/v2.1.1/src/http2_client.jl#L761-L766)) — `REFUSED_STREAM` is indistinguishable from fatal codes;
- `_retryable_request_error` ([`src/http_client_retry.jl:48-70`](https://github.com/JuliaWeb/HTTP.jl/blob/v2.1.1/src/http_client_retry.jl#L48-L70)) classifies neither `ProtocolError` nor `H2GoAwayError` as retryable, so refused streams and in-flight streams rejected by GOAWAY (`stream_id > last_stream_id`) are never replayed.

Both cases are guaranteed unprocessed, hence safe to retry even for non-idempotent requests — [RFC 9113 §8.7](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.7): *"The REFUSED_STREAM error code can be included in a RST_STREAM frame to indicate that the stream is being closed prior to any processing having occurred. Any request that was sent on the reset stream can be safely retried."* Mainstream h2 clients all retry them:

| Client | Behaviour |
|---|---|
| Go `x/net/http2` | retries `ErrCodeRefusedStream` and GOAWAY'd requests on another connection ([golang/go#20985](https://github.com/golang/go/issues/20985)) |
| curl | retries REFUSED_STREAM ([curl#804](https://github.com/curl/curl/issues/804); retry cap added in [curl#5794](https://github.com/curl/curl/issues/5794)) |
| OkHttp | `RetryAndFollowUpInterceptor` marks the connection `noNewStreams` and retries on another one ([square/okhttp#2543](https://github.com/square/okhttp/issues/2543)) |
| .NET `SocketsHttpHandler` | retries REFUSED_STREAM and requests pending on GOAWAY'd connections ([dotnet/runtime#44669](https://github.com/dotnet/runtime/issues/44669)) |
| Rust `reqwest` | built-in retry treats a remote reset with `Reason::REFUSED_STREAM` as retryable ([`src/retry.rs`](https://docs.rs/reqwest/latest/src/reqwest/retry.rs.html)) |
| Chromium / Firefox | retry refused/unprocessed streams transparently |

Practical impact: behind an enterprise front that intermittently refuses streams, the same workload self-heals over HTTP/1.1 (`EOFError`/`SystemError` are in the retryable set) but fails on first occurrence over h2 — h2 is currently strictly less reliable than h1 in HTTP.jl. (Hit in production through a corporate proxy; the front advertises h2 via ALPN but resets streams under some conditions.)

Suggestion: read `rst.error_code` and treat `REFUSED_STREAM` — plus streams rejected by GOAWAY — as retryable per §8.7 (ideally regardless of idempotency, like Go/.NET/reqwest), and include the code in the `ProtocolError` message for the remaining cases so callers can tell `CANCEL` from `INTERNAL_ERROR` or `ENHANCE_YOUR_CALM`. HTTP.jl 1.x already applied the same guaranteed-unprocessed principle at the TCP level (#964).

mathieu17g:
Fixes #1294.

- New `H2StreamResetError` carries the RFC 9113 §7 error code (the `RSTStreamFrame` handler previously discarded it); `showerror` names the code, e.g. `HTTP/2 stream reset by peer: REFUSED_STREAM`.
- `_retryable_request_error` now classifies `REFUSED_STREAM` resets and `H2GoAwayError` as retryable, and `_h2_guaranteed_unprocessed` lifts the idempotent-method gate for them per §8.7 (the replayable-body gate still applies).
- Retries land on a fresh connection via the existing `_drop_h2_conn!`-on-error behaviour.
- Tests: a scripted h2 server refuses the first request (`RST_STREAM(REFUSED_STREAM)` / `GOAWAY(last_stream_id=0)`) then answers 200 — red before the fix (single attempt surfaced as an error), green after (two attempts, fresh connection).

Note: stream resets now raise `H2StreamResetError` (an `HTTPError`) instead of a generic `ProtocolError` — same pattern as the existing `H2GoAwayError`.

### DIFF
--- a/test/http2_refused_stream_retry_tests.jl
+++ b/test/http2_refused_stream_retry_tests.jl
@@ -50,7 +50,7 @@ end
 # REFUSED_STREAM, or GOAWAY rejecting the in-flight stream), then answers 200.
 # An accept loop plus a per-connection headers loop keeps it agnostic to
 # whether the client retries on the same connection or on a new one.
-function _rsr_serve!(listener; scenario::Symbol, refuse_first::Int = 1)
+function _rsr_serve!(listener; scenario::Symbol, refuse_first::Int = 1, rst_code::UInt32 = _RSR_REFUSED_STREAM)
     attempts = Threads.Atomic{Int}(0)
     conns = Threads.Atomic{Int}(0)
     accept_task = errormonitor(Threads.@spawn begin
@@ -72,7 +72,7 @@ function _rsr_serve!(listener; scenario::Symbol, refuse_first::Int = 1)
                         attempt = Threads.atomic_add!(attempts, 1) + 1
                         if attempt <= refuse_first
                             if scenario === :rst
-                                _rsr_write_frame!(conn, HT.RSTStreamFrame(hf.stream_id, _RSR_REFUSED_STREAM))
+                                _rsr_write_frame!(conn, HT.RSTStreamFrame(hf.stream_id, rst_code))
                             elseif scenario === :goaway
                                 _rsr_write_frame!(conn, HT.GoAwayFrame(UInt32(0), _RSR_NO_ERROR, UInt8[]))
                                 break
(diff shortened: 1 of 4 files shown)


test/http2_refused_stream_retry_tests.jl
quinnj, line 1: These look like good tests, but they need to be included in the test/runtests.jl list of test files so they actually get run via CI

mathieu17g: Fixed — the file is now in the test/runtests.jl list (1f1016c, in the branch since your rebase), so the tests run in CI

mathieu17g:
The remaining codecov gap was the §8.7 non-idempotent path (_h2_guaranteed_unprocessed), which GET requests short-circuit; 25171e61 covers it with POST scenarios (refused → retried, GOAWAY-rejected → retried, cancelled → surfaced without retry).

mathieu17g:
Should be ready for merge

quinnj (approved):

PR #1299, opened 2026-06-11
Add SOCKS5 proxy support (socks5:// and socks5h://)

quinnj:
## Summary

Adds SOCKS5 proxy transport to the client, configured the same way as HTTP proxies:

- `socks5://` and `socks5h://` URLs work with `ProxyURL`, `ProxyConfig`, the `proxy=` request kwarg, and `ALL_PROXY`-style environment variables. Following Go's `net/http` behavior, both schemes send domain targets to the proxy for remote resolution.
- Username/password credentials come from the proxy URL userinfo (percent-decoded) and are negotiated via RFC 1929 username/password auth when the proxy selects it.
- The SOCKS handshake runs on the pooled TCP connection before anything else, so plain HTTP/1.1, HTTPS (h1 and ALPN h2), and WebSockets all ride the proxied stream; `connect_timeout` bounds the handshake via the existing connect deadline.
- Requests over SOCKS write origin-form targets and never attach `Proxy-Authorization`; SOCKS plans get distinct pool keys and proxied connections are pooled/reused normally.
- SOCKS handshake failures surface as `HTTP.ConnectError` (with the SOCKS error as `cause`), keeping the "HTTP-typed exceptions only" public error contract; handshake timeouts surface as `HTTP.TimeoutError`. Neither is retried, matching CONNECT-tunnel failure behavior.

Hardening that came out of review:

- strict SOCKS proxy URL authority/port parsing (`socks5://::1:1080`, hex/signed ports, etc. are rejected)
- `_new_conn_tls!` now closes the underlying TCP connection when the TLS handshake fails instead of leaking the fd until GC (covers direct, CONNECT-tunnel, and SOCKS paths)

## Dependencies

Requires Reseau >= 1.3.0, which includes [JuliaServices/Reseau.jl#123](https://github.com/JuliaServices/Reseau.jl/pull/123) (`Reseau.SOCKS`) and is available in General.

## Tests

- proxy URL parsing and plan-mode coverage for socks5/socks5h, including no_proxy bypass, case-insensitive schemes, bracketed IPv6 proxy hosts, default ports, and malformed-authority rejection
- end-to-end: plain h1 through a scripted SOCKS proxy (with RFC 1929 auth), HTTPS h1 and ALPN h2 over bridged SOCKS streams, WebSockets over SOCKS, pool reuse across sequential requests through one handshake
- failure modes: command rejection, auth-method mismatch, empty-username URLs (rejected before any proxy I/O), and a stalled proxy hitting `connect_timeout`
- full local suite green on macOS (aarch64) against registered Reseau v1.3.0; `test/http_client_proxy_tests.jl` standalone: 162 passing

Co-authored by Codex


.github/workflows/ci.yml
```diff
@@ -58,6 +58,9 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v6
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1'
```
quinnj: we shouldn't need this/these change(s)?

src/http_proxy.jl
```diff
@@ -262,19 +270,125 @@ function _matches_no_proxy(matcher::NoProxy, host::AbstractString, port::Integer
...
+    text = String(value)
+    scheme_idx = findfirst("://", text)
+    if scheme_idx === nothing
+        return "http", "http://" * text
+    end
+    scheme = lowercase(String(SubString(text, firstindex(text), prevind(text, first(scheme_idx)))))
+    return scheme, text
+end
+
+function _parse_socks_proxy_target(value::AbstractString, scheme::String)::_ProxyTarget
```
quinnj: it feels like we have a _lot_ of url-parsing-like logic here; more than I would have expected. I guess i would have expected that we would already have the parsed url here to work on? or can use url-parsing helper utilities? can we or is this simple enough that we shouldn't worry about it?

PR #1307, opened 2026-06-15
Bind outbound connections to a local address/interface via local_addr (#834)

### LINKED ISSUE #834: Send http request through specific network interface
I want to send request through specific network interface using HTTP.jl like it works in [here](https://stackoverflow.com/questions/48996494/send-http-request-through-specific-network-interface) . In lib python requests it works correct. Or if to use curl, request should look like this:
`curl --interface <ip/name_of_interface> <host>`

But in HTTP.jl I didnt find this feature. if I send request, my server is receiving request from tun0 interface.
Does anyone know how this can be realize using the HTTP package. How to select the desired interface from the HTTP package

quinnj:
Closes #834.

Adds a `local_addr` keyword to `Transport` and `Client` that binds outbound connections to a specific source IP — and therefore the outgoing interface — modeled on Go's [`net.Dialer.LocalAddr`](https://pkg.go.dev/net#Dialer.LocalAddr) (and `curl --interface`). Useful on multi-homed hosts and for separating traffic by interface.

```julia
# every request from this client leaves via 192.0.2.10
client = HTTP.Client(local_addr = "192.0.2.10")
HTTP.get("http://example.com"; client = client)

# canonical home — local_addr is a connection-pool property:
client = HTTP.Client(transport = HTTP.Transport(local_addr = "192.0.2.10"))
```

## Design

Reseau's `HostResolver`/`TCP.connect` already accept a local bind endpoint (`Reseau = "1.3"` has it); this just threads it through HTTP's transport. Two deliberate choices, both following Go:

- **Per-Transport/Client, not per-request.** `local_addr` is a *connection-identity* property: a connection bound to source A cannot serve a request that wants source B. Go puts it on the `Dialer`, not the request, precisely so it folds into the pool. Each bound `Client` keeps its own pool, so binding needs **no pool-key changes** and can't cause cross-binding connection reuse.
- **IP, not interface name.** Accepts an IP-literal `String` (kernel-chosen ephemeral source port) or a `Reseau.TCP.SocketAddrV4`/`SocketAddrV6` for a fixed source port. Interface names (`curl --interface eth0`) are intentionally not accepted — like Go, the address is an IP, keeping name resolution out of the transport.

Binding to an address not assigned to a local interface fails fast at `bind()`. Passing `local_addr` alongside an explicit `transport` is rejected as ambiguous.

## Tests

`http_client_transport_tests.jl`: the normalizer (string→endpoint, IPv6, passthrough, invalid→`ArgumentError`), client- and transport-level binding round-trips, the ambiguity guard, and that an **unassigned** source address fails — the deterministic proof the bind is actually applied rather than ignored. Documents the knob in the client guide.

### DIFF
--- a/src/http_transport.jl
+++ b/src/http_transport.jl
@@ -230,6 +230,27 @@ mutable struct H1Body <: AbstractBody
     cancel_callback::Union{Nothing,Function}
 end
 
+# Normalize a user-supplied local bind address into a Reseau `SocketEndpoint`.
+# Modeled on Go's `net.Dialer.LocalAddr`: an IP-literal string binds that source
+# IP with an OS-chosen ephemeral port (the common case — pick the interface, let
+# the kernel pick the port), while a ready-made `TCP.SocketAddrV4`/`SocketAddrV6`
+# gives full control including a fixed source port. Interface *names* (curl's
+# `--interface eth0`) are intentionally not accepted — like Go, the address is
+# an IP, which keeps resolution out of the transport.
+_normalize_local_addr(::Nothing)::Union{Nothing,TCP.SocketEndpoint} = nothing
+_normalize_local_addr(addr::TCP.SocketAddrV4)::TCP.SocketEndpoint = addr
+_normalize_local_addr(addr::TCP.SocketAddrV6)::TCP.SocketEndpoint = addr
+function _normalize_local_addr(addr::AbstractString)::TCP.SocketEndpoint
+    s = String(addr)
+    isempty(s) && throw(ArgumentError("local_addr string must be a non-empty IP literal"))
+    v4 = HostResolvers._parse_ipv4_literal(s)
+    v4 === nothing || return TCP.SocketAddrV4(v4, 0)
+    v6 = HostResolvers._parse_ipv6_literal(s)
+    v6 === nothing || return TCP.SocketAddrV6(v6, 0)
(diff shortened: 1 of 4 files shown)


src/http_transport.jl
```diff
@@ -230,6 +230,27 @@ mutable struct H1Body <: AbstractBody
...
+function _normalize_local_addr(addr::AbstractString)::TCP.SocketEndpoint
+    s = String(addr)
+    isempty(s) && throw(ArgumentError("local_addr string must be a non-empty IP literal"))
+    v4 = HostResolvers._parse_ipv4_literal(s)
+    v4 === nothing || return TCP.SocketAddrV4(v4, 0)
+    v6 = HostResolvers._parse_ipv6_literal(s)
+    v6 === nothing || return TCP.SocketAddrV6(v6, 0)
+    throw(ArgumentError("local_addr '$s' is not a valid IPv4 or IPv6 address literal"))
+end
+_normalize_local_addr(addr) = throw(ArgumentError("local_addr must be an IP-literal String or a Reseau TCP.SocketAddrV4/SocketAddrV6, got $(typeof(addr))"))
```
quinnj: we should collapse all these into a single method. then we start with:
```julia
addr === nothing && return
addr isa Union{TCP.SWocketAddrV4,TCPSocketAddrV6} && return addr
...
throw(...)
```
that feels cleaner to just keep it all in one method and not make the compiler/inference pick between methods

PR #1315, opened 2026-06-17
Restore `HTTP.Exceptions` as a deprecating compat shim + migration note

### LINKED ISSUE #1314: `HTTP.Exceptions` submodule removed in 2.x — restore as a (deprecating) compat shim, or document it?
In 1.x, `HTTP.Exceptions` was an internal organizational submodule (`include("Exceptions.jl"); using .Exceptions`, like `Messages`/`Connections`), and the `using` re-exported the types to the top level. The documented names were the top-level ones — `HTTP.StatusError`, `HTTP.ConnectError`, `HTTP.TimeoutError` — and **those still resolve in 2.x** (v2.3.0). What's gone is the `HTTP.Exceptions` *submodule itself* (`isdefined(HTTP, :Exceptions) == false`), so code reaching in via `HTTP.Exceptions.StatusError` breaks.

It's easy to miss because the reference sits in a `catch` block:

```julia
try
    HTTP.get(url)
catch e
    if e isa HTTP.Exceptions.StatusError && e.status == 429
        ...
    end
end
```

That compiles, precompiles, and passes its happy-path tests, then throws `UndefVarError: Exceptions not defined in HTTP` the first time a real error reaches the `catch` — silent until production, unlike the `readtimeout` deprecation which warns. The [migration guide](https://github.com/JuliaWeb/HTTP.jl/blob/master/docs/src/guides/migration-1x.md) covers `readtimeout` → `read_idle_timeout` but not this — and it's not rare: a GitHub code search returns 40 repos using `HTTP.Exceptions.StatusError` and 13 using `HTTP.Exceptions.RequestError`, including AWS.jl, PromptingTools.jl, and WebSockets.jl (this org).

The 7 names 1.x's `Exceptions` exported split cleanly:

| name(s) | in 2.x | suggested |
|---|---|---|
| `HTTPError` · `StatusError` · `ConnectError` · `TimeoutError` | top-level in `HTTP` | deprecating re-export (below) |
| `RequestError` | removed, no equivalent ¹ | migration-guide note |
| `@try` · `current_exceptions_to_string` | removed | copy back if wanted (trivial) |

¹ `RequestError` (thrown, fields `.request`/`.error`) has no drop-in: 2.x lets the underlying exception propagate, with `isrecoverable(err)` as the classifier (made public in #1310). `RequestRetryError` is the `retry_if` wrapper (field `.err`), not what's thrown — so restoring it means reintroducing the type *and* its throw sites, a behavior change, not a shim.

For the four that just moved, a small deprecating submodule keeps old code working and nudges migration (slots into `src/HTTP.jl` before `end # module`):

```julia
module Exceptions
    import ..HTTP
    Base.@deprecate_binding StatusError  HTTP.StatusError  false
    Base.@deprecate_binding ConnectError HTTP.ConnectError false
    Base.@deprecate_binding TimeoutError HTTP.TimeoutError false
    Base.@deprecate_binding HTTPError    HTTP.HTTPError    false
end
```

`HTTP.Exceptions.StatusError` resolves again, forwards to `HTTP.StatusError`, and depwarns under `--depwarn=yes` / in tests — silent in production.

This mirrors the manual-`Cookie` gap (#1283), where the docs-only note (#1284) gave way to restoring the behavior (#1287); and #1267, where you were "open to … easy/simple things we can add back to 2.X for compat." So: the deprecating shim for the four moved types plus a `migration-1x.md` note for `RequestError`, or a note only? Happy to PR either.

mathieu17g:
Implements the deprecating `HTTP.Exceptions` shim and migration note discussed in #1314 (context and prevalence there).

`src/HTTP.jl` re-adds `HTTP.Exceptions` as a thin deprecated submodule forwarding the four relocated types to their top-level names:

```julia
module Exceptions
    import ..HTTP
    Base.@deprecate_binding HTTPError    HTTP.HTTPError    false
    Base.@deprecate_binding StatusError  HTTP.StatusError  false
    Base.@deprecate_binding ConnectError HTTP.ConnectError false
    Base.@deprecate_binding TimeoutError HTTP.TimeoutError false
end
```

So `catch … isa HTTP.Exceptions.StatusError` resolves again, forwarding to `HTTP.StatusError` with a `--depwarn=yes` warning — same mechanism as the existing `@deprecate escape escapeuri`, and like it, untested.

`RequestError`, `@try`, and `current_exceptions_to_string` are not reinstated; the migration note covers them (`RequestError` has no drop-in — the underlying exception propagates and `isrecoverable` classifies it, #1310).

`docs/src/guides/migration-1x.md` gains an "Exceptions" section (Before/After for `StatusError` and `RequestError`) and a Final-Checklist line.

Closes #1314.

### DIFF
--- a/docs/src/guides/migration-1x.md
+++ b/docs/src/guides/migration-1x.md
@@ -359,6 +359,55 @@ Use the timeout that matches your intent:
 Timeout failures are reported as `HTTP.HTTPTimeoutError`, an alias for
 `HTTP.TimeoutError`.
 
+## Exceptions
+
+In 1.x the client exception types lived in the `HTTP.Exceptions` submodule and were
+re-exported to the top level. 2.x keeps the types at the top level — `HTTP.StatusError`,
+`HTTP.ConnectError`, `HTTP.TimeoutError`, and their `HTTP.HTTPError` supertype — and
+`HTTP.Exceptions` is now a thin deprecated shim. Prefer the top-level names.
+
+Before:
+
+```julia
+catch e
+    e isa HTTP.Exceptions.StatusError && @info "status" e.status
+end
+```
+
+After:
+
+```julia
+catch e
+    e isa HTTP.StatusError && @info "status" e.status
+end
+```
+
+`HTTP.Exceptions.StatusError`, `ConnectError`, `TimeoutError`, and `HTTPError` still
+resolve — forwarding to the top-level names, with a deprecation warning under
+`--depwarn=yes`. The 1.x `@try` macro and `current_exceptions_to_string` helper were
+internal and are removed with no replacement.
+
+`HTTP.RequestError` is also gone. 1.x wrapped request-path failures in a `RequestError`
+(fields `.request`/`.error`); 2.x lets the underlying transport or protocol exception
+propagate directly, and exposes `HTTP.isrecoverable` to classify whether a failure is a
+transient one that is safe to retry.
+
+Before:
+
+```julia
+catch e
+    e isa HTTP.Exceptions.RequestError && @warn "request failed" e.error
+end
+```
+
+After:
+
+```julia
+catch e
+    HTTP.isrecoverable(e) || rethrow()   # otherwise it is a transient transport failure
+end
+```
+
 ## TLS, Sockets, and Proxies
 
 The old `sslconfig` and `socket_type_tls` extension points are retained for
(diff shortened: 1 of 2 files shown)


mathieu17g:
The `codecov/project` -0.16% is an indirect drop in `http2_server.jl`'s non-deterministic error/RST paths (base 10 coverage sessions vs this PR's 5; `codecov/patch` green; this PR's files untracked) — a re-run should re-aggregate and clear it.

quinnj (approved):
LGTM; thanks!

Issue #254, opened 2018-06-13
Get IP of request


essenciary:
Is it possible to get the IP address of the request, please? 

With `HttpServer` it could be done like this: 
```julia 
http.events["connect"] = (http_client) -> handle_connect(http_client)

function handle_connect(client::HTTP.Client)
  try
    ip, port = getsockname(isa(client.sock, MbedTLS.SSLContext) ? client.sock.bio : client.sock)
  catch
  end
 end
```

essenciary:
@quinnj Thanks - I've been chasing my tail but I can't figure out how to use this with `HTTP.serve`

```julia
HTTP.serve(parse(IPAddr, host), port, verbose = verbose, rate_limit = ratelimit) do req::HTTP.Request
  setup_http_handler(req)
end
```

PR #1317, opened 2026-06-22
Add HTTP.peeraddr(::Stream) for the client socket address

PingoLee:
## Summary

Adds `HTTP.peeraddr(stream)`, a public accessor returning the remote (client) `SocketAddr` of a server stream.

## Problem

HTTP 2.x has no public way to read the client IP from a server `Stream`. In 1.x, `Sockets.getpeername(::HTTP.Stream)` covered this (PR #702), but there is no v2 equivalent — callers must reach through transport internals (`stream.tracked.conn` → `.tcp`/`.tls` → `raddr`), and the naive TCP path throws on TLS connections. The client IP is needed for rate limiting, audit logging, and per-client policy. This is a long-standing ask: #699 and #254.

## Change

`peeraddr(stream)` reads the remote endpoint via the **public** `remote_addr` accessor that both the TCP and TLS transports expose (`TLS.remote_addr` delegates to the underlying TCP conn), so there is **no internal field access** and no TLS footgun. It resolves the connection from `tracked` (the normal path for both HTTP/1 and HTTP/2; an h2 stream is built with `tracked` plus a shared `h2_conn`), so it works across protocols and both plain-TCP and TLS. Returns `nothing` when the peer endpoint is unavailable; throws `ArgumentError` on a client-side stream.

## Naming

Named `peeraddr` rather than restoring `getpeername` because the return type is now a Reseau `SocketAddr` (not a `Sockets` `(ip, port)` tuple); reusing the old name with a new return type would be a silent migration trap. Happy to rename to `remoteaddr` for transport-layer consistency if preferred.

## Compatibility

Purely additive. `peeraddr` is declared `public`, documented in the server API reference, and the 1.x→2.x migration guide maps `getpeername` → `peeraddr`.

## Testing

New testsets cover HTTP/1, HTTP/2, and TLS server streams (client reported as IPv4 loopback with a non-zero source port), plus the client-side `ArgumentError` and no-live-connection (`nothing`) edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added `HTTP2Settings` to configure HTTP/2 receive flow-control windows (per-stream `initial_window_size` and connection-level `connection_window_size`). Pass it via the `http2_settings` keyword on `Client`, `Server`, `listen!`, `serve!`, `serve`, and `connect_h2!`. Defaults preserve the protocol-default 65535-byte windows, and the per-stream receive buffer cap is derived from the window. Raising the windows improves single-stream throughput on links with non-trivial latency.
+- Added `HTTP.peeraddr(::HTTP.Stream)`, returning the remote (client) `SocketAddr` of a server stream for both plain-TCP and TLS connections and both HTTP/1 and HTTP/2. This is the supported way to obtain the client IP (for rate limiting, audit logging, and per-client policy) without reaching into transport internals, and restores the capability `Sockets.getpeername(::HTTP.Stream)` provided in HTTP.jl 1.x.
 
 ### Fixed
 - Percent-decode `userinfo` before building the `Basic` auth header (RFC 3986); fixes wrong credentials for request URLs and proxies containing percent-encoded characters.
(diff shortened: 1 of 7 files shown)


quinnj (approved):
LGTM; thanks!

fredrikekre:
Why not overload `Sockets.getpeername`?

PingoLee:
Hi @fredrikekre 

If I understood correctly, HTTP.jl 1.x Sockets.getpeername(::HTTP.Stream) delegated to Julia’s Sockets.getpeername, returning the standard (ip::Sockets.IPAddr, port::UInt16) tuple. HTTP.jl v2 is built on Reseau, whose native remote endpoint type is Reseau.TCP.SocketAddr (SocketAddrV4/SocketAddrV6), with .ip as an NTuple of octets and .port::UInt16.

So I think that even keeping the name, some refactoring would be necessary.

Therefore, changing the name might draw more attention to the change and facilitate the migration.

But of course, in the age of AI, perhaps this is just nitpicking on my part.

quinnj:
I've tried to keep Reseau/HTTP completely Sockets-free during the migration/2.0 release, mostly as a sanitization thing and because AI models will reach for it all over the place while making fairly mundane changes/features. That said, I agree it's worth considering now that 2.0 has mostly settled; but then again, I think @PingoLee has some good points that, while similar, it's still different machinery under the hood.

PR #1318, opened 2026-06-23
client: don't synthesize the default port into the Host header

krynju:
Found this issue when porting to HTTP.jl 2 and AWS S3 presigned links were not working as is.

## The problem

For a URL **without an explicit port**, the client puts the scheme's default
port into the `Host` header:

```julia
HTTP.get("https://example.com/")
# sends:  Host: example.com:443
```

The `Host` header should mirror the URL's authority **as written**. A bare-host
URL should send a bare `Host: example.com`. HTTP.jl 1.x did this; 2.0 regressed
because `request.host` is now built from the *connection address*
(`example.com:443`), which always carries a port for dialing.

## Why it matters

`Host: example.com:443` is technically legal (RFC 9110 §7.2 allows a port), but
many servers compare the `Host` value verbatim and a synthesized default port
breaks them. The case that bit us is **AWS S3 presigned URLs (SigV4)**:

- The signature is computed over the canonical host `s3.amazonaws.com`.
- HTTP.jl 2.0 sends `Host: s3.amazonaws.com:443`.
- S3 recomputes the signature from the received `Host`, gets a different value,
  and rejects the request: `403 SignatureDoesNotMatch`.

The same request succeeds with `curl` and with HTTP.jl 1.x, which send a bare
`Host`.

## What other clients do

Go's `net/http` keeps the `Host` header equal to the URL authority and adds the
default port **only** to the dial address, never to `Host`:

```go
req, _ := http.NewRequest("PUT", "https://s3.amazonaws.com/b/k", body)
httputil.DumpRequestOut(req, false)
//   PUT /b/k HTTP/1.1
//   Host: s3.amazonaws.com            <- bare; :443 not added

req2, _ := http.NewRequest("PUT", "https://s3.amazonaws.com:443/b/k", body)
//   Host: s3.amazonaws.com:443        <- explicit port preserved
```

`curl` and HTTP.jl 1.x behave the same way. HTTP.jl 2.0 is the outlier.

## The fix

Add a `host_header` view on the parsed URL that mirrors the authority as
written — an explicit port is preserved, the default port is never synthesized
(and IPv6 brackets are kept) — and build `request.host` from it. The connection
address is unchanged, so dialing still targets the correct port.

| URL | dial address (unchanged) | `Host` header (before → after) |
|-----|--------------------------|--------------------------------|
| `https://example.com/`        | `example.com:443` | `example.com:443` → **`example.com`** |
| `https://example.com:443/`    | `example.com:443` | `example.com:443` → `example.com:443` |
| `http://minio:9000/`          | `minio:9000`      | `minio:9000` → `minio:9000` |
| `https://[2001:db8::1]/`      | `[2001:db8::1]:443` | `[2001:db8::1]:443` → **`[2001:db8::1]`** |

Unit tests added in `test/http_client_tests.jl` covering bare host, explicit
default port, http default port, a custom port, and IPv6 (bracketed) forms.

Verified end-to-end: a real S3 presigned PUT that returns `403` on current
`master` returns `200` with this change.

### DIFF
--- a/src/http_client.jl
+++ b/src/http_client.jl
@@ -1944,7 +1944,7 @@ function request(
             parsed.target;
             headers=req_headers,
             body=normalized_body.body,
-            host=parsed.address,
+            host=parsed.host_header,
             content_length=normalized_body.content_length,
             context=req_context,
         )
--- a/src/http_client.jl
+++ b/src/http_client.jl
@@ -895,7 +895,7 @@ function _do_incoming!(
             previous_secure = current_secure
             previous_address = current_address
             previous_target = current_request.target
-            next_address, next_secure, next_target = _resolve_redirect_target(current_address, current_secure, location, current_request.target)
+            next_address, next_secure, next_target, next_host_header = _resolve_redirect_target(current_address, current_secure, location, current_request.target, current_request.host)
             _emit_trace(
                 trace,
                 RedirectEvent(
@@ -940,7 +940,7 @@ function _do_incoming!(
                     cookies = Cookie[]
                 end
             end
-            current_request.host = current_address
+            current_request.host = next_host_header
             break
         end
     end
--- a/src/http_client_redirect.jl
+++ b/src/http_client_redirect.jl
@@ -205,7 +205,15 @@ end
 # `_urlparts_host_header!`. For a host-changing redirect both come from the
 # parsed Location; for a same-authority relative redirect the host is unchanged,
 # so `current_host_header` is carried through verbatim.
-function _resolve_redirect_target(current_address::String, current_secure::Bool, location::String, current_target::String, current_host_header::String)
+#
+# `current_host_header` is the current hop's `request.host`, which is `nothing`
+# for low-level `do!` callers that pin `Host` only in the request headers. The
+# header is stripped on each hop by `_prepare_request_for_redirect`, so the next
+# `Host` is taken from this returned value; on a relative redirect with no parsed
+# host we fall back to the dial `address` (carrying the port, as before this
+# change) rather than returning `nothing`, which would drop the `Host` header
+# entirely. An absolute redirect always yields a proper parsed `host_header`.
+function _resolve_redirect_target(current_address::String, current_secure::Bool, location::String, current_target::String, current_host_header::Union{Nothing,String})
     scheme_match = match(r"^([A-Za-z][A-Za-z0-9+\\.-]*):", location)
     if scheme_match !== nothing
         scheme = lowercase(String(scheme_match.captures[1]))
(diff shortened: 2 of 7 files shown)


quinnj:
Thanks for the focused fix. I think this needs one more pass before merge: the new `host_header` value is only wired through the high-level request path, so a few sibling client paths still synthesize default ports into `Host`.

- `HTTP.open` / streaming requests still construct the request with `parsed.address`, so `HTTP.open("PUT", "https://example.com/upload")` will still send `Host: example.com:443`.
- Redirects reset `current_request.host = current_address`, and WebSocket handshakes use `host=parsed.address` initially and after redirects, so default-port authorities can still leak there too.

Could you route those paths through the same authority-as-written value and add focused regressions for streaming and redirect/WebSocket coverage?

[findings by codex; reviewed by quinnj]

krynju:
> Thanks for the focused fix. I think this needs one more pass before merge: the new `host_header` value is only wired through the high-level request path, so a few sibling client paths still synthesize default ports into `Host`.
> 
> * `HTTP.open` / streaming requests still construct the request with `parsed.address`, so `HTTP.open("PUT", "https://example.com/upload")` will still send `Host: example.com:443`.
> * Redirects reset `current_request.host = current_address`, and WebSocket handshakes use `host=parsed.address` initially and after redirects, so default-port authorities can still leak there too.
> 
> Could you route those paths through the same authority-as-written value and add focused regressions for streaming and redirect/WebSocket coverage?
> 
> [findings by codex; reviewed by quinnj]



I think this is fully covered now in https://github.com/JuliaWeb/HTTP.jl/pull/1318/commits/5729fdc4957b2ba3bed1fa73c86763571580280e

quinnj:
Thanks for the quick follow-up. The original high-level, streaming, WebSocket, and redirect Host handling gaps look fixed now. I found one remaining low-level redirect regression: `_resolve_redirect_target` now requires `current_host_header::String`, but `_do_incoming!` passes `current_request.host` directly.

For `HTTP.do!` callers that provide `Host` only in headers (`request.host === nothing`), a redirect now throws `MethodError` instead of following the redirect. I reproduced this with `HTTP.do!(client, addr, HTTP.Request("GET", "/start"; headers=["Host" => addr], ...); redirect_limit=1, protocol=:h1)`.

[findings by codex; reviewed by quinnj]

krynju:
> Thanks for the quick follow-up. The original high-level, streaming, WebSocket, and redirect Host handling gaps look fixed now. I found one remaining low-level redirect regression: `_resolve_redirect_target` now requires `current_host_header::String`, but `_do_incoming!` passes `current_request.host` directly.
> 
> For `HTTP.do!` callers that provide `Host` only in headers (`request.host === nothing`), a redirect now throws `MethodError` instead of following the redirect. I reproduced this with `HTTP.do!(client, addr, HTTP.Request("GET", "/start"; headers=["Host" => addr], ...); redirect_limit=1, protocol=:h1)`.
> 
> [findings by codex; reviewed by quinnj]

Nice catch! Fixed in https://github.com/JuliaWeb/HTTP.jl/pull/1318/commits/73936f2554efea942c424d34fd69a8ca796dfecf

quinnj (approved):
LGTM; thanks!

PR #1319, opened 2026-06-24
Add HTTP QUERY method support

quinnj:
## Summary

Adds first-class support for the RFC 10008 HTTP `QUERY` method across HTTP.jl's client and server-facing APIs.

- Adds `HTTP.query(...)`, `HTTP.query(client, ...)`, and `HTTP.@client` helper support without adding a new bare export.
- Treats `QUERY` as safe/idempotent for high-level retries and stale reused-connection transport retries.
- Preserves `QUERY` method and replayable body across `301`, `302`, `307`, and `308` redirects while keeping `303` rewrite behavior to `GET`.
- Canonicalizes the `Accept-Query` header and documents `QUERY` request usage.
- Adds live client/router tests plus redirect, retry, transport, and header coverage.

## Validation

- `julia --project=. test/runtests.jl`
- `julia --project=docs docs/make.jl`
- Fresh temp-env focused run with clean `Reseau` 1.3.1 checkout: `http_core_tests.jl`, `http_handlers_tests.jl`, `http_client_tests.jl`, `http_client_transport_tests.jl`

Note: the local checkout has an untracked manifest pointing at an older dirty `~/.julia/dev/Reseau`, so the final focused verification used a temporary clean Reseau 1.3.1 checkout to match the current `master` compat.

Co-authored by Codex

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,17 @@ r = HTTP.post(
 returned = JSON.parse(String(r.body))
 ```
 
+For safe, idempotent requests with content, HTTP.jl supports the RFC 10008
+`QUERY` method through `HTTP.query`:
+
+```julia
+r = HTTP.query(
+    "https://api.example.com/search";
+    headers = ["Content-Type" => "application/json"],
+    body = JSON.json(Dict("select" => ["name", "email"])),
+)
+```
+
 Stream directly into an `IO` sink with `response_stream`, or use `HTTP.open` when
 you want pull-based control over the response stream. The `do`-block form of
 `HTTP.open` returns the final `HTTP.Response`, not the value returned by the
--- a/docs/src/api/client.md
+++ b/docs/src/api/client.md
@@ -22,6 +22,7 @@ HTTP.roundtrip!
 HTTP.request
 HTTP.get
 HTTP.head
+HTTP.query
 HTTP.post
 HTTP.put
 HTTP.patch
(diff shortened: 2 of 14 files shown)


quinnj:
Reviewed against RFC 10008 — the implementation looks correct and complete.

The redirect matrix is the subtle part and it's exactly right: QUERY is preserved on 301/302/307/308 and only rewritten to GET on 303, matching §2.5 — including the easy-to-miss note that "the exceptions for redirecting a POST as a GET after a 301 or 302 do not apply to QUERY." (Verified the preserved-method path carries the body via `_copy_request` and is still gated by `_redirect_body_replayable`, same as 307/308 POST.) Safe + idempotent retry treatment at both the high-level (`_retryable_request_method`) and stale-connection (`_retryable_method`) layers, `Accept-Query` canonicalization, the `query`/`query(client,…)`/`@client` surface with no new bare export, and the e2e router test all check out.

One QUERY-specific nuance worth considering (non-blocking): §2/§2.1 makes `Content-Type` a hard requirement — a server **MUST** fail a QUERY whose `Content-Type` is missing (400), unlike POST where a missing type is usually tolerated. `Dict`/`NamedTuple` bodies auto-set `application/x-www-form-urlencoded`, so those are covered, but a raw string/bytes body relies on the caller to set it (the JSON doc example does). Since a missing type is a guaranteed round-trip-to-400 for QUERY specifically, it might be worth failing fast client-side (a warning, or surfacing it) when a QUERY carries a non-empty body without `Content-Type` — purely optional.

PR #1330, opened 2026-07-11
fix: trim-compatibility for Router registration, the precompile workload, and one error path

quinnj:
## Summary

- specialize router registration on concrete handler and middleware types
- preserve concrete route segment storage
- allow static builds to disable HTTP's runtime precompile workload
- keep HTTP/2 proxy errors trim-resolvable
- narrow client request copies and server response bodies at trim-sensitive boundaries
- add focused client and server trim-compilation coverage
- flush live SSE events before producer close

## Why

Julia does not specialize function-valued arguments that are only passed through. Router registration therefore converged on widened handler types, leaving middleware wrapping and parametric route construction as runtime operations that `juliac --trim=safe` could not resolve.

The same real application build exposed a few later widening boundaries in request copying and response serialization. This PR uses concrete type parameters at registration and explicit narrowing at those boundaries. It does not change HTTP dispatch behavior or the public request and response model.

The package precompile workload starts real servers and requests. Static compilation must be able to disable that workload, so `HTTP_PRECOMPILE_WORKLOAD=0` provides an explicit build-time opt-out.

Julia 1.13 RC1 does not yet preserve `jl_new_task` starter reachability during safe trimming. Task-backed fixtures therefore remain strict on Julia 1.12 and on patched Julia 1.13+ runtimes that set `HTTP_TRIM_TASK_PATCHED=1`; vanilla 1.13+ still runs every non-task trim fixture. This keeps the package matrix honest without treating a compiler limitation as an HTTP verifier regression.

SSE response reads previously asked the producer-backed buffer to fill the entire destination, so a live event could remain buffered until the producer closed. The server path now forwards currently available bytes immediately while preserving any remainder for the next read.

## Validation

- the complete local HTTP.jl test suite passes, including all 78 trim tests
- focused HTTP/1 and HTTP/2 SSE tests prove the first event arrives before the producer closes
- HTTP client, handler, public API, HTTP/2, precompile-workload, and documentation tests pass locally
- task-backed and strict-body trim fixtures pass locally on Julia 1.12.6
- the compiler-limit gate behaves as expected on Julia 1.13 RC1
- the exact PR head passes the full upstream Linux, macOS, Windows, Julia stable, and Julia prerelease matrix
- Codecov project and patch checks pass
- LeagueEasy's 30-plus-package server graph compiles with a patched Julia nightly and `juliac --trim=safe` with zero verifier errors attributable to HTTP
- the resulting trimmed LeagueEasy binary is deployed and serving production traffic

Co-authored by Codex


quinnj:
Both review points addressed in 26c9b731:

**Consolidated narrowing helpers** — the per-site isa chains are gone. Two `@inline` helpers in http_core are now the single source of truth for the covered type set:
- `_with_body_narrowed(f, body)` — the body shapes (`nothing`/`String`/`SubString{String}`/`Vector{UInt8}`/`EmptyBody`/`BytesBody{Vector{UInt8}}`/`AbstractBody` residual/fallback). Each branch re-narrows so `f`'s call is statically dispatched; sites pass a closure (or a multi-method function like the new `_write_h2_body_shape!` shapes) with their site-specific logic. Used by `_response_has_body`, the h1 chunked/exact writes, h2 emptiness + byte normalization + streaming writer, and `_body_close_any!`.
- `_with_response_narrowed(f, response)` — the `Response{B}` equivalent, shared by both boundary dispatch shims.

Adding a supported shape is now one branch in one place.

**Trim-compile tests** — turns out master already had the trim harness + 13 client workloads, so I added two server-write workloads to it (both zero-tolerance: compile with 0 verifier errors AND run the binary with wire-level assertions, under stock juliac):
- `http_trim_server_response_shapes.jl`: serves every supported body shape and verifies each over a raw exchange — this exercises the widened-response shims, both narrowing helpers, and the h1 write chains end-to-end.
- `http_trim_server_router_registration.jl`: Router construction + every `register!` form (middleware wrap, wildcard/param paths, do-block form). Serving *through* the router is deliberately not covered yet — that's the handler-table dispatch frontier your router-dispatch design will address; the workload comment marks where the serving case should go once it lands.

Validation: both new workloads pass locally (`HTTP_TRIM_ONLY=… ` runs them in ~10s each thanks to juliac caching), h1+h2 server testsets green, and the downstream full-server graph's verifier count is unchanged by the refactor.

quinnj:
Closing this because the current application no longer needs any part of this patch.

Soleil now routes endpoints through Servo.Router. The HTTP.Router registration and HandlerFn changes are therefore outside the application boundary.

I retested every non-router trim workload from this PR against released HTTP v2.6.1. All 13 workloads pass with the verifier kept strict. This includes HTTP/1, HTTP/2, TLS, WebSocket, file-server, and mixed response-body server cases.

I also compiled the full current Soleil graph with HTTP v2.6.1 and with the old pinned HTTP commit. Both builds report the same 41 verifier errors and 0 warnings. Those errors are in Servo, Soleil, URIs, and Base. None is in HTTP. The old HTTP pin provides no improvement.

The SSE correction is already on master through #1339. No supported HTTP diff remains to merge.

Issue #1333, opened 2026-07-14
Documentation: Baked responses via streamhandler must have Vector{UInt8} bodies


frankier:
This example shows some potential confusion when attempting to use baked responses (or otherwise reuse whole response objects).

The pattern works in all cases except for with a string body and a stream response. In this case it will work once, but then the body object, reading a stream of codepoints, will be closed. This is a bit surprising and perhaps should be documented. Additionally, perhaps an exception should be thrown when attempting to send an already closed body. I will prepare a PR for the latter.

```
#=
Run from the HTTP.jl working copy:

    julia --project=. baked_server.jl

Then, for example:

    curl -i http://127.0.0.1:8081/string
    curl -i http://127.0.0.1:8081/bytes
    curl -i http://127.0.0.1:8082/string
    curl -i http://127.0.0.1:8082/bytes
=#
module BakedServer

using HTTP

const HANDLER_PORT = 8081
const STREAM_PORT = 8082

const STRING_BODY = "Hello from a baked String body!\n"^32
const BYTES_BODY = Vector{UInt8}(codeunits("Hello from a baked Vector{UInt8} body!\n"^32))

const RESP_STRING = HTTP.Response(200, ["Content-Type" => "text/plain"]; body=STRING_BODY)
const RESP_BYTES = HTTP.Response(200, ["Content-Type" => "text/plain"]; body=BYTES_BODY)

const HANDLER_ROUTER = HTTP.Router()
HTTP.register!(HANDLER_ROUTER, "GET", "/string", req -> RESP_STRING)
HTTP.register!(HANDLER_ROUTER, "GET", "/bytes", req -> RESP_BYTES)

const STREAM_ROUTER = HTTP.Router()
HTTP.register!(STREAM_ROUTER, "GET", "/string", req::HTTP.Request -> RESP_STRING)
HTTP.register!(STREAM_ROUTER, "GET", "/bytes", req::HTTP.Request -> RESP_BYTES)

function start()
    handler_server = HTTP.serve!(HANDLER_ROUTER, "127.0.0.1", HANDLER_PORT)
    stream_server = HTTP.listen!(HTTP.streamhandler(STREAM_ROUTER), "127.0.0.1", STREAM_PORT)
    @info "request-handler server listening on http://127.0.0.1:$HANDLER_PORT (/string, /bytes)"
    @info "stream-handler server listening on http://127.0.0.1:$STREAM_PORT (/string, /bytes)"
[...]

PR #1336, opened 2026-07-16
Return 500 on invalid reuse of response body

frankier:
In case a response with a consumable body is consumed, currently a 400 error will be returned. However, the error is actually from the server side, so a 500 should be returned.  This is done by throwing an error when _write_response_body_to_stream is called on a closed body. Additionally, the error path is modified so that the framing headers on the response are reset.

See https://github.com/JuliaWeb/HTTP.jl/issues/1333

Co-authored by Codex

### DIFF
--- a/src/http_server_streams.jl
+++ b/src/http_server_streams.jl
@@ -452,7 +452,9 @@ function _write_response_body_to_stream!(stream::Stream, body)::Nothing
         return nothing
     end
     if body isa AbstractBody
-        if body_closed(body)
+        if body_closed(body) &&
+           _server_stream_allows_body(stream) &&
+           stream.response.content_length != 0
             throw(ArgumentError("body is closed"))
         end
         buf = Vector{UInt8}(undef, 16 * 1024)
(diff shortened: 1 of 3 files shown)


src/http_server_streams.jl
```diff
@@ -452,6 +452,11 @@ function _write_response_body_to_stream!(stream::Stream, body)::Nothing
         return nothing
     end
     if body isa AbstractBody
+        if body_closed(body) &&
+           _server_stream_allows_body(stream) &&
+           stream.response.content_length != 0
```
quinnj: For a closed `CallbackBody` with `Content-Length: 0`, `response.content_length` remains `-1` until `startwrite` parses the header. This guard runs first and changes `master`’s `200` to `500`. Please resolve the header length here and test this case.

[reviewed by codex & quinnj]

frankier: This review seems to concern the change you made. Did you want me revise it further?

quinnj: Ah shoot. sorry about that. took care of it.

PR #1340, opened 2026-07-29
fix: join duplicate header values with "," rather than ", "

quinnj:
`appendheader` merges an adjacent duplicate header entry into the one before it, joining the values with `", "`. It should join with `","`.

```julia
HTTP.Request("GET", "/", ["My-Header1" => "value2", "My-Header1" => "value1"])
# before: "My-Header1" => "value2, value1"
# after:  "My-Header1" => "value2,value1"
```

## Why

RFC 9110 §5.3 permits either form when combining repeated field lines — "separated by a comma (`,`) and optional whitespace (OWS)" — so both are legal on the wire. But **AWS SigV4 and Azure SharedKey request canonicalization both require exactly `","` with no whitespace.**

The important part is *where* this runs: `mkheaders` calls `appendheader` on the **request construction** path, not just when parsing responses. So this is not a cosmetic difference in received headers — any library that signs cloud requests using HTTP.jl computes an incorrect `Authorization` signature whenever the outgoing request carries duplicate headers. The signer cannot correct for it after the fact, because by the time it sees the request the values are already merged.

## Evidence

Checked against the official AWS SigV4 conformance suite. The two duplicate-header cases produce the published expected signatures with `","` and do not with `", "`:

| case | `", "` | `","` |
|---|---|---|
| `get-header-key-duplicate` | signature mismatch | matches published vector |
| `get-header-value-order` | signature mismatch | matches published vector |

## Scope

Deliberately unchanged:
- **`Set-Cookie` remains exempt** from merging.
- Merging is still only performed for an **immediately adjacent** duplicate — non-adjacent entries are left as separate entries, as before.

## Tests

Existing assertions that encoded the `", "` form are updated — all of them are direct expectations of merged header values, listed here so they can be checked rather than taken on trust:

- `test/http_core_tests.jl` — `X-Test` merge, two `x-forwarded-for` assertions, and the `Accept` comma-merge case
- `test/http1_wire_tests.jl` — the `X-Forwarded-For` wire assertions

Added a regression testset covering two adjacent duplicates, three adjacent duplicates, the `mkheaders` path (both the vector-of-pairs and vector-valued-dict forms), non-adjacent duplicates staying unmerged, and `Set-Cookie` never merging.

## CI hardening

The CI failures were unrelated to the header separator. This PR now hardens three independent timing and error-shape assumptions.

### Reseau timeout wrapper

CI now resolves Reseau 1.3.2 or later. These releases preserve a caller-owned TLS deadline as a nested `TLSError` with a `DeadlineExceededError` cause. HTTP then wraps that error in `HTTP.TLSHandshakeError`.

The timeout test helper already followed the Reseau wrapper, but it stopped at the HTTP wrapper. It now follows both layers. This keeps the test compatible with the direct `HTTP.TimeoutError` from Reseau 1.3.1 and the nested deadline error from Reseau 1.3.2 or later.

### Cold precompile workload

The local precompile workload used 1-second and 5-second request deadlines. Those deadlines include cold native compilation. A fresh macOS ARM runner needed more than 5 seconds to compile and run the first local TLS path.

The workload now uses finite 60-second deadlines. This keeps the anti-hang protection while allowing slow cold compilation.

### Retry refund test

The retry refund test used a 1-second request deadline. Cold Windows compilation could consume that deadline before the first local response. Its retry delay was also random from 0 to 60 seconds, so the test did not always force deadline preemption.

The test now returns `Retry-After: 60` and uses a 10-second request deadline. The retry layer rejects the fixed delay immediately because it exceeds the remaining deadline. The test does not wait for 60 seconds.

Local validation:

- `test/http_core_tests.jl`
- `test/http1_wire_tests.jl`
- `test/http_client_tests.jl`
- `test/precompile_tests.jl`, three consecutive runs
- `test/http_retry_tests.jl`, five consecutive runs
- Full `Pkg.test()`, including all 63 trim-compilation checks

Version bumped to 2.6.0 — behaviour-visible, so a minor rather than a patch.

Co-authored by Codex

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "HTTP"
 uuid = "cd3eb016-35fb-5094-929b-558a96fad6f3"
-version = "2.5.5"
+version = "2.6.0"
 authors = ["Jacob Quinn", "contributors: https://github.com/JuliaWeb/HTTP.jl/graphs/contributors"]
 
 [deps]
--- a/src/http_core.jl
+++ b/src/http_core.jl
@@ -897,7 +897,8 @@ end
 Append a header value to `headers`.
 
 If the previous stored header has the same name and the key is not
-`Set-Cookie`, the value is merged into the previous entry with a comma.
+`Set-Cookie`, the value is merged into the previous entry with a comma
+(no whitespace), the canonical combined-field-value form of RFC 9110 §5.3.
 Otherwise a new pair is appended.
 """
 function appendheader(headers::Headers, header::Pair)
@@ -905,7 +906,7 @@ function appendheader(headers::Headers, header::Pair)
     if !isempty(headers.entries)
         last_header = headers.entries[end]
         if first(item) != "Set-Cookie" && first(last_header) == first(item)
-            headers.entries[end] = first(last_header) => string(last(last_header), ", ", last(item))
+            headers.entries[end] = first(last_header) => string(last(last_header), ",", last(item))
             return headers
         end
     end
--- a/src/http_core.jl
+++ b/src/http_core.jl
@@ -898,7 +898,8 @@ Append a header value to `headers`.
 
 If the previous stored header has the same name and the key is not
 `Set-Cookie`, the value is merged into the previous entry with a comma
-(no whitespace), the canonical combined-field-value form of RFC 9110 §5.3.
+(no whitespace), as permitted by RFC 9110 §5.3 and required by common
+request-signing canonicalizations.
 Otherwise a new pair is appended.
 """
 function appendheader(headers::Headers, header::Pair)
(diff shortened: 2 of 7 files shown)


quinnj:
### CI update

The original header change was not the cause of the failures. I traced and hardened three independent CI assumptions on this branch:

- Reseau 1.3.2 or later nests a caller-owned TLS deadline below `HTTP.TLSHandshakeError`. The timeout helper now follows that wrapper.
- Cold local TLS compilation on macOS ARM exceeded the precompile workload's 5-second budget. The workload now keeps finite 60-second deadlines.
- The retry refund test used a 1-second first-response budget and a random 0-to-60-second retry delay. It now uses `Retry-After: 60` with a 10-second request budget.

Exact head `cfd8ba3e385988f4a731e3ca212048259a245ab5` is green:

- Documentation
- Julia 1 on Ubuntu, Windows, and macOS ARM
- Julia pre on Ubuntu and Windows
- Codecov patch and project

Local validation also passed:

- Full `Pkg.test()`, including all 63 trim-compilation checks
- Core, wire, and client test files
- Precompile test file, three consecutive runs
- Retry test file, five consecutive runs

PR #1372, opened 2026-09-18
Return HTTP/2 connection window credit for response bytes never read

### LINKED ISSUE #1371: HTTP/2 client leaks connection-level flow-control window when a response body is closed unread (retries, HTTP.open)
## Summary

On HTTP/2, the client never returns flow-control credit for response bytes that were received but not read. Every response body that gets closed unread (`body_close!(::H2Body)`) permanently shrinks the connection-level receive window. Once the window reaches zero the server can no longer send any `DATA` on that connection, and the next request that does try to read its body hangs until a timeout fires or the peer resets the stream.

This is easy to hit without ever touching the streaming API, because the built-in retry loop closes the body of every retried attempt unread. Polling an endpoint that answers `5xx` for a while (each `HTTP.get` = up to 5 attempts, 4 discarded bodies) exhausts the window after a handful of calls; on a real server behind a CDN this showed up as a ~5 minute hang followed by `HTTP/2 stream reset by peer: PROTOCOL_ERROR`, and the connection stayed unusable afterwards. HTTP/1.1 (`protocol=:h1`) is unaffected.

Versions: HTTP.jl v2.7.0 (also reproduced on v2.6.7), Julia 1.12.7, Linux x86_64.

## Reproducer

Fully local, only depends on HTTP.jl. The server returns a body of exactly 65535 bytes, which is the default connection window, so a single unread body is enough.

```julia
using HTTP

# Every response carries a body of exactly 65535 bytes: the default HTTP/2
# connection-level receive window. A single fully received body uses up the
# whole window, so the client must send WINDOW_UPDATE for those bytes before
# the server can send any further DATA on that connection.
const BODY = repeat("x", 65535)
const STATUS = Ref(200)
server = HTTP.serve!("127.0.0.1", 0; listenany=true) do req
    return HTTP.Response(STATUS[]; body=BODY)
end
url = "http://127.0.0.1:$(HTTP.port(server))/"

function report(f, label)
    t = time()
    try
        r = f()
        println(label, ": ok, status=", r.status, ", body=", length(r.body), " bytes (",
                round(time() - t; digits=2), "s)")
    catch e
        println(label, ": FAILED after ", round(time() - t; digits=1), "s: ", sprint(showerror, e))
    end
end

# Scenario A: discard one unread body via HTTP.open, then do a normal request
# on the same connection.
for protocol in (:h1, :h2)
    STATUS[] = 200
    client = HTTP.Client()
    # Read the headers only. The body is never consumed, so HTTP.jl closes the
    # response body (RST_STREAM CANCEL on h2) when the do-block returns.
    HTTP.open("GET", url; client, protocol, retry=false, request_timeout=5) do stream
        HTTP.startread(stream)
    end
    report("A/$protocol: GET after one discarded body") do
        HTTP.get(url; client, protocol, retry=false, request_timeout=5)
    end
    close(client)
end

# Scenario B: the same thing happens implicitly inside the built-in retry loop.
# The server answers 503, so HTTP.get retries and discards the body of every
# attempt except the last one.
for protocol in (:h1, :h2)
    STATUS[] = 503
    report("B/$protocol: GET with default retries against a 503") do
        HTTP.get(url; protocol, status_exception=false, retry_bucket=false,
                 respect_retry_after=false, request_timeout=10)
    end
end

HTTP.forceclose(server)
```

Output (identical on three consecutive runs):

```
A/h1: GET after one discarded body: ok, status=200, body=65535 bytes (0.01s)
A/h2: GET after one discarded body: FAILED after 5.0s: http timeout during request (budget 5000000000 ns, elapsed 5006082249 ns)
B/h1: GET with default retries against a 503: ok, status=503, body=65535 bytes (0.34s)
B/h2: GET with default retries against a 503: FAILED after 10.0s: http timeout during request (budget 10000000000 ns, elapsed 10000559999 ns)
```

With smaller bodies the same thing happens after more discards, e.g. a 4096-byte body stalls after the 16th unread body (16 · 4096 > 65535). Against public HTTP/2 servers the numbers line up too: with a 1576-byte 404 page, the 43rd `HTTP.open` that leaves the body unread gets `RST_STREAM INTERNAL_ERROR` from the server (65535 / 1576 ≈ 41.6).

## Where it goes wrong

The only place the client sends `WINDOW_UPDATE` is `_send_window_updates!`, and it is only called from `body_read!` for the bytes that were actually copied out to the caller:

https://github.com/JuliaWeb/HTTP.jl/blob/v2.7.0/src/http2_client.jl#L1736-L1741

`body_close!(::H2Body)` marks the stream done, sends `RST_STREAM(CANCEL)` and unregisters the stream, but never credits the bytes that are still sitting in `state.body` (or that the peer already put on the wire before it sees the reset):

https://github.com/JuliaWeb/HTTP.jl/blob/v2.7.0/src/http2_client.jl#L1762-L1784

The retry path discards responses through exactly this function, so the leak is hit by default settings:

https://github.com/JuliaWeb/HTTP.jl/blob/v2.7.0/src/http_client.jl#L1031-L1033

Per RFC 9113 §6.9 the connection-level window is consumed by every `DATA` frame received regardless of what happens to the stream afterwards, so the client has to return credit for received-but-unread bytes when it abandons a stream (and for any `DATA` that still arrives for a stream it has already reset). The per-stream window doesn't matter after the reset; the connection-level one does.

## Workarounds

- `protocol=:h1`
- avoid unread bodies, e.g. `retry=false` (or a `retry_if` that never retries on a response) when polling endpoints that return `5xx`

krynju:
Fixes #1371.

## What was wrong

The HTTP/2 client only ever sent `WINDOW_UPDATE` for body bytes that the caller actually read (`body_read!` → `_send_window_updates!`). Bytes that arrived but were never read were silently dropped:

- `body_close!` on an unread body sent `RST_STREAM(CANCEL)` and unregistered the stream, but never credited the bytes still sitting in the stream buffer. The built-in retry loop closes the body of every attempt it gives up on this way, so a plain `HTTP.get` against an endpoint returning `5xx` leaked four bodies per call.
- DATA that arrived for a stream that was already reset/unregistered (our `RST_STREAM` racing the peer's in-flight frames) was ignored without credit.
- DATA frame padding is flow-controlled but never reaches the reader, so it was never credited either.

Every leaked byte permanently shrank the shared 65535-byte connection-level window. Once it hit zero the server could not send `DATA` on *any* stream of that connection, and the next request that tried to read its body hung until a timeout or a peer reset. With the default 65535-byte window a single unread 65535-byte body is enough; see the reproducer in #1371.

## What this does

Return the credit at the point where bytes are received but will never reach a reader:

- **`body_close!`** computes the unread buffered bytes (under the stream lock, in the same section that marks the stream done) and sends a connection-level `WINDOW_UPDATE` right after the `RST_STREAM`. Same on the `Content-Length` overflow path in `body_read!`.
- **Read loop**: `_handle_stream_data!` refunds DATA for a stream that is already done or failed instead of dropping it, and the `DataFrame` branch refunds DATA for a stream that is no longer registered (only for odd stream ids below `next_stream_id`, i.e. streams we actually opened; anything else stays ignored as before).
- **Padding**: `DataFrame` gets a `padding::Int` field (pad-length octet included) filled in by `read_frame!`; the read loop credits it immediately. The 3-argument constructor is kept, so existing call sites and `write_frame!` are unchanged.
- `_send_window_updates!` gets a `stream_level` switch. Abandoned streams only get the connection-level update; a stream-level update for a stream we reset is meaningless. This mirrors the server-side helper, which already has the same switch.

No public API changes.

## How Go's HTTP/2 transport handles this

I compared with `golang.org/x/net/http2` (what `net/http` uses), since it deals with exactly these three cases:

- Closing an unread body: [`transportResponseBody.Close`](https://github.com/golang/net/blob/656874fd46a65a97eac94003e0eb8751be4403d0/http2/transport.go#L2229-L2252) aborts the stream, then takes `unread := cs.bufPipe.Len()` and sends a **connection-level** `WINDOW_UPDATE` for it ("Return connection-level flow control"). That is what `body_close!` now does.
- DATA for a stream that no longer exists: [`processData`, `cs == nil` branch](https://github.com/golang/net/blob/656874fd46a65a97eac94003e0eb8751be4403d0/http2/transport.go#L2271-L2301) — "We probably did ask for this, but canceled. Just ignore it. But at least return their flow control" — again connection-level only. That is the new `state === nothing` branch in the read loop. (Go treats a stream id it never issued as a connection error; this PR keeps HTTP.jl's existing behaviour of ignoring those, it just doesn't credit them.)
- Padding and data written to an already-closed stream: [`processData` refund block](https://github.com/golang/net/blob/656874fd46a65a97eac94003e0eb8751be4403d0/http2/transport.go#L2334-L2365) — "Return any padded flow control now, since we won't refund it later on body reads" and "Return len(data) now if the stream is already closed, since data will never be read". Go skips the stream-level update when the stream was reset (`if !didReset`), which is the `stream_level=false` case here.

One deliberate difference: Go batches window updates through [`inflow.add`](https://github.com/golang/net/blob/656874fd46a65a97eac94003e0eb8751be4403d0/http2/flow.go#L33-L53) (it withholds updates smaller than 4 KiB unless they would at least double the peer's window) and also enforces the receive window on the incoming side. HTTP.jl currently sends an update per read and does not track the receive window, and this PR keeps that model; the refunds are sent immediately, like the existing per-read updates. Batching/enforcement would be a separate change.

## Tests

- `HTTP/2 client returns connection-level window credit for a body closed unread (#1371)` — wire-level: a fake server sends a 4000-byte body without `END_STREAM`, the client closes it unread, the server asserts it sees `RST_STREAM(CANCEL)` followed by `WINDOW_UPDATE(0, 4000)`, then sends 1000 more bytes on the reset stream and asserts `WINDOW_UPDATE(0, 1000)`; no stream-level updates.
- `HTTP/2 connection stays usable after response bodies are discarded (#1371)` — `HTTP.serve!` with a 65535-byte body: `HTTP.open` + `startread` without reading, then a normal `HTTP.get`; and a `503` with default retries. Both previously hung until `request_timeout`.
- The existing padded-DATA test now also checks that the full 6-byte payload (2 data + 4 padding) is credited at the connection level and only the 2 data bytes on the stream.

Both new tests fail on `master` and pass with this change. Full test suite passes locally (Julia 1.12, 4 threads).


quinnj:
The new test requires refunds in the order `[4000, 1000]`, but the two tasks can validly send `[1000, 4000]`. I forced that order and the test failed. Please compare the refunds without requiring their order (`test/http2_client_tests.jl:1785`).

quinnj:
Both reported cases pass with this fix. One related leak remains: if a reset arrives during upload, before `H2Body` is created, `_h2_roundtrip_incoming!` unregisters the stream without refunding buffered bytes. I reproduced 65,535 bytes discarded with zero credit returned. This also fails on the base branch. Could the error cleanup refund those bytes too?

Issue #71, opened 2017-08-12
Origin header breaks AWS EC2 API


samoconnor:
See: https://github.com/JuliaWeb/HTTP.jl/commit/90061e78b96c28c11662bb56867e173df8ea5dd5#commitcomment-23615586

samoconnor:
FYI, I'm working around this by creating a `HTTP.Request` object directly (to avoid the constructor that sets default headers): https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/http.jl#L33

quinnj:
We no longer add the `Origin` header; closed via #77

Issue #48, opened 2017-05-03
Error when testing with ab


JockLawrie:
Hi there,

I have a simple server that works when making a request from Julia, but fails when making a request from the ab tool. Minimal example below. Any ideas?

```julia
using HTTP

function app(req, res)
    res.body = FIFOBuffer("Hello world!")
    res
end

server = HTTP.Server(app)
HTTP.serve(server, IPv4(127, 0, 0, 1), 8000)
````

This works:
```julia
using HTTP
res = HTTP.get("http://127.0.0.1:8000/")
```

But this fails:
```
ab -n 1 "http://127.0.0.1:8000/"
```

By contrast the following server works in both cases:
```julia
using HttpServer

function app(req, res)
    res.data = "Hello world!"
    res
end

server = Server(app)
run(server, 8000)
```

joshbode:
The error in question:

```
[HTTP - 2017-05-04T13:34:13.434]: starting server to listen on: 127.0.0.1:8000
[HTTP - 2017-05-04T13:34:29.112]: throttling 127.0.0.1
[HTTP - 2017-05-04T13:34:29.117]: new tcp connection accepted, reading request...
[HTTP - 2017-05-04T13:34:29.645]: processing on connection i=0...
[HTTP - 2017-05-04T13:34:38.446]: throttling on connection i=0
[HTTP - 2017-05-04T13:34:38.511]: reading request bytes with readtimeout=180.0
[HTTP - 2017-05-04T13:34:53.843]: received request on connection i=0
HTTP.Request:
"""
GET / HTTP/1.0
Host: 127.0.0.1:8000
Accept: */*
User-Agent: ApacheBench/2.3

"""
[HTTP - 2017-05-04T13:34:54.605]: responding with response on connection i=0
HTTP.Response:
"""
HTTP/1.1 200 OK
Connection: close
Content-Length: 12
Content-Type: text/html; charset=utf-8
Date: Thu, 04 May 2017 03:34:54
Content-Language: en
Server: Julia/0.5.1

Hello, World
"""
ERROR (unhandled task failure): ArgumentError: stream is closed or unusable
 in check_open(::TCPSocket) at ./stream.jl:258
 in uv_write(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:806
 in unsafe_write(::TCPSocket, ::Ptr{UInt8}, ::UInt64) at ./stream.jl:830
[...]

PR #1374, opened 2026-09-19
Preserve owned headers and bound buffered transfer allocations

quinnj:
Buffered cloud transfers repeatedly copy header state, and HTTP/2 uploads allocate and copy payload-sized buffers during frame construction. This change provides an explicit owned-header path and bounds transfer scratch storage while preserving retries, signing hooks, wire validation, and framing.

- Canonicalize raw header vectors once. Copy existing `Headers` without re-canonicalizing keys.
- Support `copyheaders=false` for owned `HTTP.Headers` in `request`, `open`, and `Request`. Avoid intermediate high-level request copies; retain independent per-attempt and wire header state.
- Read built-in response bodies into caller vectors/views without the extra staging buffer. Preserve capacity/EOF/trailer checks and custom-body fallbacks.
- Reuse HTTP/2 DATA-frame and streaming look-ahead buffers; avoid copying slices on partial transport writes.
- Borrow string-backed bytes through scoped, GC-rooted arrays during framing. Use safe pointer writes for string views so transport cancellation cannot outlive an unrooted owner. This avoids generic transport materialization and repeated whole-string alias checks.
- Document buffer lifetimes and add ownership/retry tests plus a Linux CI allocation gate with a separate receiver process.

Local Julia 1.12.6 loopback measurements: a 16 MiB HTTP/2 upload fell from **111,722,656 B to 432,816 B allocated**. HTTP/1 remained bounded (57,648 B → 57,200 B). These are client allocations, not cloud throughput claims. Reproduce with `julia --threads=4 --project=. bench/buffered_upload_allocations.jl --check`.

Validation: full HTTP suites passed locally on Julia 1.10.11 and 1.12.6, including protocol/retry/proxy checks. The 1.12 trim compile verifier passed; task-backed executable runs retain the existing Julia-runtime skip policy. Documentation built locally. The final string-buffer changes pass the full 1.12 suite and the eight-case HTTP allocation gate. A Windows scripted-response fixture now waits for client consumption before closing its peer. Seven additional assertions cover downstream response bodies that implement only the original Vector destination method, including views and overflow. 

Coordinates with [CloudBase #48](https://github.com/JuliaServices/CloudBase.jl/pull/48) and [CloudStore #71](https://github.com/JuliaServices/CloudStore.jl/pull/71). No new exported names.

All applicable CI checks passed on `c87c218f`.

**Review follow-up (273a141b):**
- `Headers(::Vector{Pair{String,String}})` now goes through `mkheaders` like every other vector, so adjacent duplicate keys join the same way `Request` joins them. The vector constructor still canonicalizes keys and leaves the source untouched.
- `copyheaders=false` accepts any header input. An `HTTP.Headers` collection is adopted without copying; anything else becomes a new collection as with the default, so 1.x call sites and forwarded `kw...` keep working.
- Removed the HTTP/1 string-view pointer write. String code-unit views are `StridedVector`s and were already written without a copy (8320 B allocated before, 8144 B after, for a 16 MiB upload). The HTTP/2 string borrow stays; it exists because `copyto!` alias checks hash the whole backing string per chunk.
- HTTP/2 frame buffer sized to the buffered body when smaller than one frame; comment and docs corrected.

Co-authored by Codex


src/http_core.jl
```diff
@@ -643,14 +643,19 @@ Ordered, case-canonicalized collection of header pairs.
...
+its keys without changing the source vector.
 """
 mutable struct Headers <: AbstractVector{Pair{String,String}}
     entries::Vector{Pair{String,String}}
-end
 
-"""Create and return an empty `Headers` collection."""
-Headers() = Headers(Pair{String,String}[])
+    """Create and return an empty `Headers` collection."""
+    Headers() = new(Pair{String,String}[])
```
quinnj, line 654: I'm worried about the perf hit now that we don't have a "pass-through" method and are allocating every time.

docs/src/guides/client.md
```diff
@@ -549,8 +549,57 @@ as temporary compatibility, not as the preferred API:
...
+
+HTTP/2 uploads use a reusable 16 KiB DATA-frame payload buffer. Header and payload
+are coalesced into one transport write to avoid tiny TLS records. Streaming bodies
+also use two reusable look-ahead buffers. TLS encryption, receive flow control,
+and protocol metadata still have their own costs. Reuse an `HTTP.Client` to reuse
+connections. Functional tests cover ownership, replay, and destination views.
+The allocation gate below measures buffered uploads without retries; network
+benchmarks are still needed to establish throughput.
+
+Run `julia --threads=4 --project=. bench/buffered_upload_allocations.jl --check`
```
quinnj: We should remove this paragraph; unnecessary in our client.md doc (more of a dev-focused comment)

